PR #1, opened 2019-12-21
minor changes

doanguyen:
regarding README & small non-critical problem


README.md
```diff
@@ -328,7 +328,7 @@ All work on SimpleLogin happens directly on GitHub.
 
 ### Run code locally
 
-The project uses Python 3.6+. First, install all dependencies by running the following command. Feel free to use `virtualenv` or similar tools to isolate development environment.
+The project uses Python 3.7+. First, install all dependencies by running the following command. Feel free to use `virtualenv` or similar tools to isolate development environment.
```
nguyenkims: Thanks for spotting this out! `dataclasses` is indeed only available in python 3.7+

app/config.py
```diff
@@ -3,7 +3,7 @@
 
 from dotenv import load_dotenv
 
-SHA1 = subprocess.getoutput("git rev-parse HEAD")
+SHA1 = lambda: subprocess.getoutput("git rev-parse HEAD")
```
nguyenkims: Could you explain why `SHA1` should be a function here?

app/models.py
```diff
@@ -99,7 +99,7 @@ class User(db.Model, ModelMixin, UserMixin):
 
     @classmethod
     def create(cls, email, name, password=None, **kwargs):
-        user: User = super(User, cls).create(email=email, name=name, **kwargs)
```
nguyenkims: The type hint here is actually for having better auto-completion.

app/utils.py
```diff
@@ -7,19 +7,20 @@
 from .config import WORDS_FILE_PATH
 from .log import LOG
 
-with open(WORDS_FILE_PATH) as f:
-    LOG.d("load words file: %s", WORDS_FILE_PATH)
-    _words = f.read().split()
+def _word():
```
nguyenkims: Actually the reason `_words` was loaded at module level before is to avoid calling `random_word()` or `random_words` loads the `WORDS_FILE_PATH` again. If `WORDS_FILE_PATH` is a big file, then the loading time could be expensive.

app/config.py
doanguyen: subprocess module spawn a new os-level process, the config module is interpreted at every http request, so it a waste of cpu

nguyenkims: Ah actually the `SHA1` is evaluated only once first time when `config` module is imported. You can test that by adding this snippet of toy-code to `config.py`: the `>>>b is called` should be printed only once.

```
def b():
    print(">>>b is called")
    return 10

A = b()
```

app/models.py
doanguyen: I tried and droped mypy quite a long time ago, since then it has improved the SelfType so removal is no longer valid

nguyenkims: ah ok I see how you got the error then. Yeah I hope `mypy` supports SelfType now. I'm using Pycharm personally and they don't spot any error.

app/config.py
doanguyen: You're right, I was so stupid to think the every variable is garbage collected after a request. You may now close this PR since it's built upon my misunderstanding on how Wezeurg works.

nguyenkims: No worry :), this is not clearly said in Flask doc neither. I'll close this PR but will make sure to use update the doc to use Python 3.7+.

PR #5, opened 2019-12-26
Implement "Add user setting to choose between words based random alias and uuid alias"

doanguyen:


app/dashboard/templates/dashboard/index.html
```diff
@@ -33,13 +33,34 @@
...
           </button>
         </form>
-        <form method="post">
-          <input type="hidden" name="form-name" value="create-random-email">
-          <button data-toggle="tooltip"
-                  title="Create a totally random alias"
-                  class="btn btn-success">Random Alias
-          </button>
-        </form>
+                    <div class="btn-group" role="group">
```
nguyenkims: the indentation doesn't look right here, maybe your IDE mixes up space and tab?

app/dashboard/templates/dashboard/setting.html
```diff
@@ -56,6 +56,18 @@ <h3 class="mb-0">Change password</h3>
       <input type="hidden" name="form-name" value="change-password">
       <button class="btn btn-outline-primary">Change password</button>
     </form>
+      <hr>
+
+    <h3 class="mb-0">Random Alias</h3>
+    <div class="small-text mb-3">Choose how to create your email alias by default</div>
+    <form method="post" class="form-inline">
+      <input type="hidden" name="form-name" value="change-alias-generator">
+      <select class="custom-select mr-sm-2" name="alias-generator-scheme" id="alias-generator-scheme">
```
nguyenkims: the `id="alias-generator-scheme"` can be removed here I think

app/models.py
```diff
@@ -83,13 +84,23 @@ class PlanEnum(enum.Enum):
...
+
+
 class User(db.Model, ModelMixin, UserMixin):
     __tablename__ = "users"
     email = db.Column(db.String(128), unique=True, nullable=False)
     salt = db.Column(db.String(128), nullable=False)
     password = db.Column(db.String(128), nullable=False)
     name = db.Column(db.String(128), nullable=False)
     is_admin = db.Column(db.Boolean, nullable=False, default=False)
+    alias_generator = db.Column(db.Integer, nullable=True, default=AliasGeneratorEnum.word.value)
```
nguyenkims: Maybe we can use `nullable=False` here? A field is easier to work with when it's non-nullable, that's why fields are generally non-nullable. Exception made for the ones that null value makes sense like `profile_picture_id`, `custom_domain_id`.

app/models.py
```diff
@@ -362,20 +373,27 @@ def is_expired(self):
         return self.expired < arrow.now()
 
 
-def generate_email() -> str:
-    """generate an email address that does not exist before"""
-    random_email = random_words() + "@" + EMAIL_DOMAIN
+def generate_email(scheme: int = 1, in_hex: bool = False) -> str:
```
nguyenkims: `scheme: int = 1` could be replaced by `scheme: int = AliasGeneratorEnum.word`. The reason for that is enum should be used instead of its value.

nguyenkims:
Overall the PR looks good! I have checkout it locally and tested, the feature works well 👍.

I just left some comments, some might be a bit "nitpicking" as I try to minimize the code change as much as possible ... 

Don't hesitate to use `black` to reformat the Python code. Some IDEs might break black formattings so please make sure to run `black` before creating the commit.

PR #7, opened 2019-12-27
Mfa

nguyenkims:
This PR adds MFA support to the webapp. 

- User can choose to enable MFA: they will need to scan the code in Google Authenticator or Authy app, then validate the 6-digit number
- For all future logins, whether standard login username/password or via social login, user needs to go through the MFA step.


app/auth/views/login_utils.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Redirect to the correct page after login.
+    If user enables MFA: redirect user to MFA page
+    Otherwise redirect to dashboard page if no next_url
+    """
+    if user.enable_otp:
+        session[MFA_USER_ID] = user.id
+        if next_url:
+            return redirect(url_for("auth.mfa", next_url=next_url))
+        else:
```
doanguyen: The `else` clause here is redundant since the previous line has already returned. Of course, this is only the taste on how to organize the code, not the logic itself. May be the `else` clause makes it easier to read. Please feel free to keep this way for the consistency.

app/auth/views/mfa.py
```diff
@@ -0,0 +1,52 @@
...
+    if not user.enable_otp:
+        raise Exception("Only user with MFA enabled should go to this page. %s", user)
+
+    otp_token_form = OtpTokenForm()
+    next_url = request.args.get("next")
+
+    if otp_token_form.validate_on_submit():
+        totp = pyotp.TOTP(user.otp_secret)
+
+        if otp_token_form.validate_on_submit():
```
doanguyen: I don't understand why `otp_token_form.validate_on_submit()` needs to validate twice?

app/auth/views/login_utils.py
nguyenkims: The `else` is indeed redundant.

app/auth/views/mfa.py
nguyenkims: Nice spot! This is a mistake, let me update the PR.

PR #12, opened 2019-12-29
Email notification

doanguyen:
Allow user to opt-out newsletter providing by SL


app/dashboard/views/newsletter.py
```diff
@@ -0,0 +1,6 @@
+from app.dashboard.base import dashboard_bp
```
nguyenkims: this file can be deleted right?

PR #14, opened 2019-12-31
Alias log

doanguyen:
This pr contains:

1. A naive implementation (without creating overhead to count all db objects) of paginate alias log page
2. Try to get rid of `dataclass` dependencies which is Py3.7 only

### DIFF
--- a/app/dashboard/views/alias_log.py
+++ b/app/dashboard/views/alias_log.py
@@ -1,12 +1,14 @@
 from dataclasses import dataclass
 
 import arrow
-from flask import render_template, flash, redirect, url_for
+from flask import render_template, flash, redirect, url_for, abort
 from flask_login import login_required, current_user
 
 from app.dashboard.base import dashboard_bp
 from app.extensions import db
 from app.models import GenEmail, ForwardEmailLog, ForwardEmail
+
+LIMIT = 3
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


app/dashboard/views/alias_log.py
```diff
@@ -1,27 +1,32 @@
...
-    is_reply: bool
-    blocked: bool
+    __slots__ = [
+        "website_email",
+        "website_from",
+        "alias",
+        "when",
+        "is_reply",
+        "blocked",
+    ]  # memory efficiency
```
nguyenkims: Ah could you tell more about why we should not use `dataclass` please?

app/dashboard/views/alias_log.py
```diff
@@ -1,27 +1,32 @@
...
-import arrow
-from flask import render_template, flash, redirect, url_for
+from flask import render_template, flash, redirect, url_for, abort
 from flask_login import login_required, current_user
 
 from app.dashboard.base import dashboard_bp
 from app.extensions import db
 from app.models import GenEmail, ForwardEmailLog, ForwardEmail
 
+LIMIT = 15
```
nguyenkims: to avoid other modules from using this constant, I think we could add `_` prefix to its name

app/dashboard/views/alias_log.py
doanguyen: For no good reason actually, but without the dataclass dependent, the app can also run on py36. Other solution is to install dataclass package on pypi for py<37, but I think it's overcomplicated

nguyenkims: Ah I think we don't have to support python 3.6 as the code is meant to be run inside a container. There are alternatives to dataclasses but they (attrs for ex) but dataclasses has the advantage of being included in the standard lib.

doanguyen: Couldn't agree more on the standard lib of py37. I think SL should support as many Python versions as it can, since the lower Python version is, the more mature and the number of developers are working with it. You're right that the code is meant to run inside a container, but for developing the app, for myself, I have to set up another version of Python. So to me, even this is not a big concern, I will create another PR regarding just this story and let it be there, we will consider later.

nguyenkims: ah ok I see. In this case https://github.com/python-attrs/attrs could be a good option indeed.

PR #20, opened 2020-01-03
Alias description

digitex3d:


nguyenkims:
I just pulled and tested, the feature works fine 👍. I have some feedbacks though:

- The description is only shown for "default" and not custom-domain alias. 
image

Would be great to show this for custom-domain alias too. In this case the current UI could be too much crowded, I think we might use the UI from the browser extension

image


- User cannot update the description. Would be better to have the description field in the alias listing page so user you can update it. I suggest this UI

image

- The PR is missing a migration file. This works fine locally as the database is created from scratch but would fail on production. The readme contains a small section about the migration.

- (minor) I've been thinking about the name of this feature, maybe `note` would be more natural for users than `description`.

digitex3d:
To answer your feedbacks:

- The description is only shown for "default" and not custom-domain alias.
I totally agree with your UI proposal, maybe just show 'simplelogin.io' as default domain.

- User cannot update the description. Would be better to have the description field in the alias listing page so user you can update it. I suggest this UI
I agree with your UI proposal.

-(minor) I've been thinking about the name of this feature, maybe note would be more natural for users than description.
I have really no preference on that.

nguyenkims:
I just created this PR https://github.com/simple-login/app/pull/73 that adds the alias note/description feature :)

PR #21, opened 2020-01-04
change the smtp response code

doanguyen:
I think `250` is when we successfully forward the email, in this case, the email is not forwarded but we refuse to do so, thus `450` is more consistent.

### DIFF
--- a/email_handler.py
+++ b/email_handler.py
@@ -244,7 +244,7 @@ class MailHandler:
                 "",
             )
 
-            return "250 ignored"
+            return "450 ignored"
 
         # remove DKIM-Signature
         if msg["DKIM-Signature"]:


nguyenkims:
Ah I also thought about this a lot when using the 2** code. The idea behind is when user disabled an alias, that means they don't want to hear from the sender at this moment but they might re-enable them in the future. Returning a 4** error might stop the sender from sending the emails to the alias.

doanguyen:
https://github.com/simple-login/app/pull/21/files#diff-e1645a46354215ec708e4a51ea83a269R233

In this case, the sender is not authenticated to send the email, so I think regardless the alias is enable or not it should return an error code?

nguyenkims:
Yeah if user is not authenticated to "reply" the code returns  "450 ignored":
https://github.com/simple-login/app/blob/0e7016b2e09d821a81cfae1369829c5e395adc6d/email_handler.py#L247

doanguyen:
Oops, this is the link from this pr, in the current master branch, it returns 250 instead of 450

nguyenkims:
ah ok I see, this should be `450` instead indeed.

PR #22, opened 2020-01-05
Alias dashboard

doanguyen:


app/dashboard/views/alias_log.py
```diff
@@ -38,6 +38,15 @@ def alias_log(alias, page_id):
         return redirect(url_for("dashboard.index"))
 
     logs = get_alias_log(gen_email, page_id)
+    base = (
+        db.session.query(ForwardEmail, ForwardEmailLog)
+        .filter(ForwardEmail.id == ForwardEmailLog.forward_id)
+        .filter(ForwardEmail.gen_email_id == gen_email.id)
+    )
+    total = base.count()
+    email_forwarded = base.filter(ForwardEmailLog.is_reply == False).count()
```
nguyenkims: `nb_blocked==False` should be added as condition to compute `email_forwarded`

app/dashboard/templates/dashboard/alias_log.html
```diff
@@ -12,7 +73,37 @@ <h3 class="page-title col">
...
+            <div class="card-counter primary">
+                <i class="fa fa-reply"></i>
+                <span class="count-numbers">{{ email_replied }}</span>
+                <span class="count-name">Email Replied</span>
+            </div>
+        </div>
+        <div class="col-3">
+             <div class="card-counter danger">
+                <i class="fa fa-ban"></i>
+                <span class="count-numbers">{{ email_forwarded }}</span>
```
nguyenkims: I think this should be `email_blocked` here

server.py
```diff
@@ -444,6 +445,6 @@ def do_not_track():
         context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
         context.load_cert_chain("local_data/cert.pem", "local_data/key.pem")
 
-        app.run(debug=True, host="0.0.0.0", port=7777, ssl_context=context)
+        app.run(host="0.0.0.0", port=7777, ssl_context=context)
```
nguyenkims: These lines are only executed in local environment so debug (gunicorn is used on production) should always be True I think.

server.py
```diff
@@ -67,6 +68,8 @@
 def create_app() -> Flask:
     app = Flask(__name__)
     app.url_map.strict_slashes = False
+    app.debug = DEBUG
+    os.environ["FLASK_DEBUG"] = str(DEBUG)
```
nguyenkims: do you know what changes if debug is set to true?

nguyenkims:
New UI seems really nice!

server.py
doanguyen: you mean the first or the second line?

server.py
doanguyen: This is one of the step I want to introduce the debug flag to the flask `app` instance. Because from several previous PRs, you added the `flask profiler` dependency, so we can turn it on/off just by checking the debug flag

nguyenkims: I think it's more clear that each setting does only one thing and ideally, should be independent to the others. If someone wants to turn FLASK_PROFILER off, he can leave this section uncommented (it's uncommented by default).  "umbrella" setting like DEBUG tends to affect other settings, from my experience it usually causes confusion.

server.py
nguyenkims: I mean the second line. This setting is used by `flask cli` I think and as `server.py` is not launched via `flask cli`, I'm not sure that setting could have any effect.

server.py
doanguyen: You're right, I used it in my development workflow but didn't think too much on the downside. It is fixed in the last PR.

server.py
doanguyen: (y) i rolled back these flags, it is indeed has no effect with the server script

PR #29, opened 2020-01-12
Dashboard column

doanguyen:
First implementation of dashboard in columns manner instead of cards.

Things to continue doing:
* [ ] mobile friendly for the table/column

### DIFF
--- a/app/dashboard/templates/dashboard/alias_log.html
+++ b/app/dashboard/templates/dashboard/alias_log.html
@@ -74,28 +74,28 @@
     </h3>
   </div>
     <div class="row">
-        <div class="col-3">
+        <div class="col-6 col-lg-3 col-sm-6">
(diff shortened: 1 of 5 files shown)


nguyenkims:
The new layout looks very good 👍. I have some feedbacks though:

- the new layout is not responsive. Either we could make it response or we can use the current layout (but keeping the new `index_column.html` here) until the new layout is fully response?

- alias activity and contact manager now uses `alias_id` instead of `alias` to avoid `/` introduced by directory alias. 

- Some UI feedbacks I put directly in the screenshot:
Alias | SimpleLogin 2020-01-12 22-25-58

doanguyen:
this pr is indeed not a ready to ship one, i'm pushing it so that we can discuss on how to make it mobile friendly together with other things

PR #33, opened 2020-01-19
Use Cloudflare DNS nameserver

nguyenkims:
- Use CloudFlare DNS nameserver 1.1.1.1
- Replace 1.1.1.0/24 by 240.0.0.0/24 in docker network: 240.0.0.0/4 is a reserved IP https://en.wikipedia.org/wiki/Reserved_IP_addresses

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,8 @@
+.idea/
+*.pyc
 db.sqlite
+.env
+.pytest_cache
+.vscode
+.DS_Store
+config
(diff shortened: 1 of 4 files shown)


doanguyen:
Could you tell me the reason behind this pr? Is the cloudflare provides better dns service than the one with google?

nguyenkims:
Yeah Cloudflare nameserver is supposed to be more privacy-focused than the Google one.

PR #45, opened 2020-01-23
Improve the Dockerfile

NicolasCARPi:
* reduce number of layers
* remove vim installation (I'm guessing it was here for debug purposes, also this is not a good way to install a package in Docker)
* add LICENSE and README to .dockerignore
* don't cache with pip
* add missing EXPOSE directive

This set of changes reduces the image size by almost 100 Mb!

As a side note, I would recommend renaming this repo to something more meaningful than `app`. When cloning a repo, the organization doesn't appear, only the repo name. In my `$dev` folder, I have several repositories, and their name is the name of the application, imagine if they were all called `app` ;)

Cheers
~Nico

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,5 @@ db.sqlite
 .vscode
 .DS_Store
 config
+LICENSE
+README.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,12 @@ FROM python:3.7
 
 WORKDIR /code
 
-# copy everything into /code
-COPY . .
-
 # install dependencies
+COPY ./requirements.txt ./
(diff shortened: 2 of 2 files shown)


nguyenkims:
Hey thank you for the PR, great that you can get the size 100MB less!

`vim` is indeed for debugging purpose only :)

It's better to copy the `requirements.txt` first and install all dependencies before copying everything. As this file is rarely changed, this allows to re-use the Docker cache.

NicolasCARPi:
Yes, that makes sense. Anyway this change has no real impact on size so I've pushed another commit splitting again the requirements and the rest of the code.

nguyenkims:
Great, thanks!

PR #60, opened 2020-01-29
Hide seperator from 'from header' if no name present

SibrenVasse:
If in original message no name is present, email is forwarded like this:
**- hi at simplelogin.io**

This commit adds a conditional to hide the unnecessary seperator.
**hi at simplelogin.io**

Not runtime tested.


nguyenkims:
Cool, thanks for the PR!

PR #66, opened 2020-02-02
use user.name when reply

nguyenkims:


SibrenVasse:
Why wasn't this merged?

nguyenkims:
The reason I discarded this PR (that I made :) ) because there are actually 2 categories of alias:
- aliases to hide our real identity: the most popular use of alias.
- aliases that we want to associate with our name. Alias created as business emails often fall into this category.

Using a single name for both categories is not ideal.
I think this feature needs to be carefully analysed before starting the implementation.

PR #128, opened 2020-04-13
Dark mode implemetation

doanguyen:
We used CSS variables to implement the dark mode feature

image

### DIFF
--- a/app/dashboard/templates/dashboard/domain_detail/dns.html
+++ b/app/dashboard/templates/dashboard/domain_detail/dns.html
@@ -8,7 +8,7 @@
 
 
 {% block domain_detail_content %}
-  <div class="bg-white p-4" style="max-width: 60rem; margin: auto">
+  <div class="p-4" style="max-width: 60rem; margin: auto">
     <h1 class="h3"> {{ custom_domain.domain }} </h1>
     <div class="">Please follow the steps below to set up your domain.</div>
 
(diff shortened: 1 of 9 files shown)


app/dashboard/templates/dashboard/domain_detail/dns.html
```diff
@@ -28,15 +28,15 @@ <h1 class="h3"> {{ custom_domain.domain }} </h1>
       </div>
 
       <div class="mb-2">Add the following MX DNS record to your domain. <br>
-        Please note that there's a point (<em>.</em>) at the end target addresses. <br>
-        Also some domain registrars (Namecheap, CloudFlare, etc) might use <em>@</em> for the root domain.
+        Please note that there's a point (<em class="text-default">.</em>) at the end target addresses. <br>
```
nguyenkims, line 31: do you think there's a way to apply `.text-default` automatically to `<em>`?

app/dashboard/templates/dashboard/domain_detail/dns.html
```diff
@@ -28,15 +28,15 @@ <h1 class="h3"> {{ custom_domain.domain }} </h1>
...
+        Also some domain registrars (Namecheap, CloudFlare, etc) might use <em class="text-default">@</em> for the root domain.
       </div>
 
       {% for priority, email_server in EMAIL_SERVERS_WITH_PRIORITY %}
-        <div class="mb-3 p-3" style="background-color: #eee">
-          Domain: <em>{{ custom_domain.domain }}</em> or <em>@</em> <br>
+        <div class="mb-3 p-3 bg-secondary">
```
nguyenkims, line 36: 👍

static/assets/js/core.js
```diff
@@ -104,4 +104,18 @@ $(document).ready(function() {
       });
     });
   }
+
+  /**Dark mode controller */
+  var darkmode = localStorage.getItem('dark-mode');
```
nguyenkims, line 109: `localStorage` doesn't work in private mode in some browsers (I learnt it the hard way :) ). I think we can use `local-storage-polyfill.js` here which falls back to `cookie` automatically, here's how it's currently used in the code https://github.com/simple-login/app/blob/70ce48cd792b85914147c96bc70babbc4083edac/server.py#L475

nguyenkims:
@doanguyen the PR seems great overall! Just wonder is there a way to not replace the whole `dashboard.css` file and instead putting necessary change in another file? This file is replaced when `tabler` releases a new version actually.

I wonder maybe we can even have a `dashboard.dark.css` file that is a dark-built of `tabler` and switch to this one when user chooses the dark theme? This might require some changes in the back-end though to serve the `dashboard.dark.css` instead of `dashboard.css` according to user preference.

app/dashboard/templates/dashboard/domain_detail/dns.html
doanguyen: Yep, we can apply the style to the em directive globally. I am going to update it

static/assets/js/core.js
doanguyen: I saw the local storage polyfill but didn't know that it has problems with private mode. I'm going to update it

doanguyen:
The bad news is that I need to hook into the `dashboard.css` file to add this functionality (the steps that I took were: add some css variables for bright/dark modes, replace the  hard-coded colors from some elements with the new css variables). Maybe I can extract the new  feature to another css file for better understand/compatibility

doanguyen:
Closed in favor of #177

PR #154, opened 2020-05-03
Add default alias name to custom domain

SibrenVasse:
Implementation for https://trello.com/c/dz9npgnw/66-able-to-set-alias-name-for-a-domain

The name in the from header is first set by the alias name, then the default alias name from the domain and otherwise not present.

Implementation runtime tested. I'm not sure if I did the migration correctly, as I could not find any documentation on this. I did my best to try and follow a similar style to the rest of the codebase. Let me know what you think.


app/dashboard/views/domain_detail.py
```diff
@@ -141,6 +141,14 @@ def domain_detail(custom_domain_id):
             return redirect(
                 url_for("dashboard.domain_detail", custom_domain_id=custom_domain.id)
             )
+        elif request.form.get("form-name") == "set-name":
+            custom_domain.name = request.form.get("alias-name")
+            db.session.commit()
+            flash(f"Default alias name for Domain {custom_domain.domain} has been set", "success")
```
nguyenkims, line 147: this line is too long and will fail `black` check. You can run `black .` to format all files

email_handler.py
```diff
@@ -518,11 +518,19 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
...
 
     # make the email comes from alias
+    from_header = alias.email
+    # add alias name from alias
     if alias.name:
         LOG.d("Put alias name in from header")
         from_header = formataddr((alias.name, alias.email))
-    else:
-        from_header = alias.email
+    elif alias_domain not in ALIAS_DOMAINS:
```
nguyenkims, line 526: We can use `alias.custom_domain_id` here I think.  This field is filled up if the alias is dynamically created thanks to the directory feature.

We can also create a relationship like `custom_domain = db.relationship("CustomDomain", foreign_keys=[custom_domain_id])` to avoid doing a `CustomDomain.get()`, this is the same in terms of number of query to the database though.

nguyenkims:
@SibrenVasse Thanks a lot for the PR! 

For the database migration, this requires having a `Postgres` database indeed. On local environment, a `sqlite` database is created by the `db.create_all()` that bypasses the migration. This is convenient for development and unit tests as we don't have to wait for the migration.

If you have docker, you can create the migration by the following command: it first creates a postgres database for SimpleLogin, then run `flask db upgrade` to upgrade the DB to the latest stage and finally `flask db migrate` to generate the migration script. Let me know if this makes sense :). 

```
docker rm -f sl-db
docker run -p 5432:5432 --name sl-db -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=sl -d postgres

env DB_URI=postgresql://postgres:postgres@127.0.0.1:5432/sl flask db upgrade

env DB_URI=postgresql://postgres:postgres@127.0.0.1:5432/sl flask db migrate

docker rm -f sl-db
```

email_handler.py
SibrenVasse: I ended up adding the relationship in the model, this enabled me to display the domain-wide alias name as a placeholder in the alias view.

This does make alias.custom_domain_id redundant, but it has quite a few users left.

SibrenVasse:
The migration (up and down) ran successfully with your instructions.

email_handler.py
```diff
@@ -518,11 +518,19 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
...
+    from_header = alias.email
+    # add alias name from alias
     if alias.name:
         LOG.d("Put alias name in from header")
         from_header = formataddr((alias.name, alias.email))
-    else:
-        from_header = alias.email
+    elif alias.custom_domain_id:
+        LOG.d("Put domain default alias name in from header")
+        custom_domain: CustomDomain = CustomDomain.get_by(domain=alias_domain)
```
nguyenkims, line 528: I think we can use `alias.custom_domain` here instead of `CustomDomain.get_by`.

SibrenVasse:
I've resolved your comments.

nguyenkims:
@SibrenVasse perfect, thanks a lot! We'll test on staging and then deploy onto production today or tomorrow.

Issue #157, opened 2020-05-04
WebAuthn Support


developStorm:
Are there any plans for supporting the WebAuthn?

nguyenkims:
@developStorm we have this ticket in our roadmap https://trello.com/c/oWi7p2yp/55-support-other-mfa-types, feel free to vote on it to make it happen sooner :).

developStorm:
Oh, thanks! By the way, do you accept pull requests from users? If this is not your prioritize I can probably do something on this.

nguyenkims:
@developStorm we love PRs from users! 

For example, yesterday we had this PR https://github.com/simple-login/app/pull/154 that is created by a user who wants to have a feature more quickly. The feature is now live.

developStorm:
Thanks! I see there is [another ticket about this](https://trello.com/c/oWi7p2yp/55-support-other-mfa-types) which has more votes than the ticket you mentioned above. Do you have an estimated time of officially starting of development of U2F? Just make sure we will not have any conflict.

nguyenkims:
Nice spot! They are duplicated tickets indeed, I just removed the other one. 

We haven't planned for this ticket yet, a ticket is planned when it's moved to the "Nearer term" column.

developStorm:
Thanks! Since we already have tickets now, I'll close this issue to avoid duplicates.

PR #159, opened 2020-05-05
Implement WebAuthn

developStorm:
#157


nguyenkims:
Thanks very much for the PR! Currently, we don't have a Yubikey with us (due to lockdown) and can't test this feature. I'm ordering a Yubikey and will test the PR as soon as I receive the device.

developStorm:
Oh! You can also test it by this extension: https://github.com/google/virtual-authenticators-tab

Although it will have no prompts when requesting authentication (this will auto permit all requests), it should be enough to test the core functions.

static/assets/js/vendors/webauthn.js
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) 2017 Duo Security, Inc. All rights reserved.
```
nguyenkims, line 1: do you know if this file can be installed via `npm`? Even if it's quite small, updating it via npm is a bit easier :)

static/assets/js/vendors/base64.js
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2017 Duo Security, Inc. All rights reserved.
```
nguyenkims, line 1: same question for this one, can it be installed via `npm`?

app/dashboard/templates/dashboard/setting.html
```diff
@@ -85,18 +85,32 @@
...
     <!-- END change name & profile picture -->
 
+    <div class="card">
+      <div class="card-body">
+        <div class="card-title">Security Key (WebAuthn)</div>
+        <div class="mb-3">
+          You can secure your account by linking either your FIDO-supported physical key such as Yubikey, Google Titan,
+          or a device with appropriate hardware to your account.
+        </div>
+        {% if current_user.fido_uuid is none %}
```
nguyenkims, line 95: This check can be replaced by `{% if not current_user.fido_uuid %}` I think

app/dashboard/views/fido_setup.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class FidoTokenForm(FlaskForm):
+    sk_assertion = HiddenField("sk_assertion", validators=[validators.DataRequired()])
+
+
+@dashboard_bp.route("/fido_setup", methods=["GET", "POST"])
+@login_required
+def fido_setup():
+    if current_user.fido_uuid is not None:
```
nguyenkims, line 25: this check can be shortened by `if current_user.fido_uuid`, this saves some characters :)

app/dashboard/views/fido_setup.py
```diff
@@ -0,0 +1,87 @@
...
+        flash("You have already registered your security key", "warning")
+        return redirect(url_for("dashboard.index"))
+
+    fido_token_form = FidoTokenForm()
+
+    rp_id = urlparse(SITE_URL).hostname
```
nguyenkims, line 31: As `rp_id` never changes, should it be a constant in this module?

app/dashboard/views/fido_setup.py
```diff
@@ -0,0 +1,87 @@
...
+        flash("Security key has been activated", "success")
+
+        return redirect(url_for("dashboard.index"))
+    
+    # Prepare infomation for key registration process
+    fido_uuid = str(uuid.uuid4())
+    challenge = secrets.token_urlsafe(32)
+
+    credential_create_options = webauthn.WebAuthnMakeCredentialOptions(
+        challenge, 'Simple Login', rp_id, fido_uuid,
```
nguyenkims, line 73: the service name is actually `SimpleLogin` (without space between "Simple" and "Login").

app/models.py
```diff
@@ -134,6 +134,12 @@ class User(db.Model, ModelMixin, UserMixin):
         db.Boolean, nullable=False, default=False, server_default="0"
     )
 
+    # Fields for WebAuthn
+    fido_uuid = db.Column(db.String(), nullable=True, unique=True)
+    fido_credential_id = db.Column(db.String(), nullable=True, unique=True)
+    fido_pk = db.Column(db.String(), nullable=True, unique=True)
+    fido_sign_count = db.Column(db.Integer(), nullable=True)
```
nguyenkims, line 141: Do you know how `fido_sign_count` would be used?

app/auth/views/fido.py
```diff
@@ -0,0 +1,106 @@
...
+    user_id = session.get(MFA_USER_ID)
+
+    # user access this page directly without passing by login page
+    if not user_id:
+        flash("Unknown error, redirect back to main page", "warning")
+        return redirect(url_for("auth.login"))
+
+    user = User.get(user_id)
+
+    if not (user and (user.fido_uuid is not None)):
```
nguyenkims, line 35: `(user.fido_uuid is not None)` can be replaced by `user.fido_uuid` :)

app/auth/views/fido.py
```diff
@@ -0,0 +1,106 @@
...
+        flash("Only user with security key linked should go to this page", "warning")
+        return redirect(url_for("auth.login"))
+
+    fido_token_form = FidoTokenForm()
+
+    next_url = request.args.get("next")
+
+    rp_id = urlparse(SITE_URL).hostname
```
nguyenkims, line 43: as `rp_id` is also used in fido setup, I wonder maybe it makes sense to put it in `config.py` as a parameter (that is computed from `URL`

app/auth/views/fido.py
```diff
@@ -0,0 +1,106 @@
+import json
+import secrets
+import webauthn
+from app.config import URL as SITE_URL
```
nguyenkims, line 4: same comment as in `fido_setup`, any reason to change the import name here?

app/auth/views/fido.py
```diff
@@ -0,0 +1,106 @@
...
+        return redirect(url_for("auth.login"))
+
+    fido_token_form = FidoTokenForm()
+
+    next_url = request.args.get("next")
+
+    rp_id = urlparse(SITE_URL).hostname
+
+    webauthn_user = webauthn.WebAuthnUser(
+            user.fido_uuid, user.email, user.name, False,
```
nguyenkims, line 46: I wonder if user changes email, would this affect the `WebAuthnUser` here?

app/auth/views/fido.py
```diff
@@ -0,0 +1,106 @@
...
+            SITE_URL,
+            uv_required=False
+        )
+
+        is_webauthn_verified = False
+        try:
+            new_sign_count = webauthn_assertion_response.verify()
+            is_webauthn_verified = True
+        except Exception as e:
+            flash('Key verification failed. Error: {}'.format(e), "warning")
```
nguyenkims, line 73: Does this return the full error to the user? In this case, maybe it's more secured and user-friendly to do a `LOG.error()` here and returns only user-friendly user error message.

app/dashboard/views/fido_setup.py
```diff
@@ -0,0 +1,87 @@
+import uuid
+import json
+import secrets
+import webauthn
+from app.config import URL as SITE_URL
```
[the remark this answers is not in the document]
developStorm: I think `URL` is not a clear enough variable name. It can refer to any arbitrary URL. So I want to make sure I and other maintainers don't get confused about it.

However the `URL` is a naming convention in this project, so I'm fine if you wish me to change it to `URL`

static/assets/js/vendors/webauthn.js
developStorm: I don't think so. [They](https://github.com/duo-labs/py_webauthn/) didn't publish this as an npm package but as a pip one.

An alternative is using [Github's solution](https://github.com/github/webauthn-json). Unfortunately, I found this project after finished all these works so I didn't use it in the current version. This one may not fully compatible with the previously mentioned python package, so probably need some extra works. 

Let me know how you wish to handle this!

app/dashboard/templates/dashboard/setting.html
developStorm: The reason that I am checking if it's `None` is to emphasize its type. `fido_uuid` could possibly be `None` or `str`, using `not` will allow some unintended behavior like if I set `fido_uuid` to `''` this is also a `False` but not `None`.

app/dashboard/views/fido_setup.py
developStorm: Same as [this](https://github.com/simple-login/app/pull/159#discussion_r421367986)

app/dashboard/views/fido_setup.py
developStorm: Make sense, I'll change it

app/dashboard/views/fido_setup.py
```diff
@@ -0,0 +1,87 @@
...
+        current_user.fido_uuid = fido_uuid
+        current_user.fido_sign_count = fido_credential.sign_count
+        current_user.fido_credential_id = str(fido_credential.credential_id, "utf-8")
+        db.session.commit()
+
+        flash("Security key has been activated", "success")
+
+        return redirect(url_for("dashboard.index"))
+    
+    # Prepare infomation for key registration process
```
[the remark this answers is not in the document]
developStorm: Resolved

static/assets/js/vendors/base64.js
developStorm: Same as [this](https://github.com/simple-login/app/pull/159#discussion_r421366256)

app/dashboard/views/fido_setup.py
developStorm: Sorry for this mistake 😓

app/models.py
developStorm: I think it's to prevent the replay attack.
https://github.com/duo-labs/py_webauthn/blob/18c837594059bdcca354e3154231ebd468d74b41/webauthn/webauthn.py#L1089

app/auth/views/fido.py
developStorm: Same as [this](https://github.com/simple-login/app/pull/159#discussion_r421367986)

app/auth/views/fido.py
developStorm: Putted.

app/dashboard/views/fido_setup.py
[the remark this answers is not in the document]
developStorm: No longer valid due to https://github.com/simple-login/app/pull/159#discussion_r421343108

app/auth/views/fido.py
developStorm: No longer valid due to https://github.com/simple-login/app/pull/159#discussion_r421343108

app/auth/views/fido.py
developStorm: According to the [W3C](https://www.w3.org/TR/webauthn/#dom-publickeycredentialentity-name), it is "intended only for display". So should be fine. The actual identifier will be fido_uuid.

app/auth/views/fido.py
developStorm: 0052dad13ecb1e4cfe28034ef42e22aa019a6150

developStorm:
@nguyenkims Hi, I think all comments were replied/resolved now. Let me know if you have any other concerns!

app/auth/views/fido.py
```diff
@@ -0,0 +1,111 @@
+import json
```
nguyenkims, line 1: Probably a nitpicking thing here but the imports are not sorted and sometimes unused imports are not removed :). The imports are usually sorted by first built-in libs, then external libs (i.e. those installed via `pip`) and lastly in-project modules. If you use Pycharm the "Optimize Imports" can do the job or https://github.com/timothycrosley/isort seems to be a good option.

app/auth/views/fido.py
```diff
@@ -0,0 +1,111 @@
...
+            sk_assertion = json.loads(fido_token_form.sk_assertion.data)
+        except Exception as e:
+            flash("Key verification failed. Error: Invalid Payload", "warning")
+            return redirect(url_for("auth.login"))
+
+        challenge = session["fido_challenge"]
+        credential_id = sk_assertion["id"]
+
+        webauthn_assertion_response = webauthn.WebAuthnAssertionResponse(
+            webauthn_user, sk_assertion, challenge, SITE_URL, uv_required=False
```
nguyenkims, line 65: The `SITE_URL` is not unresolved here.

app/auth/views/fido.py
```diff
@@ -0,0 +1,111 @@
...
+    # Handling POST requests
+    if fido_token_form.validate_on_submit():
+        try:
+            sk_assertion = json.loads(fido_token_form.sk_assertion.data)
+        except Exception as e:
+            flash("Key verification failed. Error: Invalid Payload", "warning")
+            return redirect(url_for("auth.login"))
+
+        challenge = session["fido_challenge"]
+        credential_id = sk_assertion["id"]
```
nguyenkims, line 62: `credential_id` is not used in the code I believe

app/auth/views/fido.py
```diff
@@ -0,0 +1,111 @@
...
+        is_webauthn_verified = False
+        try:
+            new_sign_count = webauthn_assertion_response.verify()
+            is_webauthn_verified = True
+        except Exception as e:
+            LOG.error(f"An error occurred in WebAuthn verification process: {e}")
+            flash("Key verification failed.", "warning")
+
+        if is_webauthn_verified:
+            user.fido_sign_count = new_sign_count
```
nguyenkims, line 77: `new_sign_count` can be unresolved here if an exception happens on line 70

app/auth/views/login_utils.py
```diff
@@ -14,7 +14,15 @@ def after_login(user, next_url):
     If user enables MFA: redirect user to MFA page
     Otherwise redirect to dashboard page if no next_url
     """
-    if user.enable_otp:
+    if user.fido_uuid is not None:
```
nguyenkims, line 17: (Optional) As `user.fido_uuid is not None` is frequently used, I wonder if it makes sense to a method like `User.fido_enabled` to wrap this logic. It seems quite useless for now but in the next step, we'll add something like `User.can_use_fido` as a way to beta-test this feature so the `fido_enabled()` can be written as:

```
def fido_enabled(self) -> bool:
     if not self.can_use_fido: return False
     if self.fido_uuid is not None: return True
     return False
```

app/dashboard/views/fido_setup.py
```diff
@@ -0,0 +1,95 @@
...
+        except Exception as e:
+            flash("Key registration failed. Error: Invalid Payload", "warning")
+            return redirect(url_for("dashboard.index"))
+
+        fido_uuid = session["fido_uuid"]
+        challenge = session["fido_challenge"]
+
+        fido_reg_response = webauthn.WebAuthnRegistrationResponse(
+            RP_ID,
+            SITE_URL,
```
nguyenkims, line 44: `SITE_URL` is unresolved here

app/auth/views/fido.py
developStorm: If an exception happened, `is_webauthn_verified` will be `False` so we should be safe

static/assets/js/vendors/webauthn.js
nguyenkims: Do you think the Github one offers the same features as the Duo-Labs one? It seems to me that packaging on Github repo is proper and seems to be more maintained. The Duo-Labs one looks more like a proof-of-concept for me but that might be just a feeling.

app/auth/views/fido.py
nguyenkims: True. I wonder if this block can be written as `try/catch/else` and so `is_webauthn_verified` can be removed. Something like 

```
        try:
            new_sign_count = webauthn_assertion_response.verify()
            is_webauthn_verified = True
        except Exception as e:
            LOG.error(f"An error occurred in WebAuthn verification process: {e}")
            flash("Key verification failed.", "warning")
        else:
            user.fido_sign_count = new_sign_count
            db.session.commit()
            del session[MFA_USER_ID]

            login_user(user)
            flash(f"Welcome back {user.name}!", "success")

            # User comes to login page from another page
            if next_url:
                LOG.debug("redirect user to %s", next_url)
                return redirect(next_url)
            else:
                LOG.debug("redirect user to dashboard")
                return redirect(url_for("dashboard.index"))
```

app/auth/views/login_utils.py
developStorm: f7e33202421df53ca6f7ee09736c51153b3038a3

app/auth/views/fido.py
developStorm: Hmm, I don't think VS Code has this feature built-in. Could you handle this instead?

app/dashboard/views/fido_setup.py
developStorm: e4895b52a00ca2bb9810858d33199392652d0d36

app/auth/views/fido.py
developStorm: e4895b52a00ca2bb9810858d33199392652d0d36

app/auth/views/fido.py
developStorm: b0c39635a5e5f7ddde85869085266269f87c944e

app/auth/views/fido.py
developStorm: 2290a90b09468bebbf2e043bfbc6bbfef7fad446

static/assets/js/vendors/webauthn.js
developStorm: I'm not sure if they encoded the data in the same way...

nguyenkims: I see, let's stay with the current way then.

app/auth/views/fido.py
nguyenkims: Yes I will run Pycharm auto-import tool after merging this PR.

nguyenkims:
LGTM!

PR #161, opened 2020-05-07
CI Improvements

developStorm:
- Upgrade `actions/checkout` and `actions/setup-python`
- Cache pip files, presumably accelerate CI
- Create a test only CI for pull requests. https://github.com/simple-login/app/pull/159#pullrequestreview-406985398 
- Tag images by their branch bf45a2bff51c7d32916ba54d7e488761e0fa4660


.github/workflows/tester.yml
```diff
@@ -0,0 +1,40 @@
+name: Run tests
```
nguyenkims, line 1: I wonder why do we need this one? It seems in the `main.yml` we already have all steps right?

.github/workflows/main.yml
```diff
@@ -12,37 +12,52 @@ jobs:
...
+          pip install -r requirements.txt
+
+      - name: Test with pytest
+        run: |
+          pip install pytest
+          pytest
+
+      - name: Extract tag name
+        shell: bash
+        run: echo ::set-output name=tag::$(echo ${GITHUB_REF##*/})
```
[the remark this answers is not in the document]
developStorm: It is here: https://stackoverflow.com/a/58034787/13459000

You will probably notice that there is a comment discussed an issue of this with `Gitflow-style`, however, I intentionally ignored that issue since Docker doesn't allow `/` appears in the tag name (so if we fix that probably it will fail the CI).

.github/workflows/main.yml
```diff
@@ -12,37 +12,52 @@ jobs:
...
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      - name: Test formatting
+        run: |
+          pip install black
+          black --check .
```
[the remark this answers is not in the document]
developStorm: That will be a good idea with Nodejs since it can configure `dev` environment packages. 

However, I think `black` is currently cached. The hash of `requirement.txt` passed to the `action/cache` is just to flush the cache when there is some requirement changed. Even we didn't specify `black` in the requirement, it should be cached since it will still go to the cached folder.  

The only issue may be if there is no requirement upgraded but `black`, action will not cache that version of `black`. But this sounds not a big deal.

.github/workflows/tester.yml
developStorm: This one is particular for all `pull requests`. Therefore contributors know if there is an issue before actually merged.

developStorm: I can probably make this better I think by using
```
if: github.event_name == 'pull_request'
```

developStorm: Implemented in 20b22d66f6b1684e3ceec58e0db43be14e7d9339. This could be even better if Github supports to check another job's status before running a job. Currently, an extra ~25 seconds are used to build two Docker images which aren't actually used.

developStorm:
Just find that Docker has an [official action](https://github.com/docker/build-push-action) now and solved that tag name issue. Going to replace with that.

nguyenkims:
@developStorm I just created this PR https://github.com/simple-login/app/pull/169 based on your suggestion :).

nguyenkims:
I close this PR then.

PR #164, opened 2020-05-07
Enforce SPF

SibrenVasse:
Block emails to reverse alias if sender (soft-)fails SPF check.

I added pyspf to requirements.in and ran pip-compile. This caused a lot of changes in requirements.txt


app/models.py
```diff
@@ -1116,6 +1116,7 @@ class Mailbox(db.Model, ModelMixin):
     user_id = db.Column(db.ForeignKey(User.id, ondelete="cascade"), nullable=False)
     email = db.Column(db.String(256), unique=True, nullable=False)
     verified = db.Column(db.Boolean, default=False, nullable=False)
+    force_spf = db.Column(db.Boolean, default=True, nullable=False)
```
nguyenkims, line 1119: I think we can add the `server_default` option here so alembic can auto-generate the SQL migration correctly.

app/dashboard/templates/dashboard/mailbox_detail.html
```diff
@@ -64,6 +64,35 @@ <h1 class="h3">{{ mailbox.email }}
...
+      {% if spf_available %}
+      <div class="card">
+        <form method="post">
+          <input type="hidden" name="form-name" value="force-spf">
+
+          <div class="card-body">
+            <div class="card-title">
+              Enforce SPF
+              <div class="small-text">
+                Block emails to reverse alias if sender is not validated by SPF,
```
nguyenkims, line 76: I think we'll need to explain more about SPF as this concept is not easy to understand for non-tech users :). But don't worry, I can take care of that later :).

email_handler.py
```diff
@@ -465,6 +468,23 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
...
+    if ENFORCE_SPF_AVAILABLE and mailb.force_spf:
+        if msg['X-SimpleLogin-Client-IP']:
+            r = spf.check2(i=msg['X-SimpleLogin-Client-IP'], s=envelope.mail_from.lower(), h=None)
+            # TODO: Handle temperr case (e.g. dns timeout)
+            # only an absolute pass, or no SPF policy at all is 'valid'
+            if r[0] not in ["pass", "none"]:
+                LOG.d(
+                    "SPF validation failed for %s (reason %s)", mailbox_email, r[0],
+                )
+                # TODO: Send warning to mailbox
```
nguyenkims, line 481: In case of impersonation attacks, we can send **too many** alert emails to user. I think it's relatively safe to just return an error here as if the email is sent from the user's mailbox, it should receive an error anyway.

email_handler.py
```diff
@@ -465,6 +468,23 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
         handle_bounce(contact, alias, msg, user, mailbox_email)
         return False, "550 SL E6"
 
+    mailb: Mailbox = Mailbox.get_by(email=mailbox_email)
+    if ENFORCE_SPF_AVAILABLE and mailb.force_spf:
+        if msg['X-SimpleLogin-Client-IP']:
+            r = spf.check2(i=msg['X-SimpleLogin-Client-IP'], s=envelope.mail_from.lower(), h=None)
```
nguyenkims, line 474: The library `spf` seems to be created by Scott Kitterman, the same guy behind `policyd-spf` so it should be quite battle tested :).

README.md
```diff
@@ -375,6 +375,9 @@ smtpd_sender_restrictions =
     reject_unknown_sender_domain,
     permit
 
+body_checks = pcre:/etc/postfix/body_checks
```
nguyenkims, line 378: I think this setup could be extracted in a separate doc to reduce the number of steps necessary to have SimpleLogin running.

SibrenVasse: If we omit this from the README, I think ENFORCE_SPF should default to false in example.env, right?

SibrenVasse:
For future reference:

```
body_checks = pcre:/etc/postfix/body_checks
smtpd_client_restrictions = pcre:/etc/postfix/client_headers.pcre
```

Create the `/etc/postfix/client_headers.pcre` file with the following content.
```
/^([0-9a-f:.]+)$/ prepend X-SimpleLogin-Client-IP: $1
```

Create the `/etc/postfix/body_checks` file with the following content.
```
/^X-SimpleLogin-Client-IP:/    IGNORE
```

SibrenVasse:
I think I resolved all your comments.

SibrenVasse:
Fixed formatting as CI was complaining.

nguyenkims:
The PR looks good! 

I'm going to merge the PR and add some small changes like
- add a doc section to setup SPF
- do not return 550 as this generates a bounce email to the original mailbox. Which is a bit annoying for users if they are target of an attack.
- update the migration script so it points to the current latest migration.

PR #170, opened 2020-05-08
🐛 WebAuthn bug fixes

developStorm:
- User may not have a name - didn't realize this since the local fake data have it as default
- user_verification should be discouraged to work on iOS

### DIFF
--- a/app/dashboard/views/fido_setup.py
+++ b/app/dashboard/views/fido_setup.py
@@ -81,9 +81,10 @@ def fido_setup():
         RP_ID,
         fido_uuid,
         current_user.email,
-        current_user.name,
+        current_user.name if current_user.name else current_user.email,
         False,
         attestation="none",
+        user_verification="discouraged",
     )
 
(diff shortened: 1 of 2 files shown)


app/dashboard/views/fido_setup.py
```diff
@@ -81,9 +81,10 @@ def fido_setup():
         RP_ID,
         fido_uuid,
         current_user.email,
-        current_user.name,
+        current_user.name if current_user.name else current_user.email,
         False,
         attestation="none",
+        user_verification="discouraged",
```
nguyenkims, line 87: I'm not sure to see how this is linked to iOS ... Can you elaborate or send a link about this topic?

developStorm: https://developers.yubico.com/WebAuthn/Supporting_U2F_or_FIDO2_Security_Keys_on_iOS_or_iPadOS/Known_Issues.html

nguyenkims: I see, thanks!

PR #171, opened 2020-05-09
Session cookie flags.

SibrenVasse:
Currently firefox displays this warning:
```
Cookie “slapp” will be soon rejected because it has the “sameSite” attribute set to
“none” or an invalid value, without the “secure” attribute.
To know more about the “sameSite“ attribute,
read https://developer.mozilla.org/docs/Web/HTTP/Cookies
```
- This commit sets the 'secure' flag to true when the base url starts with https
- Set the SameSite attribute to 'strict'
- Set Flask session_protection to 'strong' (see https://flask-login.readthedocs.io/en/latest/#session-protection)

### DIFF
--- a/app/extensions.py
+++ b/app/extensions.py
@@ -5,4 +5,5 @@ from flask_sqlalchemy import SQLAlchemy
 
 db = SQLAlchemy()
 login_manager = LoginManager()
+login_manager.session_protection = "strong"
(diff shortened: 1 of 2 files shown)


nguyenkims:
Thanks, it's better to set these attributes indeed!

PR #174, opened 2020-05-09
Spf update

nguyenkims:


nguyenkims:
@SibrenVasse This PR adds some minor things to the enforce-SPF feature:

- make sure to remove "X-SimpleLogin-Client-IP" header in forward phase. 
- make sure DNS errors don't affect the reply phase
- Hide the enforce-SPF toggle: as this option is "on" by default and I don't really see why an user would want to disable it, I think we can hide it for now.
- Add a doc on how to enable this option for self-hosting

SibrenVasse:
As a user, If you're using a mailbox in a domain you have no control over you cannot choose to use send your mail from a different SMTP server, even if the SPF is not configured like that.
We are absolutely violating the spec here! That's why I added the toggle.

nguyenkims:
@SibrenVasse that makes sense indeed. 

I think if this option is "on" by default, we need to inform user when the SPF fails so they can decide to turn it off if necessary.

We also need to make sure that user's mailbox is not bombed with the alert emails in case they are the target of a repetitive attack.

I'll create another PR to address that.

PR #177, opened 2020-05-10
Dark mode implementation

doanguyen:
This PR

* Using CSS variable as a solution
* The new dark mode CSS is now live in another file so that upgrading
dashboard.css doesn't affect the dark mode itself
* Used a naive dark mode controller by client javascript. No persistence
states is stored in the backend at the moment

image
image


nguyenkims:
Thanks for the PR, I test the PR locally and everything works perfectly 👍 !

PR #178, opened 2020-05-10
Multiple mailboxes

nguyenkims:
https://trello.com/c/DPpJwo92/27-an-alias-can-have-multiple-recipents

This PR allows an alias to have several mailboxes:

- emails sent to an alias are forwarded to all owning mailboxes
- any mailbox can reply or send email from this alias

This PR is quite long as it needs to take into account
- bounce email
- unauthorized email (email sent from a mailbox that is not an owner of the alias)


nguyenkims:
@SibrenVasse Would be great if you can have a look at this PR :)

SibrenVasse:
I've been playing around with it for a while, it seems to be working as expected!

nguyenkims:
@SibrenVasse Cool, thanks for testing it!

SibrenVasse:
On Mailbox deletion: Mailbox.delete() in models.py#L1206
All Aliases corresponding to this Mailbox are moved to trash. I think this should only happen when the alias has only this one mailbox.

nguyenkims:
@SibrenVasse Nice catch (as usual :))! I just pushed a commit that would only put an alias that has the deleted mailbox as its *only* mailbox to the global trash.

PR #181, opened 2020-05-12
Auto activate WebAuthn authentication

developStorm:


nguyenkims:
@developStorm thanks! I tested it locally and works great 👍

PR #183, opened 2020-05-13
API Error handling for 404 and 500

nguyenkims:


SibrenVasse:
Can you also add the handler for 405?

```python
@api_bp.app_errorhandler(405)
def wrong_method(e):
    return jsonify(error="Method not allowed"), 405
```

nguyenkims:
@SibrenVasse good idea! I just added it.

PR #186, opened 2020-05-14
Frontend improvements

SibrenVasse:
1. **When using the dark theme, the page flashes white before loading javascript (my eyes!).**
This commit stores the dark theme setting in a cookie, instead of the local store. This way the server can send the HTML with the correct data attribute, preventing the page from starting to draw the light theme for a few frames.
2. **Small space fix**
Before
image
After
image
3. **Change pagination style to fit better in the dark theme.**
Before
image
After
image
4. **Fix last_page check on index and alias_log.**
This correctly disables the next/previous button on those pages. (They currently allow you to go past the boundaries)
5. **Set input background to white for the light theme (looks better I think)**
Before
image
After
image

### DIFF
--- a/static/assets/js/core.js
+++ b/static/assets/js/core.js
@@ -104,17 +104,4 @@ $(document).ready(function() {
       });
     });
   }
-
-  /** Dark mode controller */
-  if (store.get('dark-mode') === true) {
-    document.documentElement.setAttribute('data-theme', 'dark')
-  }
-  $('[data-toggle="dark-mode"]').on('click', function () {
-    if (store.get('dark-mode') === true) {
-      store.set('dark-mode', false);
-      return document.documentElement.setAttribute('data-theme', 'light')
-    }
-    store.set('dark-mode', true)
-    document.documentElement.setAttribute('data-theme', 'dark')
-  })
 });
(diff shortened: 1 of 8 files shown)


static/assets/js/core.js
```diff
@@ -12,6 +12,31 @@ let hexToRgba = function(hex, opacity) {
   return 'rgba(' + rgb.r + ', ' + rgb.g + ', ' + rgb.b + ', ' + opacity + ')';
 };
 
+let setCookie = function(name, value, days) {
```
nguyenkims, line 15: This `core.js` is actually from the template (https://tabler.io/) so I think it'd be better to put custom code in a different file and include it in `base.html`.

static/assets/js/core.js
```diff
@@ -106,15 +131,15 @@ $(document).ready(function() {
   }
 
   /** Dark mode controller */
-  if (store.get('dark-mode') === true) {
```
nguyenkims, line 109: This should be in a different file indeed, I must have missed it in the previous PR (the one that introduced the dark mode)

templates/base.html
```diff
@@ -1,7 +1,7 @@
 {% from "_formhelpers.html" import render_field, render_field_errors %}
 
 <!doctype html>
-<html lang="en" dir="ltr">
+<html lang="en" dir="ltr" data-theme="{% if request.cookies.get('dark-mode') == 'true' %}dark{% endif %}">
```
nguyenkims, line 4: That's a clever way to set the theme before the page finishes loading!

PR #190, opened 2020-05-15
Add more From: sender formatting options

SibrenVasse:
0. John Wick - john at wick.com
1. john@wick.com via SimpleLogin (default)
2. John Wick - john(a)wick.com
3. John Wick - john@wick.com


app/models.py
```diff
@@ -95,19 +95,28 @@ def get_url(self, expires_in=3600):
         return s3.get_url(self.path, expires_in)
 
 
-class PlanEnum(enum.Enum):
+class EnumE(enum.Enum):
+    @classmethod
+    def has_value(cls, value: int) -> bool:
+        return value in cls._value2member_map_
```
nguyenkims, line 101: I wonder if we can use this code that doesn't use the private attribute `_value2member_map_`:

```
def has_value(cls, value: int) -> bool:
     try:
           cls(value)
     except ValueError:
           return False
     else:
           return True
```

app/models.py
```diff
@@ -171,10 +180,12 @@ def fido_enabled(self) -> bool:
...
-    # Use the "via" format for sender address, i.e. "name@example.com via SimpleLogin"
-    # If False, use the format "Name - name at example.com"
-    use_via_format_for_sender = db.Column(
-        db.Boolean, default=True, nullable=False, server_default="1"
+    # Specify the format for sender address
+    # John Wick - john at wick.com  -> 0
+    # john@wick.com via SimpleLogin -> 1
+    # John Wick - john@wick.com     -> 2
+    sender_format = db.Column(
+        db.Integer, default="0", nullable=False, server_default="1"
```
nguyenkims, line 188: should we use the same value for `default` and `server_default` here?

app/models.py
```diff
@@ -853,21 +864,31 @@ def website_send_to(self):
...
+            or user.sender_format == SenderFormatEnum.VIA.value
+        ):
             new_name = f"{self.website_email} via SimpleLogin"
-        else:
+        elif user.sender_format == SenderFormatEnum.AT.value:
             name = self.name or ""
             new_name = (
                 name + (" - " if name else "") + self.website_email.replace("@", " at ")
             ).strip()
+        elif user.sender_format == SenderFormatEnum.C_AT.value:
```
nguyenkims, line 887: It misses a final `else` here and *in theory* `new_name` can be undefined at the end of the method.
I wonder if we can slightly refactor the code into 2 cases here: `user is not None` and `user is None` . Within the first case, we construct the `new_name` according to user preference.

app/models.py
```diff
@@ -853,21 +864,31 @@ def website_send_to(self):
...
+            name = self.name or ""
+            new_name = (
+                name + (" - " if name else "") + self.website_email.replace("@", "@")
```
nguyenkims, line 890: the `replace` here can be removed I think :)

app/models.py
```diff
@@ -171,10 +180,12 @@ def fido_enabled(self) -> bool:
 
     profile_picture = db.relationship(File, foreign_keys=[profile_picture_id])
 
-    # Use the "via" format for sender address, i.e. "name@example.com via SimpleLogin"
-    # If False, use the format "Name - name at example.com"
-    use_via_format_for_sender = db.Column(
```
nguyenkims, line 176: If we remove this column without migrating to the new `sender_format`, user who has set their preference here will see their choice "disappeared". I guess we could have 2 choices here:

1. modify the migration script to migrate `use_via_format_for_sender` data to `sender_format`
2. keep `use_via_format_for_sender` but deprecate it by `sender_format`.

In my opinion 1. is cleaner but also more difficult :).

app/models.py
SibrenVasse: Do we want to try the different at symbol? https://trello.com/c/30iiOVXS
https://unicode-table.com/en/search/?q=at

nguyenkims: Yeah we definitely can! But I think we need to make sure user understands that these are 2 different characters on the Settings page to avoid surprises.

SibrenVasse: So the reasoning is that some email clients do not display the symbol correctly? Do you have an example so I can do some tests?

nguyenkims: Ah the main reason is sometimes people can try to copy/paste the sender email address that would not work because the character is not the usual `@`.

app/models.py
SibrenVasse: That's essentially what's already there: (It's essentially a switch statement, with the default at the top)
```
if (no user) or (unknown format code) or (format = VIA)
---
elif (format = AT)
---
elif (format = C_AT)
```

SibrenVasse:
I think I resolved all your comments.

migrations/versions/2020_051515_5cad8fa84386_.py
```diff
@@ -0,0 +1,45 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column("users", "use_via_format_for_sender", server_default=None)
+    op.alter_column(
+        "users",
+        "use_via_format_for_sender",
+        new_column_name="sender_format",
+        type_=sa.Integer(),
+        postgresql_using="use_via_format_for_sender::integer",
```
nguyenkims, line 28: Nice!

nguyenkims:
The PR is OK for me! 

There's a conflict in `test_models` though, could you please resolve the conflict and then we can merge the PR?

SibrenVasse:
Done!

PR #196, opened 2020-05-17
Move api error handling to global error handler

SibrenVasse:
Move API error handling. Dashboard currently does not show the HTML error page. Only return json if request URL starts with /api/.

Test with:
```
http https://app.simplelogin.io/dashboard/not_existing
http https://app.simplelogin.io/api/not_existing
```


nguyenkims:
Perfect, thanks for the fix!

PR #200, opened 2020-05-20
Fix reverse alias replacement

SibrenVasse:
1. Replacing reverse alias needs to be done before DKIM signing.
2. Replace with contact.website_email, not alias.email

### DIFF
--- a/email_handler.py
+++ b/email_handler.py
@@ -555,6 +555,16 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
         envelope.rcpt_options,
     )
 
+    # replace the "ra+string@simplelogin.co" by the alias in the email body
+    # as this is usually included in when replying
+    if user.replace_reverse_alias:
+        payload = (
+            msg.get_payload()
+            .encode()
+            .replace(reply_email.encode(), contact.website_email.encode())
+        )
+        msg.set_payload(payload)
+
     if alias_domain in ALIAS_DOMAINS:
         add_dkim_signature(msg, alias_domain)
     # add DKIM-Signature for custom-domain alias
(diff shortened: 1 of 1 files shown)


email_handler.py
```diff
@@ -555,6 +555,16 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
         envelope.rcpt_options,
     )
 
+    # replace the "ra+string@simplelogin.co" by the alias in the email body
+    # as this is usually included in when replying
+    if user.replace_reverse_alias:
```
nguyenkims, line 560: The DKIM signature is generated using the `from` and `to` header (https://github.com/simple-login/app/blob/93d972df09088ffa6f199fa3d3c7e4b91552765f/app/config.py#L125-L124) so we don't have to move the email body modification before DKIM I think.

email_handler.py
```diff
@@ -555,6 +555,16 @@ def handle_reply(envelope, smtp: SMTP, msg: Message, rcpt_to: str) -> (bool, str
...
+        payload = (
+            msg.get_payload()
+            .encode()
+            .replace(reply_email.encode(), contact.website_email.encode())
```
nguyenkims, line 564: Nice spot! I don't know why I didn't realize that it doesn't make sense to replace the `reply_email` by the `alias` before ...

email_handler.py
nguyenkims: If this section is moved back to just before `smtp.sendmail` then we can still use the `msg_raw.replace(reply_email.encode(), alias.email.encode())` and don't need to use the payload I guess.

SibrenVasse: Example DKIM header (with modified body -> body hash fail) :
```
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/simple; d=simplelogin.co;
  i=@simplelogin.co; q=dns/txt; s=dkim; t=1590054495; h=from : to;
  bh=jVYQQAujVte8BGVQLkJI30GCN7JIjo8vGtbkR1fjboU=;
  b=TR+sgcw2NQ6qehsC1QnFma5m7M0oG/dvvqHQTR1xAj7/s0H6TVo4BA78wHTh0xb7mPCnL
  VdM4/Rwn3zgP8bW/pGiSLmWT6DCbB5f6HfcKOUjAG0KD2V5cB2P1/zw1dxtMuCwk0kEtA/w
  Kiv4deWvKg2Akhy5mOgIhyR96hN4+/E=
Signed-by: reply+rzaivkdavxjwdaldbbzbwwczdhhiav@simplelogin.co
Expected-Body-Hash: cI8Ka+2r7E5IHcORorQyhZqfR6Ay7G4xzCv+DKqjaqg=

DKIM-Result: fail (wrong body hash: jVYQQAujVte8BGVQLkJI30GCN7JIjo8vGtbkR1fjboU=)
```
1. field 'bh' is the body hash
2. field 'b' is the signature of the email content (body + headers specified by the 'h' field).

nguyenkims: I see, DKIM does indeed use the body in the hash computation!

PR #203, opened 2020-05-22
Optional 'remember MFA' for browser

SibrenVasse:
**1. Use MFA cookie to whitelist browser for MFA for 30 days.**
In this PR only for TOTP. Maybe also add this for FIDO?
**2. Prevent OTP replay attacks by invalidating last token**
See security checklist on https://pyotp.readthedocs.io/en/latest/
```
Deny replay attacks by rejecting one-time passwords that have been used by the client
(this requires storing the most recently authenticated timestamp, OTP,
or hash of the OTP in your database, and rejecting the OTP when a match is seen)
```
**3. Add autofocus to email field on login screen**
**4. Remove cookies from browser on logout**
**5. Fix flash() calls in mfa.py**


SibrenVasse:
Also a thing from the checklist is rate limiting the authentication endpoints.
This can be done very easily with https://flask-limiter.readthedocs.io/en/stable/#
What do you think?

app/auth/views/logout.py
```diff
@@ -8,4 +8,9 @@
 def logout():
     logout_user()
     flash("You are logged out", "success")
-    return redirect(url_for("auth.login"))
+    response = make_response(redirect(url_for("auth.login")))
+    response.delete_cookie("slapp")
```
nguyenkims, line 12: I wonder if we can use `session.pop` here to make sure to remove all cookies except the `_flashes` which is used for `flash`?

https://stackoverflow.com/a/51395792/1428034

app/auth/views/mfa.py
```diff
@@ -33,27 +44,49 @@ def mfa():
     otp_token_form = OtpTokenForm()
     next_url = request.args.get("next")
 
+    if request.cookies.get("mfa"):
```
nguyenkims, line 47: I wonder if we can use `session` here instead of using cookies directly?

app/auth/views/mfa.py
```diff
@@ -33,27 +44,49 @@ def mfa():
...
+        browser = MfaBrowser.get_by(token=request.cookies.get("mfa"))
+        if browser and not browser.is_expired():
+            login_user(user)
+            flash(f"Welcome back {user.name}!", "success")
+            # Redirect user to correct page
+            return redirect(next_url or url_for("dashboard.index"))
+
+        MfaBrowser.delete(browser.token)
```
nguyenkims, line 55: I might have missed something ... I don't understand why do we need to delete the `mfa_browser` here?

app/auth/views/mfa.py
```diff
@@ -33,27 +44,49 @@ def mfa():
...
     if otp_token_form.validate_on_submit():
         totp = pyotp.TOTP(user.otp_secret)
 
-        token = otp_token_form.token.data
+        token = otp_token_form.token.data.replace(" ", "")
 
-        if totp.verify(token):
+        if totp.verify(token) and user.last_otp != token:
             del session[MFA_USER_ID]
+            user.last_otp = token.replace(" ", "")
```
nguyenkims, line 64: the `replace(" ", "")` seems to be called twice, it's already called on `otp_token_form.token.data.replace(" ", "")`

app/auth/views/mfa.py
```diff
@@ -33,27 +44,49 @@ def mfa():
...
-            flash(f"Welcome back {user.name}!")
+            flash(f"Welcome back {user.name}!", "success")
+
+            # Redirect user to correct page
+            response = make_response(redirect(next_url or url_for("dashboard.index")))
+
+            if otp_token_form.remember.data:
+                browser = MfaBrowser.create_new(user=user)
+                db.session.commit()
+                response.set_cookie(
```
nguyenkims, line 76: Same as another comment above, I wonder if we could use `session` here?

app/auth/views/mfa.py
```diff
@@ -33,27 +44,49 @@ def mfa():
...
 
-        if totp.verify(token):
+        if totp.verify(token) and user.last_otp != token:
             del session[MFA_USER_ID]
+            user.last_otp = token.replace(" ", "")
+            db.session.commit()
 
             login_user(user)
-            flash(f"Welcome back {user.name}!")
+            flash(f"Welcome back {user.name}!", "success")
```
nguyenkims, line 68: nice catch! I always forgot to add the second param when using `flash()` ...

app/models.py
```diff
@@ -497,6 +498,38 @@ def generate_oauth_client_id(client_name) -> str:
     return generate_oauth_client_id(client_name)
 
 
+class MfaBrowser(db.Model, ModelMixin):
+    user_id = db.Column(db.ForeignKey(User.id, ondelete="cascade"), nullable=False)
+    token = db.Column(db.String(64), default=False, nullable=False)
```
nguyenkims, line 503: should the `token` be `unique`?

app/models.py
```diff
@@ -497,6 +498,38 @@ def generate_oauth_client_id(client_name) -> str:
...
+    expires = db.Column(ArrowType, default=False, nullable=False)
+
+    user = db.relationship(User)
+
+    @classmethod
+    def create_new(cls, user, token_length=64) -> "MfaBrowser":
+        return MfaBrowser.create(
+            user_id=user.id,
+            token=random_string(token_length),
```
nguyenkims, line 512: if the `token` is unique, we might need to do a loop until a unique token is found, something like https://github.com/simple-login/app/blob/ec5821d89bb4bcf9ea70966745b6731cdabb3012/app/models.py#L1174

app/auth/views/mfa.py
SibrenVasse: The session contains server session variables. Cookies are associated with the request object.

app/auth/views/logout.py
SibrenVasse: The session contains server session variables. Cookies are associated with the request object.

app/models.py
SibrenVasse: Yes but for a 64 character lowercase string (26^64 possibilities), this is **very** unlikely. Like a [1 in 196 duovigintillion chance](https://www.wolframalpha.com/input/?i=1%C3%B7%28%281%C3%B7%2826%5E64%29%29%C3%972%5E64%29) for a collision even when you generate 2^64 tokens.

nguyenkims: I agree :) but just **in theory** this can happen right? The `random_string` is also pseudorandom so in a very unlikely case, someone could affect the random generation, putting the data in an inconsistent state.

app/auth/views/mfa.py
nguyenkims: I'm not sure to understand, can you elaborate please? Why can't we use `session.get()` here?

app/models.py
SibrenVasse: Haha yeah absolutely true ;)

app/auth/views/mfa.py
SibrenVasse: The client sends the 'mfa' cookie to the server. To access this value, we need the request object.  Session is used to store MFA_USER_ID for example.

SibrenVasse:
Are you sure you're testing it right? If you use the logout button, the 'mfa' cookie is also deleted. (A user action logout should in remove all user cookies)
Easiest way to test is to manually remove the 'slapp' cookie in the browser dev tools and login again.

nguyenkims:
> Are you sure you're testing it right? If you use the logout button, the 'mfa' cookie is also deleted. (A user action logout should in remove all user cookies)
> Easiest way to test is to manually remove the 'slapp' cookie in the browser dev tools and login again.

Here's how I tested the feature locally:
- setup TOTP
- log out, login again, make sure to tick the "remember for 30 days" 
- logout, login again: I'm still asked for TOTP token. When "remember for 30 days" is ticked, I should not be asked again right?

SibrenVasse:
Yeah so in logout.py:L13 the mfa cookie is deleted. So if you use the 'logout' button to exit the session, it won't work. Only if the session expires (i.e. the cookie slapp is no longer valid) and the user then goes to login, it does work.

So in short:
1. Setup TOPT
2. log out, login again, make sure to tick the "remember for 30 days"
3. close browser
4. login

Or we can just remove logout.py:L13 and it will always work.

nguyenkims:
@SibrenVasse I see! I also understand now why you need to use the cookie directly. The PR is good for me, I will merge it when we merge the multiple FIDO keys PR.

SibrenVasse:
I'll also add the feature to the FIDO path.
Shall I take a look into the rate limiting?

nguyenkims:
> I'll also add the feature to the FIDO path.
> Shall I take a look into the rate limiting?

Great, thanks! I think the rate-limiting can be in a separate PR.

SibrenVasse:
> > I'll also add the feature to the FIDO path.
> > Shall I take a look into the rate limiting?
> 
> Great, thanks! I think the rate-limiting can be in a separate PR.

Yeah I'll do that.

PR #209, opened 2020-05-24
Implement rate limiting

SibrenVasse:
Based on #199 #203. 
Implementation for rate limiting, specified limit is per IP.
- Puts limit on endpoints like username/password authentication and mfa endpoints.
- Only counts 'bad requests', e.g. failed login attempt.
- I arbitrarily chose 10/min limit. Any thoughts? As it's only failed requests, we could possibly lower this. More complex limits are possible:
```
@limiter.limit("100/day;10/hour;1/minute")
```

Nginx snippet
```
server {
    server_name  app.mydomain.com;

    location / {
        proxy_pass http://localhost:7777;
        proxy_set_header X-Forwarded-For $remote_addr;
    }
}
```

### DIFF
--- a/app/api/views/auth.py
+++ b/app/api/views/auth.py
@@ -1,9 +1,8 @@
-import random
-
 import facebook
 import google.oauth2.credentials
 import googleapiclient.discovery
-from flask import jsonify, request
+import random
+from flask import jsonify, request, g
 from flask_cors import cross_origin
 from itsdangerous import Signer
 
@@ -17,13 +16,16 @@ from app.email_utils import (
     send_email,
     render,
 )
-from app.extensions import db
+from app.extensions import db, limiter
 from app.log import LOG
 from app.models import User, ApiKey, SocialAuth, AccountActivation
 
 
 @api_bp.route("/auth/login", methods=["POST"])
 @cross_origin()
+@limiter.limit(
+    "10/minute", deduct_when=lambda r: hasattr(g, "deduct_limit") and g.deduct_limit
+)
 def auth_login():
     """
     Authenticate user
@@ -52,6 +54,8 @@ def auth_login():
     user = User.filter_by(email=email).first()
 
     if not user or not user.check_password(password):
+        # Trigger rate limiter
+        g.deduct_limit = True
         return jsonify(error="Email or password incorrect"), 400
     elif not user.activated:
         return jsonify(error="Account not activated"), 400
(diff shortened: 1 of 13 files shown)


developStorm:
It's ok for giving webauthn more quotas since it's nearly impossible to be brute-force attacked

SibrenVasse:
Yeah but is there really a good reason to fail the webauthn check 10 times in a minute? How would you manage that.

developStorm:
Oops I were thought the "more complex example" is the actual rule (1/min)😅10/min sounds good!

app/api/views/auth.py
```diff
@@ -164,6 +178,9 @@ def auth_activate():
 
 @api_bp.route("/auth/reactivate", methods=["POST"])
 @cross_origin()
+@limiter.limit(
```
nguyenkims, line 181: the `limiter` is currently not used in this endpoint right?

app/auth/views/forgot_password.py
```diff
@@ -12,6 +13,9 @@ class ForgotPasswordForm(FlaskForm):
 
 
 @auth_bp.route("/forgot_password", methods=["GET", "POST"])
+@limiter.limit(
+    "10/minute", deduct_when=lambda r: hasattr(g, "forgot_fail") and g.forgot_fail
```
nguyenkims, line 17: `forgot_fail` should be renamed to `deduct_limit` here right :) ?

templates/error/429.html
```diff
@@ -0,0 +1,15 @@
+{% extends "error.html" %}
+
+{% block error_name %}
+  429
+{% endblock %}
+
+{% block error_description %}
+  Slow down there, pardner!
```
nguyenkims, line 8: 😅

server.py
```diff
@@ -369,6 +371,13 @@ def forbidden(e):
         else:
             return render_template("error/403.html"), 403
 
+    @app.errorhandler(429)
+    def forbidden(e):
+        if request.path.startswith("/api/"):
```
nguyenkims, line 376: can we add a `log.error` here please? We got paged when an error happens so this allows us to see either an attack is going on or is this rather a false positive.

app/limiter.py
```diff
@@ -0,0 +1,14 @@
+from flask import request
+from flask_limiter import Limiter
+from flask_limiter.util import get_remote_address
+
+# Setup rate limit facility
+limiter = Limiter(key_func=get_remote_address)
```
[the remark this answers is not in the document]
SibrenVasse: Ah yes, that's the location I was looking for but failed to find ;)

PR #212, opened 2020-05-26
Fix reverse alias replacement multipart message

SibrenVasse:


nguyenkims:
@SibrenVasse I tested different scenarios and the reverse-alias is correctly replaced 👍! I'm going to merge and deploy this PR.

PR #221, opened 2020-06-02
Create POST /api/v3/alias/custom/new

nguyenkims:


README.md
```diff
@@ -823,7 +823,7 @@ For ex:
 }
 ```
 
-#### POST /api/v2/alias/custom/new
+#### POST /api/v3/alias/custom/new
```
nguyenkims, line 826: @ntnhon this new endpoint version uses `mailboxes` as input to create multiple mailboxes for the new alias.

ntnhon: @nguyenkims I would suggest saying `mailbox_ids` instead of `mailboxes` to be consistent with `PUT /api/aliases/:alias_id`

nguyenkims: Good idea! I just updated the PR.

PR #260, opened 2020-08-12
Improvements for Self Hosting & More

FabioWidmer:
Footer is now more dynamic and uses environment variables. When adding a new contact the "Sent to" format uses now the user defined sender format. I assumed it is possible to use the @ character in the format. If not please let me know.

Second commit makes all emails ready for self hosting instances. Also links to the App Store, Google Play Store, Chrome Webstore and Firefox Addons have been shortened.


FabioWidmer:
Fixed the checks. Sorry for the multiple commits. Would appreciate if this could also get merged with the next release.

README.md
```diff
@@ -7,11 +7,11 @@
 [SimpleLogin](https://simplelogin.io) | Privacy-First Email Forwarding and Identity Provider Service
 ---
 <p>
-<a href="https://chrome.google.com/webstore/detail/simplelogin-protect-your/dphilobhebphkdjbpfohgikllaljmgbn">
+<a href="https://chrome.google.com/webstore/detail/dphilobhebphkdjbpfohgikllaljmgbn">
```
nguyenkims, line 10: Do you know what is the difference between these 2 urls? The new one looks definitely shorter and more correct 👍

app/auth/templates/auth/login.html
```diff
@@ -49,8 +49,10 @@
     Don't have an account yet? <a href="{{ url_for('auth.register') }}">Sign up</a>
   </div>
 
+  {% if (GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET) or (GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET) or (FACEBOOK_CLIENT_ID and FACEBOOK_CLIENT_SECRET) %}
```
nguyenkims, line 52: to make these variables available in the Jinja2 templates, I think you'd need to include them here: https://github.com/simple-login/app/blob/5636b7ba32c84d0ff1a7cb1717664f684689a93c/server.py#L449

app/auth/templates/auth/social.html
```diff
@@ -12,20 +12,26 @@
       <div class="card-title text-center">Social login
       </div>
 
+      {% if GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET %}
```
nguyenkims, line 15: good call 👌

app/dashboard/templates/dashboard/api_key.html
```diff
@@ -22,11 +22,11 @@ <h1 class="h3"> API Key
...
         An API key is used by the SimpleLogin browser extensions.
         <br><br>You can install the official SimpleLogin browser extensions through the following links:
-        <a href="https://chrome.google.com/webstore/detail/simplelogin-extension/dphilobhebphkdjbpfohgikllaljmgbn"
+        <a href="https://chrome.google.com/webstore/detail/dphilobhebphkdjbpfohgikllaljmgbn"
            target="_blank" rel="noopener">Chrome<i class="fe fe-external-link"></i></a>,
-        <a href="https://addons.mozilla.org/en-GB/firefox/addon/simplelogin/"
+        <a href="https://addons.mozilla.org/firefox/addon/simplelogin/"
            target="_blank" rel="noopener">Firefox<i class="fe fe-external-link"></i></a> &
-        <a href="https://apps.apple.com/us/app/simplelogin/id1494051017?mt=12&fbclid=IwAR0M0nnEKgoieMkmx91TSXrtcScj7GouqRxGgXeJz2un_5ydhIKlbAI79Io"
+        <a href="https://apps.apple.com/app/id1494051017"
```
nguyenkims, line 29: the new URL does look a lot better!

app/models.py
```diff
@@ -1078,7 +1078,18 @@ def website_send_to(self):
...
         name = self.name
+        email = self.website_email
+
+        if (
+            not user
+            or not SenderFormatEnum.has_value(user.sender_format)
+            or user.sender_format == SenderFormatEnum.AT.value
+        ):
+            email = email.replace("@", " at ")
+        elif user.sender_format == SenderFormatEnum.A.value:
```
nguyenkims, line 1091: should we take into account `SenderFormatEnum.VIA` and `SenderFormatEnum.FULL` as well?

templates/emails/base.html
```diff
@@ -448,8 +448,8 @@
       <table class="email-content" width="100%" cellpadding="0" cellspacing="0" role="presentation" style="width: 100%; -premailer-width: 100%; -premailer-cellpadding: 0; -premailer-cellspacing: 0; margin: 0; padding: 0;">
         <tr>
           <td class="email-masthead" style="word-break: break-word; font-family: Helvetica, Arial, sans-serif; font-size: 16px; text-align: center; padding: 25px 0;" align="center">
-            <a href="https://simplelogin.io" class="f-fallback email-masthead_name" style="color: #A8AAAF; font-size: 16px; font-weight: bold; text-decoration: none; text-shadow: 0 1px 0 white;">
-              <img src="https://simplelogin.io/logo.png" style="width: 150px; margin: auto">
+            <a href="{{URL}}" class="f-fallback email-masthead_name" style="color: #A8AAAF; font-size: 16px; font-weight: bold; text-decoration: none; text-shadow: 0 1px 0 white;">
```
nguyenkims, line 451: I guess you might want to use `LANDING_PAGE_URL` here.

templates/emails/com/welcome.html
```diff
@@ -13,7 +13,7 @@ <h1 style="margin-top: 0; color: #333333; font-size: 22px; font-weight: bold; te
 {% block content %}
   {{ render_text("My name is Son. I’m the founder of SimpleLogin and I wanted to be the first to welcome you on board.") }}
 
-  {{ render_text('To better secure your account, I recommend enabling Multi-Factor Authentication (MFA) on your <a href="https://app.simplelogin.io/dashboard/setting/#totp">Setting page</a>.') }}
+  {{ render_text('To better secure your account, I recommend enabling Multi-Factor Authentication (MFA) on your <a href="' ~ URL ~ '/dashboard/setting/#totp">Setting page</a>.') }}
```
nguyenkims, line 16: The `render_text` is slowly replaced by the `{% call text() %}` which is much nicer and avoids the escape-hell :), an example is here:
https://github.com/simple-login/app/blob/9ddb8ff2d4134516255e9eca86bfb250d5f395d2/templates/emails/transactional/spf-fail.html#L12

I wonder if we should take the opportunity to use the `{% call text() %}` here?

README.md
FabioWidmer: It just doesn't include the name of the extension. Usually it gets included to make it more user friendly. But for this use case I think it makes more sense because you could change the name of the extension.

app/dashboard/templates/dashboard/api_key.html
FabioWidmer: It doesn't include all the localisation and Facebook parameters.

app/models.py
FabioWidmer: I already thought about those two: `FULL` is already supported as it doesn't modify the email. `VIA` also uses the `@` but the text "via SimpleLogin" doesn't make sense for me here. What do you think?

templates/emails/com/welcome.html
FabioWidmer: I agree - `{% call text() %}` is better. But there are many more code parts which have to be updated. I will do that in another pull request (as soon as this one is merged) to not overload this one.

app/auth/templates/auth/login.html
FabioWidmer: Thanks for letting me know (I'm new to Jinja). I included them now. Is it possible to include all `env` variables by default? From my point of view this would make sense.

FabioWidmer:
@nguyenkims Thanks for the comments! I made a new commit with a few changes. Let me know what you think regarding my other comments.

templates/emails/base.html
```diff
@@ -448,8 +448,8 @@
       <table class="email-content" width="100%" cellpadding="0" cellspacing="0" role="presentation" style="width: 100%; -premailer-width: 100%; -premailer-cellpadding: 0; -premailer-cellspacing: 0; margin: 0; padding: 0;">
         <tr>
           <td class="email-masthead" style="word-break: break-word; font-family: Helvetica, Arial, sans-serif; font-size: 16px; text-align: center; padding: 25px 0;" align="center">
-            <a href="https://simplelogin.io" class="f-fallback email-masthead_name" style="color: #A8AAAF; font-size: 16px; font-weight: bold; text-decoration: none; text-shadow: 0 1px 0 white;">
-              <img src="https://simplelogin.io/logo.png" style="width: 150px; margin: auto">
+            <a href="{{LANDING_PAGE_URL}}" class="f-fallback email-masthead_name" style="color: #A8AAAF; font-size: 16px; font-weight: bold; text-decoration: none; text-shadow: 0 1px 0 white;">
+              <img src="{{URL}}/static/logo.png" style="width: 150px; margin: auto">
```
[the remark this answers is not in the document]
nguyenkims: Yeah .svg support for email clients is quite hit-and-miss

nguyenkims (approved):
LGTM! Thanks for taking the time to make this PR 👍

PR #271, opened 2020-08-24
Enable TLS on Postfix submission

nguyenkims:


README.md
```diff
@@ -428,6 +428,21 @@ Finally, restart Postfix
...
 
+#### [Optional] Enable TLS on Postfix submission
+
+In case your Postfix server is on another server, it's recommended to enable TLS on Postfix submission to 
+secure the connection between SimpleLogin email handler and Postfix.
+
+This can be enabled by adding those lines at the end of `/etc/postfix/master.cf`
+
+```
+submission inet n       -       y       -       -       smtpd
```
FabioWidmer, line 439: According to this [article](https://www.faqforge.com/linux/how-to-enable-port-587-submission-in-postfix/) you only need to uncomment this line or add it if it doesn't exist. Not sure what the best practice would be for the lines below.

README.md
```diff
@@ -428,6 +428,21 @@ Finally, restart Postfix
 sudo systemctl restart postfix
 ```
 
+#### [Optional] Enable TLS on Postfix submission
+
+In case your Postfix server is on another server, it's recommended to enable TLS on Postfix submission to 
+secure the connection between SimpleLogin email handler and Postfix.
+
+This can be enabled by adding those lines at the end of `/etc/postfix/master.cf`
```
FabioWidmer, line 436: And setting the `POSTFIX_SUBMISSION_TLS` to `true` in the .env file.

README.md
nguyenkims: Some Postfix default config file uses a slightly different version of these lines. So I think it's less error-prone to add new lines instead of uncommenting existing lines.

README.md
nguyenkims: Good call! I just updated the PR to include that and move Postfix-TLS instructions to its own file.

PR #293, opened 2020-09-29
Add dot after DKIM

TheLastProject:
When I was setting up SimpleLogin (cool stuff by the way), DKIM went wrong because the record ended up being converted to `dkim._domainkey.simplelogin.co.example.com` due to the lack to ending dot.

This PR proposes adding a `.` at the end of the Value in the DNS screen to hopefully prevent others from making my mistake.

Please note: I haven't tested this myself, haven't bothered setting up the Docker container yet.


nguyenkims:
@TheLastProject thanks! I tested and it works :). Btw the app can run without Docker. Docker is mostly used for deployment.

PR #297, opened 2020-10-01
Dedupe email address

TheLastProject:
I sometimes get emails where the sender's name is the same as the email address. This causes the same info to be displayed twice. This code deduplicates it.

So, if you get an email like `From: example@example.com <example@example.com>` it should only display as `example@example.com <reply_address>` instead of `example@example.com - example@example.com <reply_address>`.

I hope this code and change is useful to you. Feel free to tell me if you want any changes.


nguyenkims:
The PR looks perfect, thanks!

PR #298, opened 2020-10-02
Show icon if address was autogenerated

TheLastProject:
Choosing an icon for this is HARD. I ended up going with `fa-inbox` because it was generated because of an auto-generated email and I couldn't find any better icon on https://fontawesome.com/icons?d=gallery&m=free. However, if you have a suggestion, I'll gladly change it.

image


nguyenkims:
Thanks for the PR! Choosing the right icon is indeed hard, I tried to look for an icon that suggests the auto-creation property but couldn't find any ...

PR #302, opened 2020-10-05
Support random suffix for personal domains

TheLastProject:
I ended up adding it to the new alias page, it just felt most logical to me:

image

To make this work I had to rewrite some permission checking though. Please please please triple-check my code. I am pretty sure I did it right and the tests still pass, but this is a rather critical code-path I had to edit that could cause trouble if I made a mistake.


app/dashboard/views/custom_alias.py
```diff
@@ -174,41 +175,29 @@ def verify_prefix_suffix(user, alias_prefix, alias_suffix) -> bool:
 
     # make sure alias_suffix is either .random_word@simplelogin.co or @my-domain.com
     alias_suffix = alias_suffix.strip()
-    if alias_suffix.startswith("@"):
```
nguyenkims, line 177: This code part is indeed overly complex as it needs to take into account the self-hosting case ... I think the new code misses the check on the `alias_domain_prefix` though as we want to make sure users can't inject any string here. I wrote a small gist for it: https://github.com/simple-login/app/commit/43857495b3da2c99e5ee51f0a52e51bc3d55f9d8, let me know if this makes sense

nguyenkims (commented):
I tested the code locally and it works well 👌. Just left a comment on the prefix-suffix check.

I also think it's better to add a new settings that is defaulted to `False`, and when set to `True`, add the random string to custom domain (as you did in `available_suffixes`). The reason for that is users like to see the random suffix "disappear" with their own domains so it might be confusing for them to see this random bit still there :).

Btw it's also better to start off from the latest master as I have moved from `pip` to `poetry` recently and upgraded a lot of packages :).

app/dashboard/views/custom_alias.py
TheLastProject: It mostly makes sense but you seem to have removed the DISABLE_ALIAS_PREFIX check? Why is that?

TheLastProject:
Adding a new setting makes sense, yeah. I'll add that.

app/dashboard/views/custom_alias.py
nguyenkims: Yeah when re-writing this part, it seems that there's no need to distinguish the `DISABLE_ALIAS_PREFIX` case. I guess this case has been added as a hotfix back then 🤔.

TheLastProject:
I think I have now addressed all of your concerns :)

nguyenkims (approved):
LGTM! Thanks for making this PR!

PR #309, opened 2020-10-12
add docker-compose examples and apache2 setup to self host instructions

fareszr:
hi, after setting up simplelogin my self, i have noticed some stuff that would be useful for new comers, so i added them into the readme file.

## top changes:
- added docker-compose examples
- added apache2 example setup
- added 'local_recipient_maps' = note in postfix
- added check for pgsql availability with `postconf -m` command
- added postgres command to make all accounts premium by default.
- added an option to add your own words for random_aliases

note the apache file is built on default-ssl.conf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ sudo apt-get install -y postfix postfix-pgsql -y
 
 Choose "Internet Site" in Postfix installation window then keep using the proposed value as *System mail name* in the next window.
 
-to make sure pgsql is working correctly, run `postconf -m` and make sure its listed.
+to make sure pgsql is working correctly, run `postconf -m` and make sure `pgsql` is listed.
 
 Replace `/etc/postfix/main.cf` with the following content. Make sure to replace `mydomain.com` by your domain.
 
@@ -536,7 +536,47 @@ sudo docker run --rm \
     simplelogin/app:3.2.2 python init_app.py
 ```
 
-Now, it's time to run the `webapp` container!
+now it time to setup the `webapp` and `email handler` containers
+
+for docker-compose add this to your file:
+
+```
+    simplelogin:
+        image: simplelogin/app:3.2.2
+        container_name: sl-app
+        ports:
+            - "8009:7777"
+        restart: unless-stopped
+        volumes:
+        # this is to add more words to create more unique aliases
+        # - ./words.txt:/code/local_data/words.txt
+        - ./sl:/sl
+        - ./sl/upload:/code/static/upload
+        - ./simplelogin.env:/code/.env
+        - ./dkim.key:/dkim.key
+        - ./dkim.pub.key:/dkim.pub.key
+
+    simplelogin-email_handler:
+        image: simplelogin/app:3.2.2
+        container_name: sl-email
+        command: python email_handler.py
+        ports:
+            - "20381:20381"
+        restart: unless-stopped
+        volumes:
+        # this is to add more words to create more unique aliases you have to create a word.txt file.
+        # - ./words.txt:/code/local_data/words.txt
+        - ./sl:/sl
+        - ./sl/upload:/code/static/upload
+        - ./simplelogin.env:/code/.env
+        - ./dkim.key:/dkim.key
+        - ./dkim.pub.key:/dkim.pub.key
+
+```
+
+for docker-cli:
+
+`webapp` container:
 
 ```bash
 sudo docker run -d \
@@ -552,7 +592,7 @@ sudo docker run -d \
     simplelogin/app:3.2.2
 ```
 
-Next run the `email handler`
+`email handler` container:
 
 ```bash
 sudo docker run -d \
--- a/README.md
+++ b/README.md
@@ -350,6 +350,8 @@ sudo apt-get install -y postfix postfix-pgsql -y
 
 Choose "Internet Site" in Postfix installation window then keep using the proposed value as *System mail name* in the next window.
 
+to make sure pgsql is working correctly, run `postconf -m` and make sure its listed.
+
 Replace `/etc/postfix/main.cf` with the following content. Make sure to replace `mydomain.com` by your domain.
 
 ```
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ sudo docker run -d \
     postgres:12.1
 ```
 
-if you want to use docker-compose you can paste this inside your docker-compose.yml file
+if you want to use docker-compose you can paste this inside your `docker-compose.yml` file
 
 ```
 version: "3"
@@ -368,6 +368,10 @@ readme_directory = no
 # See http://www.postfix.org/COMPATIBILITY_README.html -- default to 2 on
 # fresh installs.
 compatibility_level = 2
+
+# if you have an issue local recipient tables uncomment the nextline to 
+# remove the table completely
+# local_recipient_maps = 
 
 # TLS parameters
 smtpd_tls_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
--- a/README.md
+++ b/README.md
@@ -305,6 +305,33 @@ sudo docker run -d \
     postgres:12.1
 ```
 
+if you want to use docker-compose you can paste this inside your docker-compose.yml file
+
+```
+version: "3"
+#connecting to the pre made sl-network.
+networks:
+    default:
+        external:
+            name: sl-network
+
+# all file paths are relative to the docker-compose file
+
+services:
+    postgres:
+        image: postgres:12.1
+        container_name: sl-db
+        ports:
+            - "5432:5432"
+        volumes:
+            - ./sl/db:/var/lib/postgresql/data
+        environment:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -503,7 +536,66 @@ sudo docker run --rm \
...
+```
+    simplelogin:
+        image: simplelogin/app:3.2.2
+        container_name: sl-app
+        ports:
+            - "7777:7777"
+        restart: unless-stopped
+        volumes:
+        # this is to add more words to create more unique aliases
+        # - ./words.txt:/code/local_data/words.txt
```
nguyenkims, line 552: by default, words are retrieved from `local_data/words_alpha.txt` which contains more words than the `local_data/words.txt` which is rather used for local development.

README.md
```diff
@@ -503,7 +536,66 @@ sudo docker run --rm \
...
+        - ./sl:/sl
+        - ./sl/upload:/code/static/upload
+        - ./simplelogin.env:/code/.env
+        - ./dkim.key:/dkim.key
+        - ./dkim.pub.key:/dkim.pub.key
+
+    simplelogin-email_handler:
```
nguyenkims, line 559: I think we can use the same container name here, for example, this one would be `sl-email`

README.md
```diff
@@ -503,7 +536,66 @@ sudo docker run --rm \
...
+        # - ./words.txt:/code/local_data/words.txt
+        - ./sl:/sl
+        - ./sl/upload:/code/static/upload
+        - ./simplelogin.env:/code/.env
+        - ./dkim.key:/dkim.key
+        - ./dkim.pub.key:/dkim.pub.key
+
+```
+
+**note**: the words.txt volume is optional and it allows you to add your own custom words for random aliases.
```
nguyenkims, line 577: To keep the self-hosting instructions small, I think it's better to move this section to a dedicated file in the `docs/` folder.

README.md
```diff
@@ -519,7 +611,7 @@ sudo docker run -d \
     simplelogin/app:3.2.2
 ```
 
-Next run the `email handler`
+`email handler` container:
```
nguyenkims, line 614: Why do we need to remove the `Next run the` here :)?

README.md
```diff
@@ -563,12 +655,53 @@ sudo systemctl reload nginx
 
 At this step, you should also setup the SSL for Nginx. [Certbot](https://certbot.eff.org/lets-encrypt/ubuntuxenial-nginx) can be a good option if you want a free SSL certificate.
 
+
+### Apache
```
nguyenkims, line 659: I think it's better to move this section to a dedicated file in `docs/`. Maybe we can add also how to get a valid certificate here as well as modern browsers will make it hard for users to authorize `/etc/ssl/certs/ssl-cert-snakeoil.pem` certificate.

README.md
```diff
@@ -563,12 +655,53 @@ sudo systemctl reload nginx
...
 ### Enjoy!
 
 If all of the above steps are successful, open http://app.mydomain.com/ and create your first account!
 
-By default, new accounts are not premium so don't have unlimited alias. To make your account premium,
-please go to the database, table "users" and set "lifetime" column to "1" or "TRUE".
+By default, new accounts are **not premium** so don't have unlimited alias. To make all accounts premium by default,
+use the following commands:
+
+`sudo docker exec -it sl-db psql -U myuser simplelogin`
```
nguyenkims, line 698: I wonder if we can group these into a single command, something like the below. I haven't tested the command so this is just for idea :).

```
echo "ALTER TABLE users ALTER COLUMN lifetime SET DEFAULT TRUE;" | sudo docker exec -it sl-db psql -U myuser simplelogin
```

nguyenkims (changes requested):
Overall the PR looks good 👍, thanks for creating it! 

I think we can change the section order so everything can be run in docker-compose. 
For example, we can re-organize the sections like this:
- Postfix 
- Prepare config file (`.env` file)
- Docker: to install Docker
- Docker-compose: a single docker-compose file that takes care of all the steps: `sl-migration`, `sl-init`, `sl-app`, `sl-email`

This way all the Docker commands can be replaced by a single `docker-compose.yml` file. 

What do you think?

README.md
fareszr: interesting, it worked for me and the aliases used the new custom words
in the example.env is says it uses words.txt:
https://github.com/simple-login/app/blob/2d9abe55a448a5ece9c4b84e2eebf365c168f18e/example.env#L110

doesn't the example file use the default value?

README.md
fareszr: alright we can create like a new page for more stuff after initial setup

it should also include how create ssl certificates since its not **essential** for the app to work

README.md
fareszr: i changed because you should add the 2 containers at the same time.

README.md
fareszr: echo "ALTER TABLE users ALTER COLUMN lifetime SET DEFAULT TRUE;" | sudo docker exec -it sl-db psql -U myuser simplelogin
responds with:
the input device is not a TTY

README.md
fareszr: fixed in https://github.com/simple-login/app/pull/309/commits/7a0bf6a681ede2ae877c1fb45b7d02d87cb2ae5c

README.md
fareszr: fixed in https://github.com/simple-login/app/pull/309/commits/324640fbae1ddc6d3a2a8ace89e7935065555d8e

README.md
fareszr: should i move all the webserver setup to the extras page or just the ssl setup ?

fareszr: instead of creating cusom guides to generate ssl certificates with certbot  why not just use digtaloceans's guides?

nginx: https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-18-04

apache2:https://www.digitalocean.com/community/tutorials/how-to-secure-apache-with-let-s-encrypt-on-ubuntu-18-04

fareszr:
> Docker-compose: a single docker-compose file that takes care of all the steps: `sl-migration`, `sl-init`, `sl-app`, `sl-email`

for the temporary containers i think its easier to just use normal docker commands

i found the second answer [here](https://stackoverflow.com/questions/56071196/how-can-you-automatically-run-and-remove-a-container-in-docker-compose) to be the best option for docker-compose.

but since you only need it once just use the normal docker commands its easier, if you are planning on transitioning the entire guide into docker-compose 2 commands aren't much.

README.md
fareszr: fixed https://github.com/simple-login/app/pull/309/commits/b374f5620091577d57b5d04322aaaf76aa291ad6

README.md
fareszr: fixed https://github.com/simple-login/app/pull/309/commits/b374f5620091577d57b5d04322aaaf76aa291ad6

nguyenkims:
@fareszr sorry for the late response! I plan to do an important reorganization of the documentation to simplify the self-hosting instructions and will integrate your PR content at this moment. Is this OK for you?

fareszr:
> @fareszr sorry for the late response! I plan to do an important reorganization of the documentation to simplify the self-hosting instructions and will integrate your PR content at this moment. Is this OK for you?

absoultley !

i would love to help!, i can help with setting up mkdocs and themes if you want to, i suggest using the docs.simplelogin.io domain and having all docs there.

nguyenkims:
@fareszr thanks a lot! Will definitely let you know when this refactoring starts!

README.md
pojhm91c7iwk: I think the current guide is useful. As a new comer, it's not helpful when you "point to a link", rather continue installation steps helps with the installation.

fareszr: the thing is the instructions are largely the same, its just adding the certificate to postifx.

pojhm91c7iwk: > the thing is the instructions are largely the same, its just adding the certificate to postifx.

For someone with experience, yes. For someone who's new would have difficulties separating the "related" code lines to execute. Just curious, why is it a big deal? I mean, we can keep the current guide as it is...

fareszr: The documention in general needs an upgarde, the less stuff to mainatain outside the focus of the project the better.

shalak:
Hello! Are there any updates on this one? Are those docker-compose files still valid for use with master simple-login builds?

fariszr:
> Hello! Are there any updates on this one? Are those docker-compose files still valid for use with master simple-login builds?

my instance is still based on the compose file in this PR.
and its working fine, however there have been changes in the readme, and due to the repo being forced push, there are merge conflicts.

fariszr:
Fixed all the merge conflicts, the PR is ready to be merged again.

akash07k:
Is there any update on this? inclusion of docker-compose instructions will greatly simplify the installation and replication.

misaligar:
Any updates?

springcomp:
@meminens
You might be interested in:
https://github.com/springcomp/self-hosted-simplelogin

PR #330, opened 2020-11-20
Fix Firefox Add-on link

herrboyer:


nguyenkims (approved):
Thanks for fixing this!

PR #331, opened 2020-11-21
Readme spelling correction.

brainynai:
"loose" is the opposite of "tight", "lose" is the opposite of "acquire".


nguyenkims:
Thanks for fixing this error!

PR #339, opened 2020-12-05
Measure test coverage

herrboyer:
Hi,

I was wondering how to contribute to simple login and thought about writing tests but without coverage reports its not easy to find where to start !

This PR:

* [x] installs [pytest-cov](https://pytest-cov.readthedocs.io) to plug [coverage.py](https://coverage.readthedocs.io) to `pytest`
* [x] adds a coverage report in the terminal to know the global % of coverage and which lines are not covered yet
* [x] adds an html report which is uploaded as a workflow artefact in GitHub CI
* [x] makes unit tests fail if coverage gets under 60% (current coverage is ~61%)

This last point is of course debatable, as test coverage is not an absolute indicator of code quality, but, IMHO, 60% is a bit low, and it would be nice to reach 80% one day.

PR #342, opened 2020-12-07
Linting

herrboyer:
Hi,

Following my comment here https://github.com/simple-login/app/pull/339#issuecomment-739481330 this PR fixes some linting problems (unused imports, etc.) and installs flake8 as a pre-commit hook and a workflow step.


nguyenkims (approved):
Thank you very much for going through all the files and making the necessary changes!

PR #365, opened 2021-02-02
refactor(build): remove dependencies in production

doanguyen:
Remove the `vim`, `telnet` on production image

Other things to consider:

- Using smaller size Python image: python:3.7-alpine/python:3.7-slim
- Run as non-root user, ref: https://github.com/hexops/dockerfile#run-as-a-non-root-user


nguyenkims (approved):
LGTM👌!

PR #366, opened 2021-02-06
refactor(config): remove python evil eval

doanguyen:
Two new functions were added to parse the os environment

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -4,8 +4,10 @@ import socket
 import string
 import subprocess
 from urllib.parse import urlparse
+from typing import List, Tuple, Optional, Any, Callable
 
 from dotenv import load_dotenv
+from ast import literal_eval
 
 SHA1 = subprocess.getoutput("git rev-parse HEAD")
 ROOT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
@@ -18,6 +20,39 @@ def get_abs_path(file_path: str):
         return file_path
     else:
         return os.path.join(ROOT_DIR, file_path)
+
+
+def parse_environment(env_var: str) -> Optional[Any]:
+    """Convert environmental string into Python object
+
+    Args:
+        env_var (str): environment constant, example: '["domain1.org", "domain2.com"]'
+
+    Returns:
+        Any
+    """
+    try:
+        return literal_eval(env_var)
+    except Exception as e:
+        print("This expression is failed to parsed: %s, with reason: %s", (env_var, e))
+
+
+def sl_get_env(key: str, default_factory: Optional[Callable] = None) -> Any:
+    """Convert environment key into Python object
+
+    Args:
+        key (str): environment constant, example: SL_DB
+        default_factory (Callable, optional): returns value if environment is None. Notice: if this argument and environment variable are not specified, this function will raised. Defaults to None.
+
+    Returns:
+        Any: Python object
+    Raises:
+        ValueError: when key value is empty and default_factory is not specified
+    """
+    env = os.getenv(key)
+    if not (env or default_factory):
+        raise ValueError("This env should be specified: %s", key)
+    return parse_environment(env) or default_factory()
 
 
 config_file = os.environ.get("CONFIG")
@@ -35,7 +70,8 @@ COLOR_LOG = "COLOR_LOG" in os.environ
 PROMO_CODE = "SIMPLEISBETTER"
 
 # Debug mode
-DEBUG = os.environ["DEBUG"] if "DEBUG" in os.environ else False
+DEBUG = bool(os.environ.get("DEBUG", default=False))
+
 # Server url
 URL = os.environ["URL"]
 print(">>> URL:", URL)
@@ -50,8 +86,8 @@ SENTRY_FRONT_END_DSN = os.environ.get("SENTRY_FRONT_END_DSN") or SENTRY_DSN
 
 # Email related settings
 NOT_SEND_EMAIL = "NOT_SEND_EMAIL" in os.environ
-EMAIL_DOMAIN = os.environ["EMAIL_DOMAIN"].lower()
-SUPPORT_EMAIL = os.environ["SUPPORT_EMAIL"]
+EMAIL_DOMAIN = os.environ.get("EMAIL_DOMAIN").lower()
+SUPPORT_EMAIL = os.environ.get("SUPPORT_EMAIL")
 SUPPORT_NAME = os.environ.get("SUPPORT_NAME", "Son from SimpleLogin")
 ADMIN_EMAIL = os.environ.get("ADMIN_EMAIL")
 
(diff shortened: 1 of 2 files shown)


app/config.py
```diff
@@ -101,43 +102,40 @@ def get_abs_path(file_path: str):
 # Useful when calling Postfix from an external network
 POSTFIX_SUBMISSION_TLS = "POSTFIX_SUBMISSION_TLS" in os.environ
 
-if "OTHER_ALIAS_DOMAINS" in os.environ:
```
nguyenkims, line 104: Hey unfortunately this value is set with `[` ... e.g. `["domain1", "domain2"]` so this change will create a regression

nguyenkims (commented):
Hey I think using `eval` isn't really risky as it's up to the person who runs the code to use the correct value. The only case using `eval` could be dangerous here is if the server is compromised, which then has more serious consequences then the `eval` risk :)

app/config.py
doanguyen: You're right, current implementation will introduce backward incompatibility. I'm going to write a simple utility function to handle

app/config.py
```diff
@@ -20,6 +22,23 @@ def get_abs_path(file_path: str):
...
+    Args:
+        env_var (str): environment constant, example: '["domain1.org", "domain2.com"]'
+
+    Returns:
+        Any
+    """
+    try:
+        return literal_eval(env_var)
+    except Exception as e:
+        print(
```
nguyenkims, line 37: I think it's better to not catch the error here so we can be informed of any failed configuration ("fail fast" principle) :)

app/config.py
```diff
@@ -20,6 +22,23 @@ def get_abs_path(file_path: str):
...
+    """Convert environment string into Python object
+
+    Args:
+        env_var (str): environment constant, example: '["domain1.org", "domain2.com"]'
+
+    Returns:
+        Any
+    """
+    try:
+        return literal_eval(env_var)
```
nguyenkims, line 35: Didn't know about `literal_eval`, seems to be the one we need indeed!

app/config.py
```diff
@@ -35,7 +54,7 @@ def get_abs_path(file_path: str):
 PROMO_CODE = "SIMPLEISBETTER"
 
 # Debug mode
-DEBUG = os.environ["DEBUG"] if "DEBUG" in os.environ else False
+DEBUG = os.environ.get("DEBUG", default=False)
```
nguyenkims, line 57: This can return a non-boolean value of DEBUG (which we don't want)

app/config.py
```diff
@@ -50,8 +69,8 @@ def get_abs_path(file_path: str):
 
 # Email related settings
 NOT_SEND_EMAIL = "NOT_SEND_EMAIL" in os.environ
-EMAIL_DOMAIN = os.environ["EMAIL_DOMAIN"].lower()
-SUPPORT_EMAIL = os.environ["SUPPORT_EMAIL"]
+EMAIL_DOMAIN = os.environ.get("EMAIL_DOMAIN").lower()
```
nguyenkims, line 72: I think for obligatory variables like `EMAIL_DOMAIN`, it's better to use `[]` to throw an exception (so the app will refuse to start)

app/config.py
nguyenkims: In this case, we can even remove `parse_environment` and just directly use `literal_eval`

app/config.py
```diff
@@ -101,43 +120,36 @@ def get_abs_path(file_path: str):
 # Useful when calling Postfix from an external network
 POSTFIX_SUBMISSION_TLS = "POSTFIX_SUBMISSION_TLS" in os.environ
 
-if "OTHER_ALIAS_DOMAINS" in os.environ:
-    OTHER_ALIAS_DOMAINS = eval(
-        os.environ["OTHER_ALIAS_DOMAINS"]
-    )  # ["domain1.com", "domain2.com"]
-else:
-    OTHER_ALIAS_DOMAINS = []
+OTHER_ALIAS_DOMAINS = parse_environment(
```
nguyenkims, line 123: maybe we can modify `parse_environment` so it accepts the env variable name instead (so `OTHER_ALIAS_DOMAINS` in this case instead of `os.environ.get("OTHER_ALIAS_DOMAINS")` and has a default value when the value isn't there, something like

```python
def parse_environment(env_name: str, default_val):
     if env_name in os.environ: 
          return literal_eval(os.environ[env_name])
     else:
          return default_val
```

nguyenkims (commented):
Hey the `literal_eval` is what we need indeed. Just left some suggestions that I think will make the change retro-compatible with existing installations.

app/config.py
doanguyen: Indeed, but I think the old way has the same problem. I've fixed on the latest push.

app/config.py
doanguyen: It's fixed on the latest push

app/config.py
doanguyen: Thanks, based on your suggestion, I've introduced an additional argument and designed it as a closure instead of fixed value. It will allows us:

- Lazy defined a default value according to our need, ex: a default constant needs other default constant
- Raises when a default constant is not defined, by letting default_factory be None

wdyt?

app/config.py
nguyenkims: Hey I don't see the update here ...

app/config.py
```diff
@@ -101,43 +137,36 @@ def get_abs_path(file_path: str):
...
+EMAIL_SERVERS_WITH_PRIORITY: List[Tuple[int, str]] = sl_get_env(
+    "EMAIL_SERVERS_WITH_PRIORITY", default_factory=list
 )  # [(10, "email.hostname.")]
 
 # these emails are ignored when computing stats
-if os.environ.get("IGNORED_EMAILS"):
-    IGNORED_EMAILS = eval(os.environ.get("IGNORED_EMAILS"))
-else:
-    IGNORED_EMAILS = []
+IGNORED_EMAILS = os.environ.get("IGNORED_EMAILS", default="").split(",")
```
nguyenkims, line 169: should we use `sl_get_env` here :)?

app/config.py
```diff
@@ -196,17 +225,15 @@ def get_abs_path(file_path: str):
     PADDLE_YEARLY_PRODUCT_ID = -1
 
 # Other Paddle product IDS
-if "PADDLE_MONTHLY_PRODUCT_IDS" in os.environ:
-    PADDLE_MONTHLY_PRODUCT_IDS = eval(os.environ["PADDLE_MONTHLY_PRODUCT_IDS"])
-else:
-    PADDLE_MONTHLY_PRODUCT_IDS = []
+PADDLE_MONTHLY_PRODUCT_IDS = os.environ.get(
```
nguyenkims, line 228: should we use `sl_get_env` here :)?

app/config.py
```diff
@@ -196,17 +225,15 @@ def get_abs_path(file_path: str):
...
+    "PADDLE_MONTHLY_PRODUCT_IDS", default=""
+).split(",")
 
 PADDLE_MONTHLY_PRODUCT_IDS.append(PADDLE_MONTHLY_PRODUCT_ID)
 
-if "PADDLE_YEARLY_PRODUCT_IDS" in os.environ:
-    PADDLE_YEARLY_PRODUCT_IDS = eval(os.environ["PADDLE_YEARLY_PRODUCT_IDS"])
-else:
-    PADDLE_YEARLY_PRODUCT_IDS = []
+PADDLE_YEARLY_PRODUCT_IDS = os.environ.get(
```
nguyenkims, line 234: same here :)

app/config.py
```diff
@@ -20,6 +22,39 @@ def get_abs_path(file_path: str):
...
+
+    Returns:
+        Any: Python object
+    Raises:
+        ValueError: when key value is empty and default_factory is not specified
+    """
+    env = os.getenv(key)
+    if not (env or default_factory):
+        raise ValueError("This env should be specified: %s", key)
+    return parse_environment(env) or default_factory()
```
nguyenkims, line 55: I think we can replace `parse_environment` by  `literal_eval` and remove `parse_environment()`, this way the code would be a bit simpler :)

app/config.py
```diff
@@ -20,6 +22,39 @@ def get_abs_path(file_path: str):
...
+    except Exception as e:
+        print("This expression is failed to parsed: %s, with reason: %s", (env_var, e))
+
+
+def sl_get_env(key: str, default_factory: Optional[Callable] = None) -> Any:
+    """Convert environment key into Python object
+
+    Args:
+        key (str): environment constant, example: SL_DB
+        default_factory (Callable, optional): returns value if environment is None. Notice: if this argument and environment variable are not specified, this function will raised. Defaults to None.
```
nguyenkims, line 45: Good idea of using a factory here!

app/config.py
```diff
@@ -35,7 +70,8 @@ def get_abs_path(file_path: str):
 PROMO_CODE = "SIMPLEISBETTER"
 
 # Debug mode
-DEBUG = os.environ["DEBUG"] if "DEBUG" in os.environ else False
+DEBUG = bool(os.environ.get("DEBUG", default=False))
```
nguyenkims, line 73: 👌

nguyenkims (commented):
Hey the PR looks almost perfect now 💪, just left some final comments :)

app/config.py
doanguyen: I think the problem is that `literal_eval` will raised for incorrect expressions, the `try/catch` block will remove this edge case

app/config.py
doanguyen: in fact ,`email1, email2` is not a expression in Python :(

nguyenkims: Ah this option is usually put as 

```
IGNORED_EMAILS = ["string1", "string2"]
```

and that's why `eval` was used :)

app/config.py
doanguyen: but `EMAIL_DOMAIN` is supposed to be a list or just a string?

app/config.py
doanguyen: done

app/config.py
nguyenkims: Oh you're right, EMAIL_DOMAIN is just a string indeed.

nguyenkims:
I close this PR as the PR code is copied in this commit :)

https://github.com/simple-login/app/commit/3af98026e36e634666f11a8bdadf0792559a68db

PR #381, opened 2021-03-10
Feature/import export tests

TheLastProject:
The export test works, but the import test fails because it wants to do a HTTP request to localhost:7777 while the server isn't running.

Because this could become a bit more complex in refactoring, I'd like to create this PR in WIP status to hear how you'd prefer the import / temporary file thing to work.


tests/api/test_import_export.py
```diff
@@ -0,0 +1,136 @@
...
+
+from app import alias_utils, s3
+from app.extensions import db
+from app.models import User, CustomDomain, Mailbox, Alias, AliasMailbox, ApiKey, File, BatchImport
+from app.import_utils import handle_batch_import
+from app.utils import random_string
+
+def test_export(flask_client):
+    # Create users
+    user1 = User.create(
```
nguyenkims, line 14: There's this method that will create a user and logs this user in so we don't have to create an api key in tests:

https://github.com/simple-login/app/blob/38790fdc84c26b9a09615ee7808dac982ab39361/tests/utils.py#L8

For example it's used here 
https://github.com/simple-login/app/blob/2b4dc3cdccea4f27898bb86e8f541efef15b2978/tests/dashboard/test_custom_alias.py#L33

tests/api/test_import_export.py
```diff
@@ -0,0 +1,136 @@
...
+        url_for("api.export_aliases"), headers={"Authentication": api_key.code}
+    )
+    assert r.status_code == 200
+    assert r.mimetype == "text/csv"
+    assert r.data == """alias,note,enabled,mailboxes
+ebay@my-domain.com,Used on eBay,True,destination@my-destination-domain.com
+facebook@my-domain.com,"Used on Facebook, Instagram.",True,destination@my-destination-domain.com destination2@my-destination-domain.com
+""".replace("\n", "\r\n").encode()
+
+def test_import_no_mailboxes(flask_client):
```
nguyenkims, line 110: As `handle_batch_import` downloads the csv file, this test will indeed fail. I wonder if we can extract file downloading from `handle_batch_import` and test the file parsing instead, something like

```
def handle_batch_import(batch_import: BatchImport):
    user = batch_import.user

    batch_import.processed = True
    db.session.commit()

    LOG.debug("Start batch import for %s %s", batch_import, user)
    file_url = s3.get_url(batch_import.file.path)

    LOG.d("Download file %s from %s", batch_import.file, file_url)
    r = requests.get(file_url)
    lines = [line.decode() for line in r.iter_lines()]
    # this is new
    import_from_csv(lines)

def import_from_csv(lines):
    # .... the rest from handle_batch_import()
```

nguyenkims:
@TheLastProject hey the PR looks good! I just left some comments/suggestions :).

TheLastProject:
I consider this ready for review now. You probably want to squash these commits if you merge it. Please tell me if there's still anything you want me to test.

nguyenkims (approved):
LGTM! Thanks very much for making this PR 👍!

PR #421, opened 2021-03-29
Missing DKIM_PUBLIC_KEY_PATH in readme and example.env

froozeify:
When I was trying to install and test simple-login, I had an error with my `~/simplelogin.env`.

Reading back the README and the `example.env` I discovered that one parameter was missing in both making all the script failed (like `python init_app.py`)

I hope my little PR will help some other people that was like me and didn't know why it won't init 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,8 @@ EMAIL_SERVERS_WITH_PRIORITY=[(10, "app.mydomain.com.")]
 # this option doesn't make sense in self-hosted. Set this variable to disable this option.
 DISABLE_ALIAS_SUFFIX=1
 
-# the DKIM private key used to compute DKIM-Signature
+# the DKIM public/private keys used to compute DKIM-Signature
+DKIM_PUBLIC_KEY_PATH=/dkim.pub.key
 DKIM_PRIVATE_KEY_PATH=/dkim.key
 
 # DB Connection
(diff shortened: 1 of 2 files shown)


froozeify:
@nguyenkims Small ping since I didn't create an issue, I don't know if you saw my PR :)

nguyenkims:
@froozeify hey thanks a lot for fixing it! This `DKIM_PUBLIC_KEY_PATH` has been removed *after* *the 3.3.0 is released and causes the issue indeed! I'll merge the PR and will remove the `DKIM_PUBLIC_KEY_PATH` again in the next version :).

froozeify:
ohh, ok, so my mistake. Not a bug.

Since the repo was on master I though it was like the release version, my bad. 😢 

An idea could be to rename the master branch as many library does to avoid that confusion (also for better evolution and scalability dut to the LTS support they often propose). 

For example (symfony)[https://github.com/symfony/symfony] the default branch is called **5.x** so when we saw that we can think to go to the `Switch branches/tags` toggle.

Next time I'll check the tags before creating a PR 😉

nguyenkims:
@froozeify  Oh no worries, this is actually our mistake 😅, we should haven't removed the `DKIM_PUBLIC_KEY_PATH` when the latest published version 3.3.0 still requires it.

Btw I like the symfony idea of using the 5.x version, we'll do that in the next version then!

PR #434, opened 2021-04-14
Delete shell.py

Naw353:


nguyenkims:
@Naw353 hey what is the reason to delete `shell.py`? It' quite practical when we need to run some ad-hoc operations :)

Issue #446, opened 2021-04-27
More documentation for custom docker networks


AzuritBadger:
Hi,

I recently got SimpleLogin to run on my own server. During the process, I discovered some things one could improve in the self hosting instructions. 

As mentioned in #44, SpamAssasin doesn't like the `240.0.0.0/24` network. Therefore a different subset should be used, like `10.0.0.1/24` or something similar. 

When I chose said subnet, SimpleLogin wasn't able to send or recive emails. After some trial and error I figured out that `POSTFIX_SERVER` inside the env file needs to point to the gateway of the docker network for SimpleLogin to function correctly.

Therefore, my suggestion is to add some more documentation about custom docker networks. For example, in the _Run SimpleLogin Docker containers_ section, one could add: If your docker network has a different subnet than the one used above, you need to set `POSTFIX_SERVER` to the gateway of your docker network. Otherwise SimpleLogin will not be able to send or recive emails.

Or SimpleLogin could use the gateway of the network as the default postfix server IP. 

So far I really enjoy SimpleLogin and hope this project can continue to grow. Thank you for this awsome service 💯

nguyenkims:
@AzuritBadger that's a good idea to have a section for a custom Docker network. Do you want to create a PR to include this doc :)?

We wanted to change the 240.0.0.0 network to 10.0.0.0 but as this network is already used in some installations, this change would break these installations ...

Issue #44, opened 2020-01-23
SpamAssasin returns RCVD_ILLEGAL_IP when sending e-mails through simplelogin.io


jarylc:
Likely because of this line in the source

`Received: from [240.0.0.3] (unknown [240.0.0.3]) by mx1.simplelogin.co (Postfix) with ESMTP id 24C385C1F7`

nguyenkims:
Hi can you give more info about your setup please? 

`240.0.0.3` is from the SimpleLogin docker container network (which is `240.0.0.0/24`).

jarylc:
I've done 2 test e-mails on mail-tester.com today (it did not trigger the flag when I tested yesterday).

I'm not entirely sure what causes Spam Assassin to flag for `RCVD_ILLEGAL_IP`, but looking at the description and the source so far, the only IP that I could think it might be flagging is `[240.0.0.3] (unknown [240.0.0.3])` (see the source below).

Unless it's Google's ipv6 address `IPv6:2a00:1450:4864:20::131` which I think it's unlikely. Or this one `2002:ac2:5e9b::` (see the source below).

[Here's the Spam Assassin response.](https://imgur.com/a/k2hcLyi)

Here's the latest source for one of them with some content snipped out:
```
Received: by mail-tester.com (Postfix, from userid 500)
	id 5F473A4979; Thu, 23 Jan 2020 12:10:40 +0100 (CET)
Authentication-Results: mail-tester.com;
	dkim=pass (1024-bit key; unprotected) header.d=simplelogin.co header.i=@simplelogin.co header.b=F1GVCCb0;
	dkim-atps=neutral
X-Spam-Checker-Version: SpamAssassin 3.4.2 (2018-09-13) on mail-tester.com
X-Spam-Level: *
X-Spam-Status: No/1.1/5.0
X-Spam-Test-Scores: DKIM_SIGNED=0.1,DKIM_VALID=-0.1,DKIM_VALID_AU=-0.1,
	DKIM_VALID_EF=-0.1,HTML_MESSAGE=0.001,RCVD_ILLEGAL_IP=1.3,
	SPF_PASS=-0.001,URIBL_BLOCKED=0.001
X-Spam-Last-External-IP: 94.237.86.150
X-Spam-Last-External-HELO: mx1.simplelogin.co
X-Spam-Last-External-rDNS: mx1.simplelogin.co
X-Spam-Date-of-Scan: Thu, 23 Jan 2020 12:10:40 +0100
X-Spam-Report: 
[...]

PR #449, opened 2021-04-29
Added documentation for custom docker networks

AzuritBadger:
Added said documentation I mentioned in #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ Choose "Internet Site" in Postfix installation window then keep using the propos
 ![](./docs/postfix-installation.png)
 ![](./docs/postfix-installation2.png)
 
-Replace `/etc/postfix/main.cf` with the following content. Make sure to replace `mydomain.com` by your domain. If your docker network uses a different subnet, make sure to change `240.0.0.0/24` to your subnet in the `mynetworks` variable.
+Replace `/etc/postfix/main.cf` with the following content. Make sure to replace `mydomain.com` by your domain. If your Docker network uses a different subnet, make sure to change `240.0.0.0/24` to your subnet in the `mynetworks` variable.
 
 ```
 # POSTFIX config file, adapted for SimpleLogin
@@ -396,7 +396,7 @@ To run SimpleLogin, you need a config file at `~/simplelogin.env`. Below is an e
 All possible parameters can be found in [config example](example.env). Some are optional and are commented out by default. 
 Some have "dummy" values, fill them up if you want to enable these features (Paddle, AWS, etc).
 
-If your Docker network has a different subnet than the one used above, you need to set `POSTFIX_SERVER` to the gateway of your Docker network. 
+If you used a different subnet for your Docker network, you need to set `POSTFIX_SERVER` to the gateway IP address of your Docker network. 
 Otherwise SimpleLogin will not be able to send or recive emails.
 
 ```.env
--- a/README.md
+++ b/README.md
@@ -226,6 +226,17 @@ sudo docker network create -d bridge \
     --gateway=240.0.0.1 \
     sl-network
 ```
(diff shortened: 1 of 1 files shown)


nguyenkims:
Hey thanks for making this PR! I've thought about this and I think we can simplify this part by:
 
- replacing `240.0.0.0` by `10.0.0.0` in the README.md file. 
- setting the POSTFIX_SERVER to `10.0.0.1` in the Postfix section as if not set, SimpleLogin uses 240.0.0.1 by default.

This shouldn't affect existing user as they will use https://github.com/simple-login/app/blob/95c8f14ea50240c1b59d008ea0dc911dd94da3cd/docs/upgrade.md#L1-L0 for upgrading existing SimpleLogin instance.

What do you think :)?

nguyenkims:
@AzuritBadger sounds good! Do you want to make this change in this PR :)?

fareszr:
@nguyenkims before you merge this
@AzuritBadger you need to edit this, change it to 10.0.0.1
https://github.com/AzuritBadger/app/blob/master/app/config.py#L101

make it `POSTFIX_SERVER = os.environ.get("POSTFIX_SERVER", "10.0.0.1")`

if you don't, the default option will not work.

nguyenkims:
@fareszr the `POSTFIX_SERVER=10.0.0.1` in this PR is set in the `simplelogin.env` would set the correct value for `POSTFIX_SERVER` I think.

fareszr:
> @fareszr the `POSTFIX_SERVER=10.0.0.1` in this PR is set in the `simplelogin.env` would set the correct value for `POSTFIX_SERVER` I think.

still i think if you are going to change the default subnet to 10.0.0.0/24, the default option should change too

nguyenkims:
@fareszr hey updating the default value to 10.0.0.1 would break existing installations as they are using 240.0.0.0 as network and they don't have `POSTFIX_SERVER` set in their config file.

AzuritBadger:
@fareszr Changing the IP in `config.py` isn't neccessary for it to work. While I was figuring out how to use custom networks, I first changed the `POSTFIX_SERVER` variable in `config.py` too, only to find out later that setting `POSTFIX_SERVER` in `simplelogin.env` does the same thing. Besides, as @nguyenkims mentioned, changing the default option would break existing installations. 

I think the best solution would be to always use the gateway of the given Docker network, though I do not know if this is possible to implement.

nguyenkims:
@AzuritBadger the docker network gateway is **usually** 172.17.0.0 though this isn't guaranteed to be the same on all computers. It's then safer to create our own Docker network.

AzuritBadger:
Sorry, I could have phrased it a little better 😅, I actually meant the network that we create for the containers, i.e. `sl-network`. But as I said, thats just an idea, the way it is now works totaly fine as long as the config file is setup as needed.

0xspade:
The new network for sl-network is not yet implemented in README.md in the main repo.

nguyenkims:
> The new network for sl-network is not yet implemented in README.md in the main repo.

The sl-network is documented in the README right?

image

PureFallen:
> > The new network for sl-network is not yet implemented in README.md in the main repo.
> 
> The sl-network is documented in the README right?
> 
> image

@0xspade probably refers to the changed `sl-network`, the one that is part of this pull request.
I recently started self-hosting SimpleLogin as well and stumbled into the same issue: `RCVD_ILLEGAL_IP`.

While the official instance changed it's network in #44 , the README not mentions the issue of `240.0.0.0` triggering SpamAssassin. Glad i was able to find this pull request by cheer luck.

nguyenkims:
In the latest commit, I've updated to use the 10.0.0.0 network instead of 240.0.0.0 one. I'll close this PR then.

PR #472, opened 2021-05-13
Add HIBP checks

TheLastProject:
This adds HIBP support and fixes #400.

When the cron is started, it will:
1. Get a list of all known breaches and populate the database (for now only with name and description, but more data exists, see https://haveibeenpwned.com/api/v3/breaches).
2. Get a list of all aliases
3. For each alias, check if it was not checked on the API in the last HIBP_SCAN_INTERVAL_DAYS (default: 7)
4. If it has been long enough, see what breaches the alias has been involved in.

There is still some TODO, but I wanted to run this by y'all first. I am also wondering how cron is exactly used on real environments. Should check_hibp perhaps only check the oldest 100 aliases or something each time the function is called? Given at most 1 API request can be done per API key per 1.5 seconds?

image

TODO:
- [x] Black formatting
- [x] Prevent multiple checks running in cron (how do you currently deal with cron and prevent this for other jobs?)
- [ ] Unit tests (not completely sure how to do this, as the data can keep changing. I have suggested https://haveibeenpwned.uservoice.com/forums/275398-general/suggestions/43425102-test-accounts-that-always-return-the-same-results to make testing this easier)

Maybe TODO now, maybe later:
- [ ] New page for HIBP details (maybe in another PR?)
- [ ] Email the alias when it was found in a new breach?


app/dashboard/templates/dashboard/index.html
```diff
@@ -202,6 +202,13 @@
                 <span class="fa fa-heart" data-toggle="tooltip"
                       title="This alias added to favorite"></span>
               {% endif %}
+
+              {% if alias.hibp_breaches | length > 0 %}
+                <a href="https://haveibeenpwned.com/">
+                  <span class="fa fa-warning text-danger" data-toggle="tooltip"
+                        title="This alias was found in {{ alias.hibp_breaches | length }} data breaches. Check haveibeenpwned.com for more information."></span>
```
nguyenkims, line 209: I wonder if there's a way to create a link to HIBP that shows the data breach for a given email address? It'd save user a step to enter their alias on HIBP :)

app/models.py
```diff
@@ -162,6 +163,14 @@ class AliasGeneratorEnum(EnumE):
     uuid = 2  # aliases are generated based on uuid
 
 
+class Hibp(db.Model, ModelMixin):
+    __tablename__ = "hibp"
+    name = db.Column(db.String(), nullable=False, unique=True, index=True)
+    breached_aliases = db.relationship("Alias", secondary="alias_hibp")
+
+    description = db.Column(db.String())
```
nguyenkims, line 171: For a long field, I think we can use `db.Text`

app/models.py
```diff
@@ -2004,6 +2020,17 @@ class AliasMailbox(db.Model, ModelMixin):
     alias = db.relationship(Alias)
 
 
+class AliasHibp(db.Model):
```
nguyenkims, line 2023: I think we can make this class inherit from `ModelMixin` too so it can have some fields and utility methods. Almost all model classes inherit from ModelMixin I think :).

app/models.py
```diff
@@ -2004,6 +2020,17 @@ class AliasMailbox(db.Model, ModelMixin):
...
+    __tablename__ = "alias_hibp"
+
+    id = db.Column(db.Integer, primary_key=True)
+    alias_id = db.Column(db.Integer(), db.ForeignKey("alias.id"))
+    hibp_name = db.Column(db.String(), db.ForeignKey("hibp.name"))
+
+    alias = db.relationship("Alias", backref=db.backref("alias_hibp", cascade="all, delete-orphan"))
```
nguyenkims, line 2030: Didn't know about the "delete-orphan", thanks !

cron.py
```diff
@@ -757,6 +761,72 @@ def delete_old_monitoring():
...
+
+    LOG.d("Updating list of known breaches")
+    r = requests.get("https://haveibeenpwned.com/api/v3/breaches")
+    for entry in r.json():
+        hibp_entry = Hibp.get_by(name=entry['Name'])
+        if hibp_entry:
+            hibp_entry.description = entry['Description']
+        else:
+            hibp_entry = db.session.add(
+                Hibp(
```
nguyenkims, line 784: Nits: we can use `Hibp.create(...)` here as this method is provided by ModelMixin: https://github.com/simple-login/app/blob/30183ac8c3f705cac7d834dfe811b96b56ff81bc/app/models.py#L75

cron.py
```diff
@@ -757,6 +761,72 @@ def delete_old_monitoring():
...
+                    name=entry['Name'],
+                    description=entry['Description']
+                )
+            )
+
+    db.session.commit()
+
+    for alias in Alias.query.all():
+        LOG.d("Checking %s on HIBP", alias.email)
```
nguyenkims, line 793: Nits: we can just use `alias` instead of `alias.email` as this usually gives more info and shorter to type :)

server.py
```diff
@@ -298,6 +298,11 @@ def fake_data():
     m1.pgp_finger_print = load_public_key(pgp_public_key)
     db.session.commit()
 
+    # example@example.com is in a LOT of data breaches
```
nguyenkims, line 301: Didn't know that 😃

nguyenkims (commented):
The PR looks really good! Just left some comments and nits :)

nguyenkims:
> I am also wondering how cron is exactly used on real environments. Should check_hibp perhaps only check the oldest 100 aliases or something each time the function is called? Given at most 1 API request can be done per API key per 1.5 seconds?

The cron is run via `yacron` https://github.com/gjcarneiro/yacron with this config file https://github.com/simple-login/app/blob/b3181c054fdcc7c9b1aacfd60a94bf49e4cf0ae4/crontab.yml#L1-L0

> Should check_hibp perhaps only check the oldest 100 aliases or something each time the function is called? Given at most 1 API request can be done per API key per 1.5 seconds?

Currently there are about 500K aliases in SimpleLogin database. If an alias takes 1.5s in average, it'd take 9 days to run checks for all of them. Given new version is deployed almost every day, I think the cronjob will have a good chance of being aborted before finish. I think we might need another strategy or ask HIBP to remove the 1.5s limit, what do you think?

app/dashboard/templates/dashboard/index.html
TheLastProject: I've pinged Troy Hunt about this on Twitter. Let's hope it is possible, would be much better yeah :)

cron.py
TheLastProject: I ended up using `get_or_create()`, as I only want to create it if it doesn't exist yet.

TheLastProject:
I have changed the cron to prioritize aliases never checked or checked very long again and set up `crontab.yml` to not allow concurrent jobs for this.

If you want to scan IDs faster, well, just buy more HIBP API keys. They're only 3.50 USD a month per key and I made it possible to add multiple keys to the `HIBP_API_KEYS` config setting. Maybe Troy Hunt is willing to cut you a deal given this is an Open Source project, maybe not. That's his and your business, not mine :)

app/dashboard/templates/dashboard/index.html
nguyenkims: Yeah, it'd be great!

TheLastProject:
In 42cfce7ce1da4c5de83379ad0d7795df9bc6f990 I have further optimised the API fetching code.

The logic is now as follows:
1. Get the list of all known breaches
2. Build a queue of all aliases that are due for a HIBP check, sorted by least-recently-checked first
3. For each api key in `HIBP_API_KEYS`, start a [Process](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process) which takes one entry from the queue, does an API request to HIBP, saves info to the database and then waits 1.5 seconds

This way the HIBP API check can scale very efficiently with multiple API keys until a CPU/network bottleneck is reached. This is probably the best I can make it.

TheLastProject:
I am thinking in the `_hibp_check` function we could send an email notification to the alias that was found in a breach if `alias.hibp_breaches` has changed.

I worry though, given this will be the first HIBP scan, that it may cause a LOT of emails outgoing and it may cause SimpleLogin to end up on some blacklist perhaps? Or do you think it would be safe to do so anyway?

nguyenkims:
@TheLastProject yes I think that's a good idea to let people know if their aliases appear in a breach! To reduce the number of emails, maybe we can aggregate the result so a user would receive at most an email (that contains a list of their aliases that appear in a breach)?

TheLastProject:
Aggregating would probably require a second cron job and storing both last breaches info and current breaches info. It would be less clean I feel. Possible, but I would prefer to avoid it if possible.

Is it possible for you to run the current code over a copy of the aliases list and do a DB query to see how many emails would be send? For each user, seeing how many aliases have a non-empty hibp_breaches list would give us an idea how many emails the average and hardest-hit user will get. Maybe the hardest-hit user would only get like 5 emails, then I think we shouldn't worry. But if they would get 200 emails, then aggregating is definitely required.

nguyenkims:
@TheLastProject that indeed makes sense to know how many aliases are actually leaked. Let me merge this PR and run it over the current database then.

cron.py
```diff
@@ -757,6 +763,106 @@ def delete_old_monitoring():
...
+    r = requests.get("https://haveibeenpwned.com/api/v3/breaches")
+    for entry in r.json():
+        Hibp.get_or_create(id=entry["Name"])
+
+    db.session.commit()
+    LOG.d("Updated list of known breaches")
+
+    LOG.d("Preparing list of aliases to check")
+    queue = multiprocessing.Queue()
+    for alias in Alias.query.order_by(Alias.hibp_last_check.asc().nullsfirst()).all():
```
nguyenkims, line 833: I wonder if this query can be sped up by using something like

```
max_date = arrow.now().shift(days=-7)
for alias in Alias.query.filter(or_(Alias.hibp_last_check.is_(None), Alias.hibp_last_check < max_date))).all():
    queue.put(alias.id)
```

so we don't have to evaluate `alias.needs_hibp_scan()` for every alias which can take some times for +500K aliases :).

cron.py
```diff
@@ -757,6 +763,106 @@ def delete_old_monitoring():
...
+    LOG.d("Updating list of known breaches")
+    r = requests.get("https://haveibeenpwned.com/api/v3/breaches")
+    for entry in r.json():
+        Hibp.get_or_create(id=entry["Name"])
+
+    db.session.commit()
+    LOG.d("Updated list of known breaches")
+
+    LOG.d("Preparing list of aliases to check")
+    queue = multiprocessing.Queue()
```
nguyenkims, line 832: I wonder if we can split the alias to handle into `len(HIBP_API_KEYS)` chunks and launch a process for handling each chunk? I feel this way is a bit more simple to understand but this is probably because I'm not familiar with using Queue ...

crontab.yml
```diff
@@ -52,3 +52,10 @@ jobs:
     shell: /bin/bash
     schedule: "0 15 * * *"
     captureStderr: true
+
+  - name: SimpleLogin HIBP check
+    command: python /code/cron.py -j check_hibp
+    shell: /bin/bash
+    schedule: "0 18 * * *"
+    captureStderr: true
+    concurrencyPolicy: "forbid"
```
nguyenkims, line 61: didn't know this `concurrencyPolicy` option, I think I'll add this for all cron jobs :)

app/models.py
```diff
@@ -162,6 +163,12 @@ class AliasGeneratorEnum(EnumE):
     uuid = 2  # aliases are generated based on uuid
 
 
+class Hibp(db.Model, ModelMixin):
+    __tablename__ = "hibp"
+    id = db.Column(db.String(), primary_key=True)
```
nguyenkims, line 168: Ah this `id` field will conflict with the ModelMixin.id and different than the convention in the code to use integer for `id` column. I think we can use the `name` with a unique constraint here as in the previous commit. It's also easier to match with the response from HIBP.

nguyenkims (commented):
Just left some comments :)

cron.py
TheLastProject: Will also need the do the order but yeah, databases tend to be way faster at this. I'll change that.

cron.py
TheLastProject: It is possible.

The benefit of my current method is though that if for some reason some web request ends up stalling (think: network hiccup and it ends up getting a timeout), it won't cause any slowdown as each of the API keys will grab an entry from the queue as soon as it is ready to do another request. If you split it in 4 beforehand and one web request would time out, it would be slower at the end. Not really a big difference, but splitting it up before starting makes the code more complex and introduces an odd edge cases with network timeouts.

app/models.py
TheLastProject: Well, it overrides the id from ModelMixin, I see no issues with the way I built it. The problem with making name a new constraint is that each breach has an unique name so it feels illogical for that to not be the primary key. Unless I'm missing something meaningful?

I can change it if you really want though, but I'm not truly convinced myself it would be cleaner.

cron.py
TheLastProject: Fixed in 73c2bece704567872cf3067bc8b4df7435535ef0.

app/models.py
nguyenkims: Hey a table can have a primary key and several unique constraints so database wise, it doesn't feel unnatural for me to have an `id` as primary key with a couple of unique constraints. I think the main benefit of having a default integer `id` in all tables is consistency as we know this column is going to be there in all tables :)

I don't like this field is called "name" though as it doesn't convey "uniqueness" but this is a HIBP choice ...

cron.py
nguyenkims: Oh I see, that makes sense 👌.

I'm wondering if asyncio can be a good fit for our usecase here? Your previous implementation is actually almost ready to  convert into a asyncio code :).

TheLastProject: Oh, perhaps. I just have never used asyncio before, so I really wouldn't know where to get started asyncio-ifying this

app/models.py
TheLastProject: Fixed in a08b0c05cc3d41c664dfab7aab10cb5addc948b7.

cron.py
TheLastProject: Well, turns out it is actually super simple. Done in 40d0dee.

nguyenkims: Wow I love the new implementation!

nguyenkims (approved):
LGTM! Thanks for making this PR!

TheLastProject:
Keep me updated on the info of the amount of aliases leaked and if we will need to aggregate warnings or not :)

TheLastProject:
Well, my original logic was to only update hibp_last_check on a succesful check, so it would just be retried next loop and that alias would be prioritized because I had `.order_by(Alias.hibp_last_check.asc().nullsfirst())`.

The reason for this was that, if for some reason some alias was causing an error at HIBP, that it wouldn't break everything.

I see now that somewhere along the line (seemingly https://github.com/simple-login/app/pull/472/commits/42cfce7ce1da4c5de83379ad0d7795df9bc6f990#diff-361f6063c4b3f3893794449c56b419ac01f700c74a7311014000cd05177a2903L820) I accidentally removed the `         alias.hibp_last_check = arrow.utcnow()` line... oops...

nguyenkims:
@TheLastProject hey the HIBP run seems to be complete now :). Most users have 1-2 aliases in breach but there's 1 user who has more than 200 leaked aliases so I guess she would appreciate an email aggregation :)

TheLastProject:
Oh dear, that are... quite a lot of leaked aliases.

I need to think of how to properly aggregate this information. Maybe a new field on each alias called something like `hibp_breaches_not_notified_user` or something (that each time hibp_breaches gets updated with a new entry gets that new entry added) and then a cronjob that, for each account, loops over all aliases, collects (and empties) the `hibp_breaches_not_notified_user` list and mails that out? That would seem the most reasonable way to aggregate this all I suppose.

nguyenkims:
What do you think about adding a filter option that only displays leaked aliases and send out this URL in the warning email in case there is a lot (>10) of leaked aliases? When the number of leaked alias is small, we can just add all these aliases in the email.

image

PR #473, opened 2021-05-13
Able to config random suffix

developStorm:
Implement #426


app/config.py
```diff
@@ -374,3 +374,5 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 ALIAS_LIMIT = os.environ.get("ALIAS_LIMIT") or "100/day;50/hour;5/minute"
 
 ENABLE_SPAM_ASSASSIN = "ENABLE_SPAM_ASSASSIN" in os.environ
+
+ALIAS_RAND_SUFFIX_LENGTH = int(os.environ.get("ALIAS_RAND_SUFFIX_LENGTH", 5))
```
nguyenkims, line 378: I wonder if we can use "RANDOM" instead of the abbreviated version "RAND" :)?

app/dashboard/templates/dashboard/setting.html
```diff
@@ -247,6 +247,22 @@
           </select>
           <button class="btn btn-outline-primary">Update</button>
         </form>
+
+        <div class="mt-3 mb-1">Select the default suffix generator for aliases.</div>
```
nguyenkims, line 251: I think we can add `id=random-alias-suffix` here so when user clicks on "Update", they will come back to this section.

app/config.py
nguyenkims: Do you think 5 characters can be enough to avoid potential conflict?

app/dashboard/views/custom_alias.py
```diff
@@ -250,6 +252,20 @@ def custom_alias():
     )
 
 
+def get_suffix(user: User) -> str:
```
nguyenkims, line 255: Hey I wonder if we should replace `random_word()` by this method in other places, for example https://github.com/simple-login/app/blob/30183ac8c3f705cac7d834dfe811b96b56ff81bc/app/models.py#L1127 ?

app/dashboard/views/custom_alias.py
```diff
@@ -281,14 +297,6 @@ def verify_prefix_suffix(user: User, alias_prefix, alias_suffix) -> bool:
             LOG.exception("User %s submits a wrong alias suffix %s", user, alias_suffix)
             return False
 
-        random_word_part = alias_domain_prefix[1:]
```
nguyenkims, line 284: Let's wait until the Safari extension (cf my comment on the Discussion thread) getting updated and then we can merge this PR :)

app/models.py
```diff
@@ -162,6 +162,11 @@ class AliasGeneratorEnum(EnumE):
     uuid = 2  # aliases are generated based on uuid
 
 
+class AliasSuffixEnum(EnumE):
+    word = 0  # Random word from dictionary file
+    rnd_string = 1  # Completely random string
```
nguyenkims, line 167: Can we replace `rnd` by `random`? :)

app/models.py
```diff
@@ -287,6 +292,16 @@ class User(db.Model, ModelMixin, UserMixin):
         db.Boolean, default=False, nullable=False, server_default="0"
     )
 
+    # whether to use random string or random word as suffix
```
nguyenkims, line 295: I think we need to create a migration file here, it can be created by running `new_migration.sh`

app/dashboard/views/custom_alias.py
nguyenkims, line 255: Hey it'd be best to make sure that the suffix isn't already used in other aliases as otherwise it'd increase the conflict chance :), for example when 2 users want to name their aliases `facebook.{random suffix}@alias.com`

nguyenkims (commented):
Hey the PR looks good, thanks for creating it! Just left some comments :)

app/config.py
developStorm: That's (26+10)^5=60,466,176 possible strings so I think it would be enough 🤔

app/dashboard/views/custom_alias.py
developStorm: I'm not sure if that will have any side effect because it looks like the implementation is a little different elsewhere (for example, there doesn't seem to be a signature of suffix?)

developStorm: I wonder why this needs to be checked in advance? If a conflict does occur, it should be caught by the database when the alias is eventually created. To ensure that no existing suffix match, we'll need to go through all existing aliases in the database. The probability of a conflict is not very high in the first place, so it seems not very efficient/worth to check for conflict in advance every time.

app/models.py
developStorm: I am unable to create that since I don't have the full installation of SL on my host computer, and it seems flask will not run in this case unless I manually find and install all missing packages

app/dashboard/views/custom_alias.py
nguyenkims: The `random_word` function is mostly used for generating an alias suffix so I guess we can almost replace it everywhere by `get_suffix()`

app/models.py
nguyenkims: Oh I see, I can run the script later when the PR is merged then.

app/dashboard/views/custom_alias.py
nguyenkims: I think it's better that we can reduce the chance of conflict :). A check like `Alias.query.filter(Alias.email.endswith(suffix)).count()` might be enough to know whether a suffix has been used right? I think we can do something similar to this to look for a suffix that hasn't been used before https://github.com/simple-login/app/blob/ef2eb7f95958c8f9df8c38f594b3006331a3d915/app/models.py#L1111

app/dashboard/views/custom_alias.py
nguyenkims: @developStorm For the heads-up, the new Safari version has been released on Monday so I think we can confidently merge that change on next Monday :).

app/dashboard/views/custom_alias.py
developStorm: My point is that avoiding conflict at this stage is not necessary. It may increase unnecessary database queries and prolong alias creation time. I think each query against the suffix needs to search all the aliases in the entire alias table, and it's not an exact match - so it may not be possible to use indexes to speed it up.

Ultimately when adding an alias, the database should ensure that there are no two conflicting aliases, so in the most extreme case it will just throw an error and allow the user to reselect the suffix. Even this extreme case should be difficult to encounter, as the prefix-suffix namespace is large enough.

If you're still skeptical ...... maybe you can put a debug log to see what the probability of entering this 1000-times loop is with the current generation mechanism. Using a random suffix should result in a lower probability of conflict than the current word generator.

app/dashboard/views/custom_alias.py
developStorm: Are we ready to merge now? 👀

app/dashboard/views/custom_alias.py
nguyenkims: Oh I see, let's keep this current approach and if too many users complain, we can add the check later :). The code makes sure that alias address is unique but displays an error if an alias already exists so this case is covered :).

app/dashboard/views/custom_alias.py
nguyenkims: Not yet 🙁 , I still see the old endpoint being called by Safari extension so I guess the new version hasn't reached all devices yet. Let's wait for another week to be totally sure.

developStorm:
@nguyenkims Seems I merged sth wrong from 3e1f098 
https://github.com/simple-login/app/pull/473/checks?check_run_id=2755982433#step:9:11

developStorm:
Any updates on this? @nguyenkims

nguyenkims (approved):
LGTM! Thanks for making this PR and all the patience!

PR #483, opened 2021-05-22
chore: remove sudo in running docker

doanguyen:
Running docker in `sudo` mode is considered harmful.
It's recommended to run docker as non-root user to
minimize the security risks.

[0]: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ If you already have a Postgres database in use, you can skip this section and ju
 Run a Postgres Docker container as your Postgres database server. Make sure to replace `myuser` and `mypassword` with something more secret.
 
 ```bash
-sudo docker run -d \
+docker run -d \
     --name sl-db \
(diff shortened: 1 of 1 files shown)


nguyenkims:
Ah on some systems, Docker can only be run via `sudo` so I think it's better to leave sudo there.

doanguyen:
> Ah on some systems, Docker can only be run via `sudo` so I think it's better to leave sudo there.

I think it's recommended to use non-root users that run docker-daemon [0], it remains quite a lot of privileges for the user group that runs docker but still reduces the security risks. 

[0] https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

nguyenkims (approved):

PR #486, opened 2021-05-24
HIBP aggregated breaches email alert

TheLastProject:
This PR does the following:

UI:
- Add "Only Aliases Found In Breaches" filter

HIBP check cron:
- Split `hibp_breaches` into `hibp_breaches_notified_user` and `hibp_breaches_not_notified_user`
- On a HIBP check, put each new breach in `hibp_breaches_not_notified_user`

HIBP notify cron:
- Notify for all `hibp_breaches_not_notified_user` entries
- Move all `hibp_breaches_not_notified_user` into `hibp_breaches_notified_user`

I haven't tested the email layout stuff thoroughly with multiple entries, so please test too.


nguyenkims:
@TheLastProject hey sorry for the late reply! I wonder if we can use the current model (i.e. without adding new models) and :

- add a filter for display all aliases that have at least a data breach. We can modify `get_alias_infos_with_pagination_v3` to take into account this, for example:
https://github.com/simple-login/app/commit/c5ba48edb981d1fb9358fce38375793e032eb885

- if a user has aliases in data breach, send them an email that contains a link to the home page with the `hibp` filter
- to make sure a user doesn't receive too many emails, we can use `send_email_with_rate_control` https://github.com/simple-login/app/blob/2f963229770154b8a67c86b16846511613ae11b8/app/email_utils.py#L299. This method is quite flexible and we can make sure this warning email isn't sent too frequently, for example less than 2/month.

For the email template, it's best to use the `{% call text() %}  {% endcall %}` as it's the most flexible way. For the CTA button, we can use the `render_button()` method. 

This PR can be a bit laborious, let me know if you want me to take over it :).

TheLastProject:
The thing is though, I don't think anyone wants to be notified over and over about the same breaches. Just a single notification per new breach, and let the user decide what to do.

You also don't want to delay notifying the user of being in a breach for too long, as being in a breach is a security risk.

The new model is really only added to store the information of breach types, so we can possibly translate them at some point (tell French users in French that their location was leaked).

Of course you're free to take over the PR, I'm here to help, not to delay you :)

nguyenkims:
@TheLastProject oh I see, that makes sense that user shouldn't be notified about the same alias twice. In this case, maybe we can add a new field `hibp_notified` to an alias and set it to True when the warning email is sent? I feel creating 2 classes because of a notification purpose (which isn't core IMO) is a bit too complex :).

> Of course you're free to take over the PR, I'm here to help, not to delay you :)

Ah no, it's very kind of you to take care of these features! (the HIBP integration will arrive much later without you 😅). It's just sending email notification isn't core and requires a lot of testing (https://github.com/simple-login/app/blob/master/CONTRIBUTING.md#test-sending-email can be useful) and if you don't want to do that, just let me know :).

TheLastProject:
Well, the two extra classes are solely for being able to give more useful information in the email though (description and the types of data that got leaked like Passwords or Location or whatever). See https://haveibeenpwned.com/api/v3/dataclasses for that. We can save the 2 models if we are okay with not telling users what kind of data was leaked (or if we hackily append it to the description or so). See https://github.com/simple-login/app/pull/486/files#diff-7e935e0b4fef52cf7eb3c57068bb65cf68de9804671085a84cdcd90df74f7e94R21.

The only real difference about preventing duplicate notifications is the the `hibp_breaches_not_notified_user` field, which just like a `hibp_notified` field would only be 1 extra field.

app/models.py
```diff
@@ -1062,11 +1076,16 @@ class Alias(db.Model, ModelMixin):
 
     # have I been pwned
     hibp_last_check = db.Column(ArrowType, default=None)
-    hibp_breaches = db.relationship("Hibp", secondary="alias_hibp")
+    hibp_breaches_notified_user = db.relationship("Hibp", secondary="alias_hibp")
```
nguyenkims, line 1079: Oh I didn't know we can share the same secondary class for a many-to-many fields!

app/models.py
```diff
@@ -162,9 +162,23 @@ class AliasGeneratorEnum(EnumE):
     uuid = 2  # aliases are generated based on uuid
 
 
+class HibpDataClass(db.Model, ModelMixin):
+    __tablename__ = "hibpdataclass"
+    description = db.Column(db.Text, nullable=False, unique=True, index=True)
```
nguyenkims, line 167: I wonder if we can use "name" or "attribute" instead of "description" here 🤔. "description" usually means a long text.

app/models.py
```diff
@@ -2061,6 +2080,27 @@ class DomainMailbox(db.Model, ModelMixin):
...
+class HibpHibpDataClass(db.Model, ModelMixin):
+    __tablename__ = "hibp_hibpdataclass"
+
+    __table_args__ = (
+        db.UniqueConstraint(
+            "hibp_id", "hibp_dataclass_id", name="uq_hibp_hibpdataclass"
+        ),
+    )
+
+    hibp_id = db.Column(db.Integer(), db.ForeignKey("hibp.id"))
```
nguyenkims, line 2092: can we add `index=True` here to make sure the foreign key is indexed? This table is rather small so indexing wouldn't make much difference here though.

nguyenkims: ( you might need to re-run the `sh new_migration.sh` to generate the migration script though )

cron.py
```diff
@@ -872,6 +895,48 @@ async def check_hibp():
...
 
 
+def notify_hibp():
+    """
+    Send aggregated email reports for HIBP breaches
+    """
+    for user in User.query.all():
+        new_breaches = (
+            Alias.query.filter(Alias.user_id == user.id)
+            .filter(Alias.hibp_breaches_not_notified_user)
```
nguyenkims, line 905: Didn't know that `.filter(Alias.hibp_breaches_not_notified_user)` is enough for sqlalchemy! I usually wrote a subquery for doing this kind of query but this way is much easier :).

cron.py
```diff
@@ -872,6 +895,48 @@ async def check_hibp():
...
+            ),
+        )
+
+        for alias in new_breaches:
+            alias.hibp_breaches_notified_user = (
+                alias.hibp_breaches_notified_user
+                + alias.hibp_breaches_not_notified_user
+            )
+            alias.hibp_breaches_not_notified_user = []
+            db.session.add(alias)
```
nguyenkims, line 935: is `db.session.add(alias)` needed?

nguyenkims (commented):
I think I finally I understand why do we need to separate hibp_breaches_notified_user and hibp_breaches_not_notified_user :), sorry for all the going back and forth! Just left some comments. Have you been able to test this locally?

app/models.py
TheLastProject: Heh, funny. I avoided name because you explicitly didn't like me using name for the Hibp model, but you accepted that because "that's how HIBP calls it". This HIBP doesn't give any explicit name, but I can change it. Just tell me if you want name or attribute.

cron.py
TheLastProject: Hmm, I hadn't tested it without but it made sense to me to explicitly add it given we just changed it and we want it to be part of the commit.

app/models.py
nguyenkims: Oh sorry, I don't like "name" neither as it doesn't convey uniqueness ... Let's try with "attribute" then as it seems that HIBP uses this word.

cron.py
nguyenkims: Oh I see, I always thought of "db.session.add()" as adding an object to the database but not the active session. I think your understanding is more correct 👍

TheLastProject:
Yeah, I've tested this locally by:

1. Running `python3 cron.py --job notify_hibp` and seeing no mail would be send
2. Running `python3 cron.py --job check_hibp` to update the list
3. Running the notify cron again to see a mail would be send
4. Running the notify cron again to see no mail would be send
5. Running check_hibp again
6. Running the notify cron again to see no mail would be send

The formatting of the mail may not be to your taste though, I consider that a "branding" thing. You would know better than anyone else what kind of branding SimpleLogin wants to go for and what phrasing best fits there. So please rewrite the mail templates in a way that makes you happiest.

nguyenkims:
@TheLastProject thanks! I'll merge this PR and deploy on our staging environment and test the email render there :).

thecosmicfrog:
Great work @TheLastProject! Thanks for merging @nguyenkims.

nguyenkims:
@TheLastProject hey I run the code locally with this address example@example.com and it seems `alias.hibp_breaches_notified_user` returns the same breaches as `alias.hibp_breaches_not_notified_user`. I think these lines below might need something ( a condition?) to distinguish hibp_breaches_notified_user and hibp_breaches_not_notified_user?

```
hibp_breaches_notified_user = db.relationship("Hibp", secondary="alias_hibp")
hibp_breaches_not_notified_user = db.relationship("Hibp", secondary="alias_hibp")
```

nguyenkims:
@TheLastProject do you think you can allow changes to this pull request? I tried to push some updates to the email template but it doesn't seem to work ...

TheLastProject:
I'll look into the breaches stuff later. Allowing changes is enabled though:

Screenshot 2021-06-03 at 11 27 54

nguyenkims:
I guess that's an issue with the `gh` cli tool then, I checked out the PR by `gh pr checkout 486`, made some changes and do `git push` but the PR isn't updated. The commits are pushed here instead https://github.com/simple-login/app/commits/feature/hibp_email_alert 🤔

TheLastProject:
> @TheLastProject hey I run the code locally with this address [example@example.com](mailto:example@example.com) and it seems `alias.hibp_breaches_notified_user` returns the same breaches as `alias.hibp_breaches_not_notified_user`. I think these lines below might need something ( a condition?) to distinguish hibp_breaches_notified_user and hibp_breaches_not_notified_user?
> 
> ```
> hibp_breaches_notified_user = db.relationship("Hibp", secondary="alias_hibp")
> hibp_breaches_not_notified_user = db.relationship("Hibp", secondary="alias_hibp")
> ```

How did you check this? I'm having trouble with looking at the SQLite database (can't find the columns at all for some reason).

I would expect the following behaviour:
1. Both empty
2. check_hibp runs
3. All are in not_notified
4. notify_hibp runs
5. All are in notified
6. check_hibp runs
7. Nothing changes because all the exploits are already in notified

nguyenkims:
@TheLastProject here's how I run the code:

1) Run the server.py to create a local database
2) Run `cron.py -j check_hibp` to fetch data from HIBP
3) Run shell.py and check the alias hibp_breaches_notified_user by something like

```
alias = Alias.get_by(email="example@example.com")
print(alias. hibp_breaches_notified_user)
print(alias. hibp_breaches_not_notified_user)
```

TheLastProject:
Hmm, not getting `shell.py` working, sadly:

```
NameError                                 Traceback (most recent call last)
~/Projects/Forks/SimpleLogin/app-1/shell.py in <module>
----> 1 alias = Alias.get_by(email="example@example.com")

NameError: name 'Alias' is not defined
```

nguyenkims:
Sorry for the late response, can you try this instead? (adding `from app.models import Alias`)

```
from app.models import Alias
alias = Alias.get_by(email="example@example.com")
print(alias. hibp_breaches_notified_user)
print(alias. hibp_breaches_not_notified_user)
```

TheLastProject:
Okay, yeah, I can reproduce this issue with those instructions. No clue yet how to fix it, but I can at least reproduce it.

nguyenkims:
I have extracted and adapted the changes from this PR in the recent commits:

https://github.com/simple-login/app/actions/runs/1077620697
https://github.com/simple-login/app/actions/runs/1077788023

and so I'll close this PR.

TheLastProject:
It seems your code will only notify once though, not whenever an alias is found in a new breach like my code attempted to do. But it is better than nothing. Maybe a future version can keep track of the most recent breach notified per alias to still notify the user if their alias ends up breached a second time.

nguyenkims:
@TheLastProject yeah, that's true. An alias will be notified only once. I think it's enough as v1 and the code required is more simple to manage.

PR #516, opened 2021-06-28
Prefix suggestion is made more useful for .co.jp and .co.uk

alung:
When a prefix suggestion is requested for a website whose domain name looks ends in "co.jp" or "co.uk" (which is very common in Japan or the UK, the Simple login extension invariably suggests "co" as a new alias prefix.

This diff proposes to go one subdomain up when available in this case. For example, for "jprs.co.jp", it will suggest "jprs" instead of "co" currently.

I've submitted some feedback on Twitter in https://twitter.com/alung/status/1402559093592805377, ended up writing a PR ^^

### DIFF
--- a/app/api/views/alias_options.py
+++ b/app/api/views/alias_options.py
@@ -8,7 +8,7 @@ from app.dashboard.views.custom_alias import (
 from app.extensions import db
 from app.log import LOG
 from app.models import AliasUsedOn, Alias, User
-from app.utils import convert_to_id
+from app.utils import convert_to_id, suggest_prefix
 
 
 @api_bp.route("/v4/alias/options")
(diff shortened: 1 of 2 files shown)


nguyenkims:
@alung thanks for making the PR! I wonder if we can use this library to extract the domain name which should take into account all special cases like .co.jp, co.uk, etc?
https://github.com/john-kurkowski/tldextract

alung:
I can make the change to use this module, I like the proposal to be more generic and the private subdomains idea (though at the expense of the package size / security maybe? you make the call :) ).
I'm new to python / github projects though, is there a file where you define the third party dependencies to include to the project?

nguyenkims:
@alung poetry (https://python-poetry.org) is used to manage dependencies in the repo and to add a new library, you can run `poetry add something`.

PR #561, opened 2021-08-09
Make the "Create" button focusable with keyboard

boarwell:
Hi, I love this product and can't do without it now! Thank you for bringing me the idea of email aliases.

This PR makes the "Create" button on ["Create New Custom Alias" page](https://app.simplelogin.io/dashboard/custom_alias) focusable by a keyboard. Technically, this just replaced `<span>` with `<button type="button">`. It's not `type="submit"`, so that browsers don't natively submit.

The following figure shows what happens when I press Tab in the focused text area with the current implementation:

before

The focus jumps to the logo link.

Also, after this commit:

after

The button is focused, and the focus ring is displayed.

### DIFF
--- a/app/dashboard/templates/dashboard/custom_alias.html
+++ b/app/dashboard/templates/dashboard/custom_alias.html
@@ -94,7 +94,7 @@
 
         <div class="row">
           <div class="col p-1">
-            <span id="submit" class="btn btn-primary mt-1">Create</span>
+            <button type="button" id="submit" class="btn btn-primary mt-1">Create</button>
           </div>
         </div>
       </form>


nguyenkims:
Thanks for fixing this issue!

nguyenkims:
@boarwell unfortunately using the `button` here isn't compatible with the Bouncer library we're using to validate inputs and I have to revert this PR to keep the custom alias page working.

boarwell:
@nguyenkims 
I'm sorry for bothering you. I will look into the problem and if I can fix it, I will create a new PR.

PR #584, opened 2021-08-15
Steps for securing docker exposed ports on the internet

prashantkamdar:
I found for my self-hosted instance, there were ports that were exposed on the internet - 20381,7777,5432.

This PR adds steps in the README file to mitigate the opened ports on the internet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,7 +509,7 @@ If you have followed the steps above, there will be 3 ports exposed over the int
 You can verify the ports are open by running the following command from a different machine
 
 ```bash
-sudo nmap -sS 107.172.193.177 -p 7777,20381,5432
+sudo nmap -sS <IP-ADDR> -p 7777,20381,5432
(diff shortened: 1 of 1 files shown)


A-Nicoladie:
Be careful, on some machines, the interface is not necessarily named `eth0` (`ens3` on mine)

Maybe it would be better to use the following command instead:  
`ip route get XXX.XXX.XXX.XXX | grep -Po '(?<=dev\s)\w+' | cut -f1 -d ' '`

The full command:  
``iptables -I DOCKER-USER -i `ip route get XXX.XXX.XXX.XXX | grep -Po '(?<=dev\s)\w+' | cut -f1 -d ' '` ! -s 127.0.0.1 -j DROP``

A-Nicoladie:
Another side effect is that with this setting the DNS zone check is no longer possible

image

PR #593, opened 2021-08-19
Re: Make the "Create" button focusable with keyboard

boarwell:
This PR is a revised version of #561, fixes [the problem](https://github.com/simple-login/app/pull/561#issuecomment-899102494
). 

The following scenarios works well on my local machine:

1. (Error) Click the "Create" button with the prefix and mailboxes empty.
image

2. (Error) Click the "Create" button with all mailboxes unchecked.
image

3. (Error) Click the "Create" button with the prefix empty.
image

4. (Success) Click the "Create" button with the prefix filled and mailboxes checked.
image
image

---

In #561, the Create button was not working due to an error caused by Bouncer. It was saying that  "submit is not a function", which means that Bouncer couldn't  refer  to  the submit method of the form.

This was caused by the button having "submit" in its id, which is similar to [this issue](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#issues_with_naming_elements).

So, I replaced the button's id value with "create" and prevent it from overriding  the submit method, it has got to work well.

### DIFF
--- a/app/dashboard/templates/dashboard/custom_alias.html
+++ b/app/dashboard/templates/dashboard/custom_alias.html
@@ -69,7 +69,7 @@
         <div class="row mb-2">
           <div class="col p-1">
             <select data-width="100%"
-                    class="mailbox-select" id="mailboxes" multiple name="mailboxes">
+                    class="mailbox-select" id="mailboxes" multiple name="mailboxes" required>
               {% for mailbox in mailboxes %}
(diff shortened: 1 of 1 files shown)


nguyenkims:
@boarwell thanks for making this PR! I tested and everything looks good! Will merge the PR now, it'll be deployed shortly :).

PR #619, opened 2021-09-14
fixed wrong description of error

Dattito:
As I understand the error message, "word or alias" makes no sense. "word or uuid" would be better


nguyenkims:
@Dattito thanks for spotting and fixing this error!

PR #651, opened 2021-10-13
Create email log page

TheLastProject:
I sometimes get spam and find it difficult in my mail client to see which alias is actually receiving the spam. Having the full log would be nice to have to more quickly disable the alias

I understand this probably won't scale as-is, but I first wanted to discuss this a bit before putting a lot of time into this.

My plans for this:
- [ ] Paginate (of course)
- [ ] Dropdown filter (Forwarded / Bounced / Blocked / Sent)
- [ ] Add disable/enable alias button next to each received email

Screenshot:

image

May also be worth making the boxes on the top of the aliases screen clickable instead, to not have an extra navigation option:
image

### DIFF
--- a/app/dashboard/__init__.py
+++ b/app/dashboard/__init__.py
@@ -28,4 +28,5 @@ from .views import (
     batch_import,
     alias_transfer,
     app,
+    email_log,
 )
--- a/app/dashboard/templates/dashboard/email_log.html
+++ b/app/dashboard/templates/dashboard/email_log.html
@@ -1,0 +1,73 @@
+{% extends 'default.html' %}
+
+{% set active_page = "email_log" %}
+
+{% block title %}
+  Email Log 
+{% endblock %}
+
+{% block default_content %}
+  <div class="row">
+    <div class="col">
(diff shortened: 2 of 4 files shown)


WhamboMPS:
@TheLastProject , please check out issue [599](https://github.com/simple-login/app/discussions/599). Would this address your need (admittedly in a completely different way)?

TheLastProject:
> @TheLastProject , please check out issue [599](https://github.com/simple-login/app/discussions/599). Would this address your need (admittedly in a completely different way)?

It would work, yeah. But it seems more complex to me and easy for things to go wrong when filtering that data back out when replying causing possible data leakage to other users. But it indeed would solve my issue.

TheLastProject:
Sorry I haven't gotten back to this, been quite busy and the current way of blocking single emails that's currently implemented in SimpleLogin is good enough for my needs. I personally would still like to see a method where I don't have to check which alias I received a spam mail on before I can block the sender, which this PR solves, but I don't think I'll have time to get back to this PR any time soon. Maybe someone else can finish it :)

PR #718, opened 2021-12-02
Update cloud flare dns to hdns for hns support

chetankatke:
Visit https://handshake.org for more information about HNS


nguyenkims:
@chetankatke thanks for making the PR! Handshake seems to be a very nice project but as it's still experimental, we can't use it in the production version at the moment.

PR #751, opened 2022-01-16
minor installation instruction fixes

### LINKED ISSUE #749: A few small corrections to the self-hosting instructions
Amazing project - love it! Thanks for not only providing a self-hosting option, but also the detailed instructions - they worked like a charm other than these small fixes:

- https://github.com/simple-login/app/blob/master/README.md?plain=1#L409 needs the `sl` folder to be in the root directory so I recommend adding `cd \` before https://github.com/simple-login/app/blob/master/README.md?plain=1#L60
- Similar to the item above, https://github.com/simple-login/app/blob/master/README.md?plain=1#L402 needs the `dkim.key` file to be in the root directory so the recommendation above would work for that as well or you can add `cd \` before https://github.com/simple-login/app/blob/master/README.md?plain=1#L77
- For my install on Ubuntu 20, I needed to change https://github.com/simple-login/app/blob/master/README.md?plain=1#L375 to `/simplelogin.env` for it to work since `~` pointed to `/root`
- Your certbot link at https://github.com/simple-login/app/blob/master/README.md?plain=1#L518 no longer works. Here's what I ran on Ubuntu 20:

```
sudo apt install -y snapd
sudo snap install core; sudo snap refresh core
sudo apt remove -y certbot
sudo snap install --classic certbot
sudo ln -s /snap/bin/certbot /usr/bin/certbot
sudo certbot --nginx
```

- For postgres newbies, here's how to make the edit described on https://github.com/simple-login/app/blob/master/README.md?plain=1#L525

```
docker exec -it sl-db psql -U myuser simplelogin
UPDATE users SET lifetime = TRUE WHERE id = 1;
exit
```

- After you create your desired account(s), it's important to add these lines to `simplelogin.env` so random Internet surfers don't register and use your private local installation:

```
DISABLE_REGISTRATION=1
DISABLE_ONBOARDING=true
```

Is there a better way than restarting the server to apply those settings?

mrbluecoat:
Closes #749

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ sudo apt update && sudo apt install -y dnsutils
 Create a directory to store SimpleLogin data:
 
 ```bash
+cd /
 mkdir sl
 mkdir sl/pgp # to store PGP key
 mkdir sl/db # to store database
@@ -74,6 +75,7 @@ Setting up DKIM is highly recommended to reduce the chance your emails ending up
 First you need to generate a private and public key for DKIM:
 
 ```bash
+cd /
 openssl genrsa -out dkim.key 1024
 openssl rsa -in dkim.key -pubout -out dkim.pub.key
 ```
@@ -372,7 +374,7 @@ sudo systemctl restart postfix
 
 ### Run SimpleLogin Docker containers
 
-To run SimpleLogin, you need a config file at `~/simplelogin.env`. Below is an example that you can use right away, make sure to
+To run SimpleLogin, you need a config file at `/simplelogin.env`. Below is an example that you can use right away, make sure to
 
 - replace `mydomain.com` by your domain,
 - set `FLASK_SECRET` to a secret string, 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,10 +57,9 @@ sudo apt update && sudo apt install -y dnsutils
 Create a directory to store SimpleLogin data:
 
 ```bash
-mkdir sl
-mkdir sl/pgp # to store PGP key
-mkdir sl/db # to store database
-mkdir sl/upload # to store quarantine emails
+mkdir -p /sl/pgp # to store PGP key
```
nguyenkims, line 60: I'm not sure using `/sl`, `/dkim.key`, etc as paths will suit most users as they might want to run SimpleLogin with the least privileges. I'd recommend to not change the paths for these files as we might need to change the commands for upgrading SimpleLogin as well (in the https://github.com/simple-login/app/blob/e5a7aeb3fb0db89707212c75315dd7c8401f386b/docs/upgrade.md#L1-L0)

nguyenkims (commented):
The PR looks good, thanks for making it! I just let a comment about changing the paths.

README.md
mrbluecoat: I originally had it as your version but SimpleLogin database failed after installing on a Linux OS where root is the default user. The login page loads but when you enter credentials it says an error occured. I then tried installing to `/opt/sl` as a regular user but your Docker images weren't compatible with that path.

mrbluecoat: Sure, I'll restore those path edits but I'm curious if your install using those instructions survives a server reboot?

nguyenkims: Yes they work after a server reboot. I don't see why they wouldn't work afer a reboot though.

nguyenkims (approved):
Thanks for making this PR!

PR #753, opened 2022-01-20
Invalid TOTP and recovery code email notifications

FozzieHi:
This PR adds notifications for invalid TOTP codes where the user has correctly entered their email and password.

This is what the email currently looks like:
image

This is just a draft though as we obviously do not want to send an email every time this happens, we'd need to implement some kind of rate limit (maybe 1 email per day). I'm wondering if send_email_with_rate_control is the best way to achieve this?

### DIFF
--- a/app/api/views/auth_mfa.py
+++ b/app/api/views/auth_mfa.py
@@ -6,6 +6,7 @@ from itsdangerous import Signer
 from app.api.base import api_bp
 from app.config import FLASK_SECRET
 from app.db import Session
+from app.email_utils import send_invalid_totp_login_email
 from app.log import LOG
 from app.models import User, ApiKey
 
@@ -53,6 +54,7 @@ def auth_mfa():
 
     totp = pyotp.TOTP(user.otp_secret)
     if not totp.verify(mfa_token):
+        send_invalid_totp_login_email(user, "TOTP")
         return jsonify(error="Wrong TOTP Token"), 400
 
     ret = {"name": user.name or "", "email": user.email}
--- a/app/auth/views/mfa.py
+++ b/app/auth/views/mfa.py
@@ -14,9 +14,9 @@ from flask_wtf import FlaskForm
 from wtforms import BooleanField, StringField, validators
 
 from app.auth.base import auth_bp
-from app.config import MFA_USER_ID, URL
+from app.config import MFA_USER_ID, URL, ALERT_INVALID_TOTP_LOGIN
 from app.db import Session
-from app.email_utils import send_email, render
+from app.email_utils import send_email_with_rate_control, render
 from app.extensions import limiter
 from app.models import User, MfaBrowser
 
@@ -92,11 +92,14 @@ def mfa():
             return response
 
         else:
-            send_email(
+            send_email_with_rate_control(
+                user,
+                ALERT_INVALID_TOTP_LOGIN,
                 user.email,
                 "There was an unsuccessful login on your SimpleLogin account",
                 render("transactional/invalid-totp-login.txt"),
                 render("transactional/invalid-totp-login.html"),
+                1,
             )
             flash("Incorrect token", "warning")
             # Trigger rate limiter
--- a/app/auth/views/mfa.py
+++ b/app/auth/views/mfa.py
@@ -16,6 +16,7 @@ from wtforms import BooleanField, StringField, validators
 from app.auth.base import auth_bp
 from app.config import MFA_USER_ID, URL
 from app.db import Session
+from app.email_utils import send_email, render
 from app.extensions import limiter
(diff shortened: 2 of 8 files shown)


FozzieHi:
This is what the email looks like now with a link to the dashboard to change the password:
image

FozzieHi:
Let me know if I'm missing any endpoints here, I don't think there is a recovery code API endpoint?

FozzieHi:
Should be good for a review now (I don't think there's a request review button unless you have write access?)

nguyenkims:
Sorry for the late reply, the PR looks really good, thanks for making it! `send_email_with_rate_control` is indeed the one to use here. We'll test on staging first and if everything goes smooth (it should), we'll merge & deploy it on production.

netcarver:
@FozzieHi  @nguyenkims

George, thank you for this PR - I really appreciate the addition of this email but Son, please could we go with the initial version of login attempt notification email **without the link to simplelogin**.  This is better because it forces SimpleLogin users to go to their account by launching their browser and typing in the SimpleLogin url themselves, helping train users not to click links in emails - especially password reset emails, which can easily be spoofed.

If the attempt notification emails from SimpleLogin **do** include a link, then phishing attempts can be made by just having the link in the fake notification phishing email point to a SimpleLogin look-alike password reset form. 

If we send notification emails which **do not** include links and we make this known to users, then phishing emails that mirror these reset-attempt emails will be immediately obvious by having to contain a link.

FozzieHi:
@netcarver Thanks for your comments and thank you for raising some good points.

I think we really need to strike a balance between convenience and security here. I think having a link to the dashboard to change your password is more likely to get a user to change their password quickly in the event of this happening.

Additionally, I don't think this attack could realistically be pulled off on a mass scale as an attacker would have to get quite a sizeable list of email addresses of SimpleLogin users. Even on a smaller more targeted scale I think there some fairly major roadblocks to this:
- The attack is realistically limited to users with TOTP enabled (and not WebAuthn only), because if the user didn't have TOTP enabled then the email would make little sense.
- The user would have to miss fairly important universal signs of a phishing email, such as the domain of the sender as well as the dashboard address in the email.
- The attacker would have to recreate the SimpleLogin dashboard (although this is easier as it is open source) and as the user has TOTP enabled, they would have to relay the TOTP code to the main SimpleLogin API within the TOTP time period.

Additionally, how would removing the link in our email affect phishing emails? It's unlikely the user would have received one of our emails before an attack would occur so they would have nothing to compare it against. Unless they have enough due diligence to check the source code and in that case they would probably identify the email as a phishing email anyway through the email sender domain and the dashboard address.

Although I think we could do some things to make our own email seem more legitimate, perhaps including the account's first name in the start of the email?

nguyenkims:
@netcarver thanks for your comment, I think spoofing a SimpleLogin email is quite difficult as SimpleLogin has a strict DMARC, SPF, DKIM policy. A good mailbox service will display a warning button to this spoofed email if it ever makes to user's inbox.

FozzieHi:
@nguyenkims To be clear, I think @netcarver is referring to phishing via a different domain instead of spoofing SimpleLogin's domain.

netcarver:
@FozzieHi Hello George. Thank you for your reply and my apologies. Following that through, I have to admit I missed the TOTP involvement there. How blind I am. I wrote that up thinking you were talking simply about notifying on failed login attempts. So, yes, I agree that the chances of a successful attack via this vector are low though not quite zero.

> Additionally, how would removing the link in our email affect phishing emails? **It's unlikely the user would have received one of our emails before an attack would occur so they would have nothing to compare it against.** Unless they have enough due diligence to check the source code and in that case they would probably identify the email as a phishing email anyway through the email sender domain and the dashboard address.

Regarding my highlighting in your quote - that's a great point regardless of the situation and I was trying to hit on that in my initial reply, though in the context of a wide phishing net using fake notification emails. 

> If we send notification emails which do not include links and **we make this known to users,** then phishing emails that mirror these reset-attempt emails will be immediately obvious by having to contain a link.

In that case receiving such a notification from SL with a link would be an immediate flag of a phishing attempt.

Anyway, coming back to confidence-increasing measures. Yes, adding an additional field (maybe not limited the username) from the user account to the notification email is a good idea. How about signing the message with SLs PGP key (if your PR doesn't already do that) to allow it to be checked vs their pubkey.  Not sure if this would speed up any actions taken in valid cases though.

Thanks again.

FozzieHi:
@netcarver Good idea on the PGP front, signing could definitely be a good solution for everyone, but maybe we also want to look into having an account-wide PGP key for encrypting all (or just security emails) from SimpleLogin. I'd have to familiarize myself with the PGP code to implement this though unless anyone else is willing to.

PR #756, opened 2022-01-21
Add Pwned Passwords integration to check passwords on login and registration

FozzieHi:
This PR adds Pwned Password checking on login and registration.

On registration if a password is in Pwned Passwords we block registration with that password.
On login if a password is in Pwned Passwords we send the user an email notifying them to change their password.

This PR loosely depends on the dashboard change password ID in #753 so we can link to the correct section of the settings page in the email. Additionally there's been a function added to email_utils so this PR will probably need to be rebased when #753 is merged.


FozzieHi:
We'll obviously need to change the tests that use "password" for authentication, maybe add some kind of a bypass?

nguyenkims:
Now that https://github.com/simple-login/app/pull/753 is merged, I think we need to rebase this PR indeed.

FozzieHi:
@nguyenkims Seemed simpler just to do a merge instead of a rebase as this probably needs a squash anyway!

app/email_utils.py
```diff
@@ -173,6 +174,18 @@ def send_change_email(new_email, current_email, link):
     )
 
 
+def send_pwnedpasswords_email(user, amount):
+    send_email_with_rate_control(
```
nguyenkims, line 196: to avoid user from having this email every time they log in, I'd recommend using `send_email_at_most_times` instead.

nguyenkims (commented):
The PR looks really good but people might be afraid to know that their passwords are sent to an external service (albeit a reputable one). I'd rather only check that at registration. For existing accounts, their password managers might already have a similar check.

app/email_utils.py
FozzieHi: I think it's good to have a reminder email though, maybe we could set nb_day to something like 30 or 90 days?

FozzieHi:
We're not sending the password though.

We're sending the first 5 characters of a SHA-1 hash of their password, this returns a list of hashes and the amount of times it's been seen (along with some other hashes which aren't in the dataset due to padding to reduce the likelihood of a MITM from determining the first 5 characters of the hash based on the response size https://www.troyhunt.com/enhancing-pwned-passwords-privacy-with-padding/).

We then check if our hash matches any of the returned hashes from the API locally.

From Pwned Password's perspective, all they know is the first 5 characters of *someone*'s SHA-1 password hash. They don't even know if that is in their database of compromised passwords. There's a great explanation of this on Troy's blog https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/ and an excellent video by Computerphile https://www.youtube.com/watch?v=hhUb5iknVJs.

PR #760, opened 2022-01-23
Add a button to delete all API Keys

FozzieHi:
This PR adds a button to delete all API Keys.
image

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1966,6 +1966,10 @@ class ApiKey(Base, ModelMixin):
 
         return super().create(user_id=user_id, name=name, code=code, **kwargs)
 
+    @classmethod
+    def deleteall(cls, user_id):
+        Session.query(cls).filter(cls.user_id == user_id).delete()
+
 
 class CustomDomain(Base, ModelMixin):
     __tablename__ = "custom_domain"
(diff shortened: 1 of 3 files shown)


app/models.py
```diff
@@ -1966,6 +1966,10 @@ def create(cls, user_id, name=None, **kwargs):
 
         return super().create(user_id=user_id, name=name, code=code, **kwargs)
 
+    @classmethod
+    def deleteall(cls, user_id):
```
nguyenkims, line 1970: I think it'd be more correct grammatically to name the method `delete_all` instead of `deleteall` :)

templates/dashboard/api_key.html
```diff
@@ -22,6 +22,16 @@ <h1 class="h3"> API Keys </h1>
         ️API Keys should be kept secret and treated like passwords, they can be used to gain access to your account.
       </div>
 
+      {% if api_keys|length > 0 %}
```
nguyenkims, line 25: I'd suggest to move the button to the bottom of the page to not "encourage" users to use it immediately when they arrive to the page.

nguyenkims (commented):
The PR looks good! Just added some comments :)

app/models.py
FozzieHi: Done

templates/dashboard/api_key.html
FozzieHi: Yup, I thought about doing this but thought it'd be a bit hard to find if you had a load of API Keys, but I guess the majority of users have only a few.

nguyenkims:
Sorry for the late response, the PR looks good and it has been merged. It's going to deployed soon :).

Issue #752, opened 2022-01-20
Testing error - TypeError: required field "lineno" missing from alias sl-test-db


FozzieHi:
Hello, I've just setup the development environment but I'm getting the following error when running the run-test script.

```
TypeError: required field "lineno" missing from alias
sl-test-db
```

I don't think I've changed anything to cause this, and I can run the development environment fine (as far as I can tell), is this a known issue?

FozzieHi:
Now it seems like there's a delay at the start with "Error: No such container: sl-test-db", then the same error occurs.

FozzieHi:
I'm guessing the delay is the sleep you added with the commit. However the no such container error occurs **before** this delay.

FozzieHi:
Right I've managed to fix this now and the tests run as expected. I recreated the virtualenv using Python 3.7.9 instead of the latest version so I think that might've been the issue.

PR #763, opened 2022-01-25
Add support for automatic alias copying after creation.

FozzieHi:
This PR adds a setting for automatically copying the email address of the alias after creation in the dashboard (suggested in #514).

Note: This doesn't currently work in Firefox due to the following error:
> Uncaught (in promise) DOMException: Clipboard write was blocked due to lack of user activation.

It seems to me that there is some certain criteria to allow clipboard copying on Firefox and simply having a window in focus is not enough? Or maybe the window is not considered to be in focus directly after a reload?

Also let me know if this is the correct way to Auto Generate Alembic migrations, I haven't even used Alembic before this and I'm wondering if you want them in a PR or if you want to generate them yourself at merge time.


FozzieHi:
@nguyenkims Re: Alembic Migrations. I didn't want to run the script in case of similar errors to #752 so I just ran "alembic revision --autogenerate" on its own. It seems to have generated the revision correctly so is this is fine?

templates/dashboard/index.html
```diff
@@ -587,5 +587,15 @@
     {% endif %}
 
     $('.highlighted').tooltip("show");
+
+    {% if current_user.automatic_alias_copy %}
+      {% if copy_alias_text %}
+        {% for alias_info in alias_infos %}
+          {% if alias_info.alias.id == highlight_alias_id %}
+            navigator.clipboard.writeText("{{alias_info.alias.email}}").catch(() => console.log('Failed to automatically copy alias address to clipboard.'));
```
nguyenkims, line 595: I think `navigator.clipboard.writeText` has to result from a user action (e.g. clicking on a button). That might explain why this doesn't work in all browsers. I think copying the new alias copied to the clipboard needs to be done on the custom alias page. This is tricky though as we might need to create a new alias via an Ajax call and then call `navigator.clipboard.writeText`.

nguyenkims (commented):
Hey I did some tests and the copy to clipboard doesn't work on all modern browsers indeed. Just left a comment about a way to make it work everywhere.

templates/dashboard/index.html
FozzieHi: Yup, that's what I thought as well. In Firefox I think there definitely has to be some kind of user interaction but I guess either the page reload takes too long or it cancels after a page reload.

In Chromium, I think copy permission is granted to the active tab, according to [this article](https://web.dev/async-clipboard/).

What other browsers have you tested? I'm assuming this also works on Edge as it's based on Chromium but I don't know if it works on WebKit based browsers such as Safari.

I agree doing it on the same page will be better, but if most modern browsers support it then I think it's alright to merge it with an Experimental tag while we work on the more difficult Ajax call?

nguyenkims: I've tested on Safari and Firefox. I'm not sure yet about whether automatically copying an alias to the clipboard is the best default behavior as some users might not want to have their clipboard altered implicitly.

FozzieHi: It's not the default behaviour, it's off by default and is marked experimental with browser compatibility. Feel free to close though if you think this should be implemented better 👍

nguyenkims: I will close this PR for now, we can always reopen it when we implement the automatic copy to clipboard feature.

PR #768, opened 2022-01-26
Add the user's name to TOTP invalid login email notifications.

FozzieHi:
This PR adds the user's name to the TOTP invalid login email notifications (#753) (see [this comment](https://github.com/simple-login/app/pull/753#issuecomment-1022096915) for more details)

### DIFF
--- a/app/email_utils.py
+++ b/app/email_utils.py
@@ -181,7 +181,7 @@ def send_invalid_totp_login_email(user, totp_type):
         user.email,
         "Unsuccessful attempt to login to your SimpleLogin account",
         render(
-            "transactional/invalid-totp-login.txt",
+            "transactional/invalid-totp-login.txt.jinja2",
             name=user.name,
             type=totp_type,
         ),
--- a/app/email_utils.py
+++ b/app/email_utils.py
@@ -182,13 +182,13 @@ def send_invalid_totp_login_email(user, totp_type):
         "Unsuccessful attempt to login to your SimpleLogin account",
         render(
             "transactional/invalid-totp-login.txt",
+            name=user.name,
             type=totp_type,
-            name=user.name
         ),
         render(
             "transactional/invalid-totp-login.html",
+            name=user.name,
             type=totp_type,
-            name=user.name
         ),
         1,
     )
--- a/app/email_utils.py
+++ b/app/email_utils.py
@@ -183,10 +183,12 @@ def send_invalid_totp_login_email(user, totp_type):
         render(
             "transactional/invalid-totp-login.txt",
             type=totp_type,
+            name=user.name
         ),
         render(
             "transactional/invalid-totp-login.html",
             type=totp_type,
+            name=user.name
         ),
         1,
     )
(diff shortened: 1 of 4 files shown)


app/email_utils.py
```diff
@@ -182,10 +182,12 @@ def send_invalid_totp_login_email(user, totp_type):
         "Unsuccessful attempt to login to your SimpleLogin account",
         render(
             "transactional/invalid-totp-login.txt",
+            name=user.name,
```
nguyenkims, line 185: as `user.name` can be None, I think we can use `name = user.name or ""` here to make sure the email never displays `None`

nguyenkims (approved):
Sorry for the late reply. The PR looks good, there's just maybe a small improvement there.

app/email_utils.py
FozzieHi: Done, I think it's better to change it in the email template though so we don't end up with any formatting issues.

templates/emails/transactional/invalid-totp-login.html
```diff
@@ -1,6 +1,7 @@
 {% extends "base.html" %}
 
 {% block content %}
+  {{ render_text("Hello," ~ (" " ~ name ~ "," if name)) }}
```
nguyenkims, line 4: Hey I think this is actually harder to read than using `name = user.name or ""` :)

nguyenkims (commented):
Hey just added a comment

templates/emails/transactional/invalid-totp-login.html
FozzieHi: Right, but if we used that then the text would be "Hello, ," if the user.name was None.

nguyenkims: That's right. For this reason in other emails we only put "Hi" without the name :). As most SL users don't set the name field, this option seems to be enough.

FozzieHi: Yup, but as outlined in [these comments](https://github.com/simple-login/app/pull/753#issuecomment-1022096915), it's good to include something SimpleLogin would know in these emails.

If readability is that much of an issue, we could just expand it into an if/else, for example:
HTML
```jinja2
{% if name %}
  {{ render_text("Hello, " ~ name ~ ",") }}
{% else %}
  {{ render_text("Hello,") }}
{% endif %}
```

Plaintext
```jinja2
{% if name %}
Hello, {{ name }},
{% else %}
Hello,
{% endif %}
```

nguyenkims: Sorry to not participate in the discussion earlier but I don't think including the user.name is really helpful as most users don't have their name set and other emails don't use this information neither. So I think the current TOTP warning emails are already good enough :).

FozzieHi: @nguyenkims Alright, feel free to close my PRs earlier in the future if you feel like this, I'm not going to be offended ;)

nguyenkims: Thanks :)

PR #784, opened 2022-02-04
Fix deprecation warnings.

FozzieHi:
This PR uses resolve() instead of query() to resolve DNS queries. It also removes the deprecation tag from tests, I don't see any point in deprecating a test? Is there a valid reason for this @nguyenkims?


nguyenkims:
@FozzieHi hey thanks for making this PR! 

> I don't see any point in deprecating a test?

This is because these tests test deprecated endpoints :).

FozzieHi:
@nguyenkims Right, but inherently a test also tests other parts of the codebase. I don't think it's needed to designate a specific test as testing a deprecated feature, although I could remove that part from the PR if you want.

PR #787, opened 2022-02-06
Update testing suite and refactor.

FozzieHi:
This PR updates pytest, pytest-cov, pre-commit, black and flake8. It also reformats some files to comply with the new updates.

pre-commit is changed to match the current installed version instead of running it against the latest versions.

.python-version is also added to the .gitignore and deleted from the repo so it can be changed locally to pyenv virtualenv names.

`test_can_be_used_as_personal_email` is only skipped when using GitHub Actions.

Updated the GitHub Action to use install-poetry and setup caching for the entire virtualenv.
Currently on non-master branches we are averaging 7m-8m run times.
[The first build for this (with a cache miss) ran in ~6m.](https://github.com/FozzieHi/app/actions/runs/1802745600)
[The second build (with a cache hit) ran in ~4m30s.](https://github.com/FozzieHi/app/actions/runs/1802759613)

Adds flake8-bugbear for more lints and for adding behavior similar to black's line-length lint (see https://github.com/PyCQA/flake8-bugbear#opinionated-warnings, B950).
These lints seem fine, but is this bare except necessary? @nguyenkims https://github.com/simple-login/app/blob/master/app/dashboard/views/domain_detail.py#L443. I've replaced it with an `except Exception:` in this PR, if there's a reason for it not to be that or a more broad `except BaseException:` then please let me know!


FozzieHi:
@nguyenkims Would it be possible to get this merged before other PRs, if you're happy with it? I'd rather have the authors format their own code with the update rather than retrospectively doing it here. This would also reduce some potential [git blame issues](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame).

.flake8
```diff
@@ -1,18 +1,17 @@
 [flake8]
-ignore = 
-    # line length is handled by black
-    E501,
-    # W503 & E203 are not PEP 8 compliant and conflicts with black
-    W503,
```
nguyenkims, line 6: Hey `W503` seems to be missing in the new flake8 config?

.pre-commit-config.yaml
```diff
@@ -1,10 +1,10 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: stable
+    rev: 22.1.0
```
nguyenkims, line 3: 👌

tests/test_email_utils.py
```diff
@@ -64,8 +64,9 @@ def test_email_belongs_to_alias_domains():
     assert not can_create_directory_for_address("hey@d3.test")
 
 
-@pytest.mark.skip(
-    reason="this test requires DNS lookup that does not work on Github CI"
+@pytest.mark.skipif(
+    "GITHUB_ACTIONS_TEST" in os.environ,
```
nguyenkims, line 68: Nice!

nguyenkims (commented):
The PR looks good, thanks for making it! I just left some questions.

.flake8
FozzieHi: We've replaced `ignore` with `extend-ignore`. `ignore` replaces the default ignore list with our own, `extend-ignore` adds on our own list to the default list. `W503` is in the default ignore list so we don't need to add it on ourselves.

app/email_utils.py
```diff
@@ -823,7 +823,7 @@ def copy(msg: Message) -> Message:
         LOG.w("deepcopy fails, try string parsing")
         try:
             return message_from_string(msg.as_string())
-        except (UnicodeEncodeError, KeyError, LookupError):
```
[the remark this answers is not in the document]
FozzieHi: `KeyError` is already caught by `LookupError`. This was caught by [flake8-bugbear's B014 lint](https://github.com/PyCQA/flake8-bugbear#list-of-warnings).

PR #792, opened 2022-02-09
Create support tickets via zendesk

acasajus:
/dashboard/support will provide a form to generate a ticket in zendesk.

To use: set ZENDESK_HOST to the zendesk domain for your ordanization


app/config.py
```diff
@@ -412,3 +412,5 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 
 PHONE_PROVIDER_2_HEADER = os.environ.get("PHONE_PROVIDER_2_HEADER")
 PHONE_PROVIDER_2_SECRET = os.environ.get("PHONE_PROVIDER_2_SECRET")
+
+ZENDESK_HOST=os.environ.get('ZENDESK_HOST', 'noone.zendesk.com')
```
nguyenkims, line 416: Should we also need a ZENDESK secret/token here?

nguyenkims, line 416: I wonder if we should use `os.environ.get('ZENDESK_HOST')` without providing a default value here? I think by default `ZENDESK_HOST` can be None which means the support page can't be used.

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+from app.models import Mailbox
+from app.config import ZENDESK_HOST
+
+VALID_MIME_TYPES = ['text/plain', 'message/rfc822']
+
+
+@dashboard_bp.route("/support", methods=["GET"])
+@login_required
+def show_support_dialog():
+    mailbox = Mailbox.get(current_user.default_mailbox_id)
```
nguyenkims, line 21: We usually use the user email address (`current_user.email`) to get in touch with them so I think we can replace `mailbox.email` here by `current_user.email`.  And as `current_user` is directly available in the html template, we don't then need to pass a variable to the template.

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+def upload_file_to_zendesk(file: FileStorage) -> Union[None, str]:
+    if file.mimetype not in VALID_MIME_TYPES and not file.mimetype.startswith('image/'):
+        flash('File {} is not an image, text or an email'.format(file.filename), "warning")
+        return None
+    escaped_filename = urllib.parse.urlencode({'filename': file.filename})
+    url = 'https://{}/api/v2/uploads?{}'.format(ZENDESK_HOST, escaped_filename)
+    headers = {'content-type': file.mimetype}
+    response = requests.post(url, headers=headers, data=file.stream)
+    if response.status_code != 201:
+        if response.status_code == 401 or 422:
```
nguyenkims, line 34: I think this line should rather be `if response.status_code in (401, 422)`

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+            LOG.debug('Could not authenticate')
+            return None
```
nguyenkims, line 36: Nits: we can just call `return` (without `None`) as None is the default return value for a function in Python.

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+        else:
+            LOG.debug('Problem with the request. Status ' + str(response.status_code))
```
nguyenkims, line 38: Nits: we can use `LOG.d` instead of `LOG.debug` ( a bit shorter to write 😅)

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@dashboard_bp.route("/support", methods=["GET"])
+@login_required
+def show_support_dialog():
+    mailbox = Mailbox.get(current_user.default_mailbox_id)
+    return render_template("dashboard/support.html", ticketEmail=mailbox.email)
+
+
+def upload_file_to_zendesk(file: FileStorage) -> Union[None, str]:
```
nguyenkims, line 25: This function returns a token which isn't obvious to tell from the method name, it's maybe useful to add a small doc here.

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+            LOG.debug('Could not authenticate')
+            return None
+        else:
+            LOG.debug('Problem with the request. Status ' + str(response.status_code))
+            return None
+    data = response.json()
+    return data['upload']['token']
+
+
+def create_zendesk_request(email: str, contents: str, files: [FileStorage]) -> bool:
```
nguyenkims, line 44: Nits: should `contents` be `content` (singular)?

app/dashboard/views/support.py
```diff
@@ -0,0 +1,93 @@
...
+    flash("Ticket was created. You should receive an email notification", "success")
+    LOG.debug('Ticket created')
+    return True
+
+
+@dashboard_bp.route("/support", methods=["POST"])
+@login_required
+def process_support_dialog():
+    contents = request.form.get("ticketContents") or ""
+    email = request.form.get("ticketEmail") or ""
```
nguyenkims, line 84: Nits: as a convention, data sent from the web is usually separated by an underscore, so `ticket_email` instead of `ticketEmail`

templates/dashboard/support.html
```diff
@@ -0,0 +1,101 @@
+{% extends 'default.html' %}
+
+{% set active_page = None %}
```
nguyenkims, line 3: I think the `active_page` can be set to `dashboard`

templates/dashboard/support.html
```diff
@@ -0,0 +1,101 @@
...
+{% block head %}
+    <style>
+        .card-title {
+            font-size: 22px;
+            font-weight: 600;
+            margin-bottom: 3px;
+        }
+    </style>
+
+    <script src="{{ url_for('static', filename='node_modules/vue/dist/vue.min.js') }}"></script>
```
nguyenkims, line 18: Usually js part is included in the `{% block script %}`. This way we don't have to use `$( document ).ready(function()`. An example of vuejs usage is https://github.com/simple-login/app/blob/1fd9a344d4d728eb5da3e232b64881be5279ca61/static/js/index.js#L256

templates/dashboard/support.html
```diff
@@ -0,0 +1,101 @@
...
+    <script>
+        $( document ).ready(function() {
+            var app = new Vue({
+                el: '#aliasGroup',
+                data: {
+                    ticketEmail: document.querySelector("input[name=ticketEmail]").value
```
nguyenkims, line 24: we can use Jinja2 variable here, for example `ticketEmail: "{{ current_user.email }}"`

templates/dashboard/support.html
```diff
@@ -0,0 +1,101 @@
...
+                    <div class="mt-4 mb-5">
+                        <label for="issueDescription" class="form-label">What happened?</label>
+                        <textarea class="form-control" name="ticketContents" id="issueDescription" rows="3" placeholder="Please provide as much information as possible. For example which alias(es), mailbox(es) ar affected, if this is a persistent issue...">{{- ticketContents or '' -}}</textarea>
+                    </div>
+                    <div class="mt-5 font-weight-bold">
+                        Attach files to support request
+                    </div>
+                    <div class="mt-1 text-muted">Only images, text and emails are accepted</div>
+                    <div class="custom-file mt-2">
+                        <input type="file" class="custom-file-input" name="ticketFiles" id="ticketFileGroup" multiple="multiple">
```
nguyenkims, line 73: Do you know if we can create a ticket without attachments?

templates/footer.html
```diff
@@ -66,6 +66,7 @@ <h3 class="h4 text-white">Company</h3>
                  href="https://github.com/simple-login/app/projects/1">Roadmap</a></li>
           <li><a class="list-group-item text-white footer-item"
                  href="https://simplelogin.io/contact/">Contact Us</a></li>
+            <li><a class="list-group-item text-white footer-item" href="/dashboard/support">Support</a></li>
```
nguyenkims, line 69: I think the support link can live in the header, for example 
image

nguyenkims (commented):
Hey the PR looks good! I haven't been able to test the whole flow though. I just left some (nitpicking) comments.

app/config.py
acasajus: No need. Zendesk tickets can be opened by anyone without having a zendesk account.

templates/dashboard/support.html
acasajus: Yes, if the user doesn't upload any attachment, the ticket won't have any.

app/config.py
acasajus: Ok. I'll change so if the HOST is None then we show a page saying the support is disabled.

templates/footer.html
acasajus: This is something I wanted to ask you. Perfect. I'll change into that.

templates/dashboard/support.html
acasajus: :+1:

app/dashboard/views/support.py
acasajus: Ok. I'll replace `mailbox.email` with `current_user.email` But I'd rather keep the variable in the template. It's used to re-fill the template if the submission is missing some value in the POST route.

app/dashboard/views/support.py
acasajus: OOps. You're totally right.

templates/dashboard/support.html
acasajus: Ok..

app/dashboard/views/support.py
acasajus: Updated the method name to make it clear

app/dashboard/views/support.py
acasajus: I'll update that

app/dashboard/views/support.py
acasajus: I've changed it to `content`.

FozzieHi:
Thanks for this! There's a few formatting issues that should be resolved, see [the docs](https://github.com/simple-login/app/blob/master/CONTRIBUTING.md#pull-request) for more info. Could we also approve a CI run on this PR? @nguyenkims

templates/dashboard/support_disabled.html
```diff
@@ -0,0 +1,34 @@
...
+    </style>
+{% endblock %}
+
+{% block default_content %}
+
+    <div class="col pb-3">
+        <!-- Current plan -->
+        <div class="card">
+            <div class="card-body">
+                <div class="card-title mb-3">Support form is disabled. Please configure Zendesk integration.</div>
```
FozzieHi, line 25: Could we instead use an error through `flash("Support form is disabled. Please configure Zendesk integration.", "error")` through the backend and then redirect the user back to the dashboard rather than having a dedicated page for it?

Example:
image

Although the "error" modal will be red instead of orange.

FozzieHi (changes requested):

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+    headers = {'content-type': 'application/json'}
+    response = requests.post(url, data=json.dumps(data), headers=headers)
+    if not check_zendesk_response_status(response.status_code):
+        return False
+    flash("Ticket was created. You should receive an email notification", "success")
+    LOG.debug('Ticket created')
+    return True
+
+
+@dashboard_bp.route("/support", methods=["POST"])
```
FozzieHi, line 83: It'd be good if we could rate limit this to something like 2/hour using `@limiter.limit("2/hour")`

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+def show_support_dialog():
+    if not ZENDESK_HOST:
+        return render_template("dashboard/support_disabled.html")
+    return render_template("dashboard/support.html", ticket_email=current_user.email)
+
+
+def check_zendesk_response_status(response_code: int) -> bool:
+    if response_code != 201:
+        if response_code in (401 or 422):
+            LOG.debug('Could not authenticate')
```
nguyenkims, line 29: Nits: we can use `LOG.d` here :)

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+        else:
+            LOG.debug('Problem with the request. Status {}'.format(response_code))
```
nguyenkims, line 31: ⬆️

app/dashboard/views/support.py
nguyenkims, line 29: I think this error only happens when the Zendesk integration is wrongly configured, in this case we can use `LOG.e` so we'll be notified (`LOG.e` will create a new Sentry alert)

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+@login_required
+def show_support_dialog():
+    if not ZENDESK_HOST:
+        return render_template("dashboard/support_disabled.html")
+    return render_template("dashboard/support.html", ticket_email=current_user.email)
+
+
+def check_zendesk_response_status(response_code: int) -> bool:
+    if response_code != 201:
+        if response_code in (401 or 422):
```
nguyenkims, line 28: This should rather be `response_code in (401,402)` right?

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+    }
+    url = 'https://{}/api/v2/requests.json'.format(ZENDESK_HOST)
+    headers = {'content-type': 'application/json'}
+    response = requests.post(url, data=json.dumps(data), headers=headers)
+    if not check_zendesk_response_status(response.status_code):
+        return False
+    flash("Ticket was created. You should receive an email notification", "success")
+    LOG.debug('Ticket created')
```
nguyenkims, line 79: We can remove this line I think

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+    return True
+
+
+@dashboard_bp.route("/support", methods=["POST"])
+@login_required
+def process_support_dialog():
+    if not ZENDESK_HOST:
+        return render_template("dashboard/support_disabled.html")
+    content = request.form.get("ticket_content") or ""
```
nguyenkims, line 88: We don't need the `or ""` here right?

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+    email = request.form.get("ticket_email") or ""
+    if not content:
+        flash("Please add a description", "warning")
```
nguyenkims, line 91: I think we can `error` level for these error messages :)

app/dashboard/views/support.py
```diff
@@ -0,0 +1,98 @@
...
+        return render_template("dashboard/support.html", ticket_email=email)
+    if not email:
+        flash("Please add an email", "warning")
+        return render_template("dashboard/support.html", ticket_content=content)
+    if create_zendesk_request(email, content, request.files.getlist('ticket_files')):
```
nguyenkims, line 96: Maybe in a future PR, I think we can add some checks on the uploaded file to make sure they aren't too big.

templates/dashboard/support.html
```diff
@@ -0,0 +1,98 @@
...
+            }
+            $(this).next('.custom-file-label').html(files.join(', '));
+        });
+    </script>
+{% endblock %}
+
+{% block default_content %}
+
+    <div class="col pb-3">
+        <!-- Current plan -->
```
nguyenkims, line 50: This comment can be removed i think :)

nguyenkims (commented):
The PR looks really good now! Just left a bunch of rather nitpicking comments though you we might need to make sure only SL can create Zendesk ticket.

app/dashboard/views/support.py
acasajus: I misunderstoond deduct as 'deduct from number of requests' not as deduct from available requests. I agree with you then :)

app/dashboard/views/support.py
acasajus: If it's the same. `debug` is more explicit and easier to read.

app/dashboard/views/support.py
acasajus: :arrow_down: :)

acasajus: Ok

templates/dashboard/support.html
```diff
@@ -0,0 +1,98 @@
...
+            font-weight: 600;
+            margin-bottom: 3px;
+        }
+    </style>
+{% endblock %}
+
+{% block script %}
+    <script>
+        new Vue({
+            el: '#aliasGroup',
```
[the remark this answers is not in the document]
acasajus: :+1:

templates/dashboard/support.html
acasajus: OOps. Removed.

templates/dashboard/support.html
```diff
@@ -0,0 +1,98 @@
...
+                        let data = await result.json();
+                        this.ticket_email = data.alias;
+                    }
+                }
+            }
+        });
+
+        $('.custom-file input').change(function (e) {
+            var files = [];
+            for (var i = 0; i < $(this)[0].files.length; i++) {
```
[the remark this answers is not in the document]
acasajus: ok

app/config.py
```diff
@@ -412,3 +412,5 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 
 PHONE_PROVIDER_2_HEADER = os.environ.get("PHONE_PROVIDER_2_HEADER")
 PHONE_PROVIDER_2_SECRET = os.environ.get("PHONE_PROVIDER_2_SECRET")
+
+ZENDESK_HOST=os.environ.get('ZENDESK_HOST')
```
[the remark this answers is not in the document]
acasajus: It was set to allow anyone to open a ticket in zendesk. I've restricted it and now we require a token. WebWidget from zendesk won't work but we're not going to be using it afaik so it should be fine to disable anonymous submission.

app/dashboard/views/support.py
acasajus: Done.

app/dashboard/views/support.py
acasajus: Done

app/dashboard/views/support.py
FozzieHi: `LOG.d` is the widely used shortcut, no other place in the codebase uses `LOG.debug`, `LOG.e` is actually short for `exception` instead of `error`.

nguyenkims (approved):

app/dashboard/views/support.py
```diff
@@ -0,0 +1,110 @@
...
+            },
+        }
+    }
+    url = "https://{}/api/v2/requests.json".format(ZENDESK_HOST)
+    headers = {"content-type": "application/json"}
+    auth = ("{}/token".format(email), ZENDESK_API_TOKEN)
+    response = requests.post(url, data=json.dumps(data), headers=headers, auth=auth)
+    if not check_zendesk_response_status(response.status_code):
+        return False
+    LOG.debug("Ticket created")
```
nguyenkims, line 82: Nits: as "There should be one-- and preferably only one --obvious way to do it." (Zen of Python) :), I think it's better to use the abbreviations of LOG, i.e. LOG.d instead of LOG.debug.

app/dashboard/views/support.py
```diff
@@ -0,0 +1,110 @@
...
+    return True
+
+
+@dashboard_bp.route("/support", methods=["POST"])
+@login_required
+@limiter.limit(
+    "2/hour", deduct_when=lambda r: hasattr(g, "deduct_limit") and g.deduct_limit
+)
+def process_support_dialog():
```
nguyenkims, line 91: Currently in the project, both GET and POST are handled in the single method so I think it's better to combine `process_support_dialog` and `show_support_dialog`

app/dashboard/views/support.py
```diff
@@ -0,0 +1,110 @@
...
+    if not check_zendesk_response_status(response.status_code):
+        return False
+    LOG.debug("Ticket created")
+    return True
+
+
+@dashboard_bp.route("/support", methods=["POST"])
+@login_required
+@limiter.limit(
+    "2/hour", deduct_when=lambda r: hasattr(g, "deduct_limit") and g.deduct_limit
```
nguyenkims, line 89: The `g.deduct_limit` is used to only limit failed attempts, for example limiting the number of failed logins, for example https://github.com/simple-login/app/blob/f439e39580543791bcb5f4d171d586def84dc64d/app/api/views/auth.py#L150 

I think we can remove it here. When 2 methods `process_support_dialog` and `show_support_dialog` are combined, we can use `@limiter.limit("2/hour", methods=["POST"])` to only apply the limiter on POST requests.

templates/dashboard/support.html
```diff
@@ -0,0 +1,97 @@
...
+    </script>
+{% endblock %}
+
+{% block default_content %}
+
+    <div class="col pb-3">
+        <div class="card">
+            <div class="card-body">
+                <div class="card-title mb-3">Report a problem</div>
+                <div class="card-subtitle mt-2 mb-2 pl-3 border-left border-warning" style="border-left-width: 5px !important;" >
```
nguyenkims, line 53: Nits: I'd recommend to use `alert alert-info` class here instead of the `card-subtitle mt-2 mb-2 pl-3 border-left border-warning` and we can also remove the `style`

templates/dashboard/support.html
```diff
@@ -0,0 +1,97 @@
...
+                    If an email cannot be delivered to your mailbox, please check <a href="/dashboard/notifications">your notifications</a> for error messages
+                </div>
+                <div class="mt-2">
```
nguyenkims, line 56: I'd recommend to use `alert alert-warning` here instead of `mt-2`

templates/dashboard/support.html
```diff
@@ -0,0 +1,97 @@
...
+                    A support ticket will be created in Zendesk. Please do not include any sensitive information in the ticket.
+                </div>
+                <form id="supportZendeskForm" method="post" enctype="multipart/form-data">
+                    <div class="mt-4 mb-5">
+                        <label for="issueDescription" class="form-label">What happened?</label>
+                        <textarea class="form-control" name="ticket_content" id="issueDescription" rows="3" placeholder="Please provide as much information as possible. For example which alias(es), mailbox(es) ar affected, if this is a persistent issue...">{{- ticket_content or '' -}}</textarea>
```
nguyenkims, line 62: we can add `required` here so the validation is also done on the frontend.

templates/dashboard/support.html
```diff
@@ -0,0 +1,97 @@
...
+                        <label class="custom-file-label" for="ticketFileGroup">Choose file</label>
+                    </div>
+                    <div class="mt-5 font-weight-bold">
+                        Where can we reach you?
+                    </div>
+                    <div class="mt-2">
+                        Conversations related to this ticket will be sent to this address. Feel free to use an alias here.
+                    </div>
+                    <div class="input-group mb-3" id="alias-group">
+                        <input type="text" class="form-control" placeholder="Email" name="ticket_email" v-model='ticket_email' aria-label="Email to send responses to" aria-describedby="button-addon2">
```
nguyenkims, line 79: we can add `required` here to have the frontend validation.

nguyenkims (commented):
The PR is almost perfect now :), just added some comments on some improvements.

acasajus:
For some reason github doesn't allow me to reply to code comments. 

- Regarding merging the two methods in one. Each of them do a different thing. It makes the code more readable, easier to understand and with a low complexity. to keep them separate. Otherwise the decorators get more complex (some only apply to some method) and we need to handle each method independently anyway. 

- For the deduct_limit. You mean that now we only limit failed attempts of that we should only limit failed attempts? In the failed login case it makes sense that the failed ones are limited as once the user has been logged in that route ideally won't be called. In this case having a invalid request takes almost no time to be rejected, but a successful one will create a ticket in zendesk, and we'd like to avoid somebody flooding it.

nguyenkims:
> For some reason github doesn't allow me to reply to code comments.
> 
> * Regarding merging the two methods in one. Each of them do a different thing. It makes the code more readable, easier to understand and with a low complexity. to keep them separate. Otherwise the decorators get more complex (some only apply to some method) and we need to handle each method independently anyway.

Hey as other routes handle both GET and POST request inside a single method, I think it's more consistent to do the same pattern here. The current logic is not complex so I think merging them shouldn't be an issue.

> * For the deduct_limit. You mean that now we only limit failed attempts of that we should only limit failed attempts?  If it's the former it should be changed. If it's the latter, in the failed login case it makes sense that the failed ones are limited as once the user has been logged in that route ideally won't be called. In this case having a invalid request takes almost no time to be rejected, but a successful one will create a ticket in zendesk, and we'd like to avoid somebody flooding it.

What I suggested is to remove the `deduct_limit` as it doesn't make sense here and simply apply the rate limiting on POST request. We can also only set the `g.deduct_limit = True` if `create_zendesk_request` succeeds so that there's no rate limiting when there's no ticket created on Zendesk. Both options seem good to me.

acasajus:
Updated PR.

nguyenkims (approved):

PR #793, opened 2022-02-09
Add required package for authentication with AWS SES

jeifour:
One of the issues I was having when setting up SimpleLogin was this missing package in my Ubuntu 20 installation. Took me quite some time to figure out where the issue was, so I figured it'd be good to include it in the instructions.


nguyenkims (approved):
Thanks for fixing it!

PR #797, opened 2022-02-15
Only allow relative and controlled redirects

cquintana92:
Currently SL supports a "next" param in the login URL, the user will be redirected to that URL after logging in.

This "next" param doesn't do any check on the server, and it's possible a bad actor uses this URL to direct user to an evil page: `https://app.simplelogin.io/auth/login?next=https%3A%2F%2Fevilzone.org%2F`


app/utils.py
```diff
@@ -74,6 +75,14 @@ def sanitize_email(email_address: str, not_lower=False) -> str:
     return email_address
 
 
+def sanitize_next_url(url: Optional[str]) -> Optional[str]:
```
nguyenkims, line 78: I think we also want to handle the case next_url is an absolute url, for example https://app.simplelogin.io/something

app/utils.py
```diff
@@ -74,6 +75,14 @@ def sanitize_email(email_address: str, not_lower=False) -> str:
...
+    if url is None or len(url) == 0:
```
nguyenkims, line 79: Nits: we can use `if not url` here which would take into account these 2 cases.

nguyenkims (commented):
The PR looks good, just left some comments :)

app/utils.py
cquintana92: Changed it :+1:

app/utils.py
cquintana92: Added support for controlled absolute urls :+1:

app/config.py
```diff
@@ -418,3 +418,5 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 ZENDESK_HOST = os.environ.get("ZENDESK_HOST")
 ZENDESK_API_TOKEN = os.environ.get("ZENDESK_API_TOKEN")
 ZENDESK_ENABLED = "ZENDESK_ENABLED" in os.environ
+
+ALLOWED_REDIRECT_DOMAINS = sl_getenv("ALLOWED_REDIRECT_DOMAINS", list) or []
```
nguyenkims, line 422: I wonder if `ALLOWED_REDIRECT_DOMAINS` isn't set, can we default its value to the hostname used in `URL` variable? This way the existing SL instance can benefit from this protection.

app/utils.py
```diff
@@ -74,6 +75,39 @@ def sanitize_email(email_address: str, not_lower=False) -> str:
     return email_address
 
 
+class NextUrlSanitizer:
+    def __init__(self, allowed_domains: List[str]):
+        self.allowed_domains = allowed_domains
+
+    def sanitize(self, url: Optional[str]) -> Optional[str]:
```
nguyenkims, line 82: If we want to use class to organise code, I think we can use `@staticmethod`, this way we don't have to instantiate a new instance of `NextUrlSanitizer`

app/utils.py
```diff
@@ -74,6 +75,39 @@ def sanitize_email(email_address: str, not_lower=False) -> str:
...
+            return None
+        parsed = urllib.parse.urlparse(url)
+        if parsed.hostname in self.allowed_domains:
+            return url
+        # Not allowed domain
+        return None
+
+    def __is_absolute_url(self, url: str) -> bool:
+        return url.startswith(
+            ("http://", "https://", "http%3A%2F%2F", "https%3A%2F%2F")
```
nguyenkims, line 102: Do you know when `http%3A%2F%2F` case can happen? The `url` should be decoded before being passed to this method right?

nguyenkims (commented):
Looks good, just left some comments :).

app/utils.py
cquintana92: `NextUrlSanitizer` converted to a class with static methods :+1:

app/utils.py
cquintana92: You're right, and if it's not decoded at this stage it's probably due to some misconfiguration and the redirect shouldn't work anyways. Removed these cases

app/config.py
cquintana92: Done :+1:

nguyenkims (approved):
Looks perfect, thanks for making it!

PR #800, opened 2022-02-15
Show a warning if the user cannot create more aliases

acasajus:
Show a warning toast if the user cannot create more aliases. The last one is alredy kept in the form.


nguyenkims (approved):
I just tested it and it works well 👌.

PR #802, opened 2022-02-16
Run parsley validation on page reload.

FozzieHi:
Currently if you input some bad data into a form with parsley validation and refresh the page, the bad data is not outlined:
image

This PR rechecks parsley validation on page reload using the `PerformanceNavigationTiming.type` property. See [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type#browser_compatibility). Now on page reload, the user is shown current validation status:
image

### DIFF
--- a/templates/dashboard/custom_alias.html
+++ b/templates/dashboard/custom_alias.html
@@ -113,11 +113,17 @@
     $('.mailbox-select').multipleSelect();
 
     // Ctrl-enter submit the form
-    $('form').keydown(function (event) {
+    let form = $('form');
+    form.keydown(function (event) {
       if (event.ctrlKey && event.keyCode === 13) {
         $("#submit").click();
       }
     })
+    if ("getEntriesByType" in performance) {
+        if (performance.getEntriesByType("navigation").find(nav => nav.type === 'reload')) {
+            form.parsley().validate();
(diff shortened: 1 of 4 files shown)


FozzieHi:
Note: We might not even need the check for the `getEntriesByType` method in `performance`. However, I couldn't find the browsers we're targeting in the contributing document. It does seem [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType#browser_compatibility). @nguyenkims it'd be nice to have a defined list of compatible browsers we're targeting with the web app for implementing front end features.

nguyenkims:
@FozzieHi hey I'm not sure how to reproduce this bug. If we refresh a page, says the `/dashboard/custom_alias` one, all filled up data is reset right?

FozzieHi:
@nguyenkims If I go to custom alias, fill out the alias name then refresh the page then the alias name is still filled in. If I completely refresh the page (using Ctrl + F5) the data is reset. Maybe this is a specific browser feature? I'm using Firefox.

nguyenkims:
I run the test with Chrome and Safari, this seems indeed to be a Firefox specific behavior.

PR #803, opened 2022-02-17
Only allow authenticated and enabled users to accept a OAuth post request

acasajus:
Currenty we allow that route to be executed by non authenticated users.


acasajus:
OOps. Forgot to push changes last week. Pushed.

PR #811, opened 2022-02-21
Allow to configure ignore or reject response for blocked contacts

cquintana92:
This MR implements a feature that allows the users to decide which response they want to offer to a blocked contact.

There are two behaviours available:

* Ignore (return a 2XX status code to the sender, but don't deliver the email to the final user). aka: silently drop
* Reject (return a 5XX status code to the sender, and don't deliver the email to the final user). aka: actively reject the email

This MR implements the following changes:

1. New migration for adding a `blocked_behaviour` column for all the users. Defaults to `return_2xx`
2. New section in settings for allowing the users to change the setting.
3. Adapt the `email_handler.py` for honoring the user's selection.


nguyenkims (approved):
LGTM! The wording in the Settings page is really good and using "reject/ignore" is also easier to understand.

PR #813, opened 2022-02-24
Make nameservers configurable

cquintana92:
This MR implements allows the to define which nameservers to use when performing DNS requests.
By default the Cloudflare nameservers (`1.1.1.1`) will be used, but if the server administrator wants to, they can be changed.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -4,7 +4,7 @@ import socket
 import string
 import subprocess
 from ast import literal_eval
-from typing import Callable, List
+from typing import Callable, List, Optional
 from urllib.parse import urlparse
 
 from dotenv import load_dotenv
@@ -35,6 +35,13 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
         return default_factory()
 
     return literal_eval(value)
+
+
+def env(var: str, default: Optional[str] = None) -> Optional[str]:
+    if var in os.environ:
+        return os.environ[var]
+    else:
+        return default
 
 
 config_file = os.environ.get("CONFIG")
@@ -431,3 +438,11 @@ def get_allowed_redirect_domains() -> List[str]:
 
 
 ALLOWED_REDIRECT_DOMAINS = get_allowed_redirect_domains()
+
+
+def setup_nameservers():
+    nameservers = env("NAMESERVERS", "1.1.1.1")
+    return nameservers.split(",")
+
+
+NAMESERVERS = setup_nameservers()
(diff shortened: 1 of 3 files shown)


app/config.py
```diff
@@ -37,6 +37,13 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
     return literal_eval(value)
 
 
+def env(var: str, default: Optional[str] = None) -> Optional[str]:
```
nguyenkims, line 40: hey this function seems to be equivalent to `os.environ.get(var, default)` which is pretty short, should we use it instead?

app/dns_utils.py
```diff
@@ -1,20 +1,19 @@
...
-    my_resolver.nameservers = ["1.1.1.1"]
+    my_resolver.nameservers = config.NAMESERVERS
 
     return my_resolver
 
 
 def get_ns(hostname) -> [str]:
     try:
-        answers = _get_dns_resolver().resolve(hostname, "NS")
+        answers = _get_dns_resolver().resolve(hostname, "NS", search=True)
```
nguyenkims, line 16: do you know what this `search` parameter does?

nguyenkims (commented):

app/config.py
cquintana92: Oh, i hadn't seen that this usage was possible. Removed it.

app/dns_utils.py
cquintana92: For some reason, my branch didn't have this commit: https://github.com/simple-login/app/commit/936d90a5f51489263185f68bca57003802b1ced8 and when replacing the `query()` call with the `resolve()` call, I read the docs about `query()` calling `resolve` and setting `search=True`, so I wanted to retain the same behaviour as the old method call.
When rebasing it, the aforementioned commit ended up into my branch and the parameter got added too. I can remove it if you want.

cquintana92: For some reason, my branch didn't have this commit: https://github.com/simple-login/app/commit/936d90a5f51489263185f68bca57003802b1ced8 and when replacing the `query()` call with the `resolve()` call, I read the docs about `query()` calling `resolve` and setting `search=True`, so I wanted to retain the same behaviour as the old method call.
When rebasing it, the aforementioned commit ended up into my branch and the parameter got added too. I can remove it if you want.

nguyenkims: Oh that's indeed a good catch, it's better to keep the same behavior of `query` by setting `search=True`, thanks for taking a look!

nguyenkims (approved):
LGTM!

PR #815, opened 2022-02-24
Allow drag and drop of keys into the text area

acasajus:
The UI is not great but it's functional. Any suggestions on how to improve the UI?

### DIFF
--- a/templates/dashboard/contact_detail.html
+++ b/templates/dashboard/contact_detail.html
@@ -49,8 +49,10 @@
 
               <textarea name="pgp"
                   {% if not current_user.is_premium() %} disabled {% endif %}
-                        class="form-control" rows=10
+                        class="form-control" rows=10 id="pgp-public-key"
                         placeholder="-----BEGIN PGP PUBLIC KEY BLOCK-----">{{ contact.pgp_public_key or "" }}</textarea>
+              <div class="text-body p-2">You can drag and drop the pgp key into the text area</div>
+
             </div>
 
             <button class="btn btn-primary" name="action"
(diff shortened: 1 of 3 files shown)


templates/dashboard/contact_detail.html
```diff
@@ -70,4 +72,37 @@ <h1 class="h3">
   </div>
 {% endblock %}
 
+{% block script %}
+    {% if current_user.is_premium() %}
+    <script>
+    const MAX_BYTES = 2048; // 2KiB
```
nguyenkims, line 78: Some keys are quite big (mine is 3K), to be on the safe side I think we can use a bigger value here, for example 10K.

templates/dashboard/contact_detail.html
```diff
@@ -70,4 +72,37 @@ <h1 class="h3">
...
+
+    function drop(event) {
```
nguyenkims, line 80: As this drag and drop feature is used in both for mailbox and contact pages, I wonder if we can put it into a common file, for example we can put the code into `static/js/utils/drag_on_drop.js` and use a more generic class name (e.g. `.drag-drop-area` instead of `#pgp-public-key`), what do you think?

templates/dashboard/contact_detail.html
```diff
@@ -70,4 +72,37 @@ <h1 class="h3">
   </div>
 {% endblock %}
 
+{% block script %}
+    {% if current_user.is_premium() %}
```
nguyenkims, line 76: The whole PGP section is disabled if user isn't premium so I think we don't have to do this check here.

templates/dashboard/contact_detail.html
```diff
@@ -70,4 +72,37 @@ <h1 class="h3">
...
+            let reader = new FileReader();
+            reader.onloadend = onFileLoaded;
+            reader.readAsBinaryString(file);
+        }
+    }
 
+    function onFileLoaded(event) {
+        const initialData = event.currentTarget.result.substr(0, MAX_BYTES);
+        $('#pgp-public-key').val(initialData);
+        console.log($('#pgp-public-key'));
```
nguyenkims, line 101: the `console.log` can be removed I guess :)

templates/dashboard/contact_detail.html
```diff
@@ -49,8 +49,10 @@ <h1 class="h3">
 
               <textarea name="pgp"
                   {% if not current_user.is_premium() %} disabled {% endif %}
-                        class="form-control" rows=10
+                        class="form-control" rows=10 id="pgp-public-key"
                         placeholder="-----BEGIN PGP PUBLIC KEY BLOCK-----">{{ contact.pgp_public_key or "" }}</textarea>
+              <div class="text-body p-2">You can drag and drop the pgp key into the text area</div>
```
nguyenkims, line 54: For the styling, I'd suggest to use `alert alert-info` class as it's often used for hints. For example something like this would look ok:

image

nguyenkims (commented):
Just tested it and it works great! Just left some comments, mostly nits.

templates/dashboard/contact_detail.html
acasajus: I moved all the js into a file that contains a function to init the drag n' drop. The id of the text area can be passed as a param so we don't tie ourselves to having a magic name. Each page can have it's own and even if we have two of them in one page we can reuse the same with two ids.

templates/dashboard/contact_detail.html
acasajus: Removed

templates/dashboard/contact_detail.html
acasajus: Oops. Removed.

templates/dashboard/contact_detail.html
acasajus: :+1:

templates/dashboard/contact_detail.html
acasajus: Increased to 10K

nguyenkims (approved):
LGTM!

PR #835, opened 2022-03-10
New/admin audit trail

acasajus:
Created an AdminAuditLog to keep track of actions done through the admin panel. Each action will insert a record into the log that can't be modified.


app/models.py
```diff
@@ -2878,4 +2878,79 @@ class PhoneMessage(Base, ModelMixin):
...
+    ACTION_CREATE_OBJECT = 1
+    ACTION_UPDATE_OBJECT = 2
+    ACTION_DELETE_OBJECT = 3
+    ACTION_MANUAL_UPGRADE = 4
+    ACTION_EXTEND_TRIAL = 5
+    ACTION_DISABLE_2FA = 6
+    ACTION_LOGGED_AS_USER = 7
+
+    @classmethod
+    def create(cls, **kw):
```
nguyenkims, line 2903: This method can be removed if the AdminAuditLog inherits from ModelMixin I think

app/models.py
```diff
@@ -2878,4 +2878,79 @@ class PhoneMessage(Base, ModelMixin):
...
+        )
+
+    @classmethod
+    def extend_trial_1w(cls, admin_user_id: int, user_id: int, trial_end: arrow.Arrow):
+        cls.create(
+            admin_user_id=admin_user_id,
+            action=cls.ACTION_EXTEND_TRIAL,
+            model="User",
+            model_id=user_id,
+            data={"trial_end": trial_end.format(arrow.FORMAT_RFC3339)},
```
nguyenkims, line 2931: Just curious, why do we use `FORMAT_RFC3339` here?

app/admin_model.py
```diff
@@ -25,6 +34,42 @@ def inaccessible_callback(self, name, **kwargs):
         # redirect to login page if user doesn't have access
         return redirect(url_for("auth.login", next=request.url))
 
+    def on_model_change(self, form, model, is_created):
```
nguyenkims, line 37: 👌

nguyenkims (commented):
Looks good, I just left some comments :)

app/models.py
acasajus: Json itself doesn't have an official date format but ISO-8601/RFC3339 is the format used by js `Date.toJSON` method. So it's kind of the unofficial standard of dates in json.

app/models.py
```diff
@@ -2878,4 +2878,79 @@ class PhoneMessage(Base, ModelMixin):
     number = orm.relationship(PhoneNumber)
 
 
+class AdminAuditLog(Base):
```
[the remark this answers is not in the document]
acasajus: It doesn't have an `updated_at` that's why it doesn't inherit from the ModelMixin. An audit log that can be 'updated' is not much of an audit. We could add it and always have it be the same as a created_at but since this class doesn't need many of the methods in the ModelMixin we can pass without it. What do you think?

app/models.py
```diff
@@ -2878,4 +2878,79 @@ class PhoneMessage(Base, ModelMixin):
...
+    created_at = sa.Column(ArrowType, default=arrow.utcnow, nullable=False)
+    admin_user_id = sa.Column(sa.ForeignKey("users.id"), nullable=False)
+    action = sa.Column(sa.Integer, nullable=False)
+    model = sa.Column(sa.Text, nullable=False)
+    model_id = sa.Column(sa.Integer, nullable=True)
+    data = sa.Column(sa.JSON, nullable=True)
+
+    admin = orm.relationship(User, foreign_keys=[admin_user_id])
+
+    ACTION_CREATE_OBJECT = 1
```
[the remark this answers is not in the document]
acasajus: Ah, didn't know we had that. I'll change it :+1:

app/models.py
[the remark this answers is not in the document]
nguyenkims: Sounds good 👌

app/models.py
nguyenkims: I think by default `.format()` will use the ISO-8601 but it doesn't hurt to be explicit :)

app/models.py
acasajus: Solved in a previous comment

app/models.py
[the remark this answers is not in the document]
acasajus: Changed.

MrMuffinIsTaken:
> Created an AdminAuditLog to keep track of actions done through the admin panel. Each action will insert a record into the log that can't be modified.

Not bing able to modify it do may be to the disliking of some people.

nguyenkims:
> > Created an AdminAuditLog to keep track of actions done through the admin panel. Each action will insert a record into the log that can't be modified.
> 
> Not bing able to modify it do may be to the disliking of some people.

This PR is about the admin page which is only used by support team in SimpleLogin :)

nguyenkims (commented):
I got this error when running `alembic upgrade head`, I guess the DB schema has changed in the meantime. I'd suggest to base on the latest master and update the `9282e982bc05` or delete the migration file and generate a new one.

```
ERROR [alembic.util.messaging] Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
  FAILED: Multiple head revisions are present for given argument 'head'; please
  specify a specific target revision, '<branchname>@head' to narrow to a specific
  head, or 'heads' for all heads
```

MrMuffinIsTaken:
> > > Created an AdminAuditLog to keep track of actions done through the admin panel. Each action will insert a record into the log that can't be modified.
> > 
> > 
> > Not bing able to modify it do may be to the disliking of some people.
> 
> This PR is about the admin page which is only used by support team in SimpleLogin :)

I was referring to the self hosted version. Some people may not like the fact that they can't edited the logs. Granted i'm not trying to tell you how to make your application and run you business. I'm just stating a simple fact. PS, the admin page is also accessible for the self hosted version right or is it an in house only thing? Because if it isn't i'm gonna have to spin up my own simplelogin instance to see the administrative side of things to see whet kind of nice "gadgets" you build in to the web app ;) seen as administrative always as peeked an interest of mine.

nguyenkims (commented):
It seems that this action doesn't generate an admin log:

image

nguyenkims:
@MrMuffinIsTaken this admin page is also accessible in the self hosted version, you can access it on `/admin` path. You need to set the admin column in the `users` table to `true` in order to use this page. This page is only useful for us as it allows to manage subscriptions. In the future we might remove it from the repo.

MrMuffinIsTaken:
> @MrMuffinIsTaken this admin page is also accessible in the self hosted version, you can access it on `/admin` path. You

Intreating, but i mean, what if someone wants to self host simplelogin as a paid service them selfs? Wouldn't they find this useful? Tbh i think it not worth removing if it's useful to some people. But then again, It's not up to me.

nguyenkims (approved):
LGTM!

MrMuffinIsTaken:
> > > @MrMuffinIsTaken this admin page is also accessible in the self hosted version, you can access it on `/admin` path. You
> > 
> > 
> > Intreating, but i mean, what if someone wants to self host simplelogin as a paid service them selfs? Wouldn't they find this useful? Tbh i think it not worth removing if it's useful to some people. But then again, It's not up to me.
> 
> SL self hosted version is rather for personal use and not to be re-sold as otherwise this would be come a competitor to our service :).

Yeah that's fair, but then you may want to state that in your ToS or pick a license that covers it (seen as there will be some people whom will take advantage of this opportunity, ya know)

PR #843, opened 2022-03-18
Parse rspamd headers and apply dmarc policy if found. 

acasajus:
rspamd  policies softfail and quarantine will quarantine the mail. Reject will simply reject it.


app/email_utils.py
```diff
@@ -1370,6 +1370,11 @@ def sl_sendmail(
 
 def get_queue_id(msg: Message) -> Optional[str]:
     """Get the Postfix queue-id from a message"""
+    header_values = msg.get_all(headers.RSPAM_QUEUE_ID)
```
nguyenkims, line 1373: 👌

email_handler.py
```diff
@@ -536,6 +536,55 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
     )
 
 
+def apply_dmarc_policy(alias: Alias, contact: Contact, msg: Message) -> Optional[str]:
+    spam_result = msg.get_all(headers.SPAMD_RESULT)
```
nguyenkims, line 540: I think it'd be more clear to extract a method `get_dmarc_status` that can return either one of the symbols in https://rspamd.com/doc/modules/dmarc.html#dmarc-in-rspamd. The rspamd symbols can also be stored as an enum. We can then create unit tests for this method with different rspamd cases.

nguyenkims: I wonder if we can get the dmarc info for all messages and send the following dmarc metrics to newrelic (via `newrelic.agent.record_custom_metric`)? This will help us to know how many messages have issues with dmarc.

```
Custom/DMARC_BAD_POLICY
Custom/DMARC_NA
Custom/DMARC_POLICY_ALLOW
Custom/DMARC_POLICY_REJECT
Custom/DMARC_POLICY_QUARANTINE
Custom/DMARC_POLICY_SOFTFAIL
```

email_handler.py
```diff
@@ -536,6 +536,55 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
...
+    if not spam_result:
+        return False
+    spam_entries = [entry.strip() for entry in spam_result[-1].split("\n")]
+    for iPos in range(len(spam_entries)):
+        sep = spam_entries[iPos].find("(")
+        if sep > -1:
+            spam_entries[iPos] = spam_entries[iPos][:sep]
+    if "DMARC_POLICY_REJECT" in spam_entries:
+        return status.E519
```
nguyenkims, line 549: I think we shouldn't return 5** here as it can return a bounce message to the possibly falsified sender. We can use the same behavior for DMARC_POLICY_QUARANTINE here I think.

email_handler.py
```diff
@@ -536,6 +536,55 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
...
+    if (
+        "DMARC_POLICY_SOFTFAIL" in spam_entries
```
nguyenkims, line 551: At the beginning I'd leave DMARC_POLICY_SOFTFAIL out of the quarantine check as it if the sender set the dmarc policy to none, they don't want any special dmarc handling applied on their emails (and therefore accept the risk of having phishing with their domain)

email_handler.py
```diff
@@ -536,6 +536,55 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
...
+        or "DMARC_POLICY_QUARANTINE" in spam_entries
+    ):
+        add_or_replace_header(msg, headers.SL_DIRECTION, "Forward")
```
nguyenkims, line 554: I wonder if we can re-arrange the code so this dmarc checking happens just before sending the email to the mailbox (i.e. the "sl_sendmail" method)?

email_handler.py
```diff
@@ -536,6 +536,55 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
...
+            mailbox_id=alias.mailbox_id,
+            contact_id=contact.id,
+            alias_id=alias.id,
+            message_id=str(msg[headers.MESSAGE_ID]),
+            refused_email_id=refused_email.id,
+            is_spam=True,
+            blocked=True,
+            commit=True,
+        )
+        return status.E519
```
nguyenkims, line 584: Should we also inform user of this quarantine?

email_handler.py
nguyenkims: We can for example do the dmarc check in `forward_email_to_mailbox` and raise an exception to abort all the processing and catch the exception in `handle_forward`

tests/api/test_serializer.py
```diff
@@ -147,7 +147,7 @@ def test_get_alias_infos_with_pagination_v3_no_duplicate_when_empty_contact(
     """
     Make sure an alias is returned once when it has 2 contacts that have no email log activity
     """
-    user = create_user(flask_client)
+    user = create_random_user()
```
nguyenkims, line 150: ah do you know why this test fails with `create_user`?

tests/utils.py
```diff
@@ -27,17 +34,46 @@ def login(flask_client) -> User:
...
+
+
+def load_eml_file(
+    filename: str, template_values: Optional[Dict[str, str]] = None
+) -> EmailMessage:
+    emails_dir = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "example_emls"
+    )
+    fullpath = os.path.join(emails_dir, filename)
+    template = jinja2.Template(open(fullpath).read())
```
nguyenkims, line 75: should we use the `with open()` pattern here to close the file automatically?

tests/utils.py
```diff
@@ -27,17 +34,46 @@ def login(flask_client) -> User:
...
-        email="a@b.c",
+        email=email,
         password="password",
         name="Test User",
         activated=True,
         commit=True,
     )
 
 
+def create_random_alias(user: User) -> Alias:
```
nguyenkims, line 52: I think we have a similar method called `Alias.create_new_random`

tests/test_email_utils.py
```diff
@@ -748,6 +748,11 @@ def test_get_queue_id():
     assert get_queue_id(msg) == "4FxQmw1DXdz2vK2"
 
 
+def test_get_queue_id_from_double_header():
+    msg = load_eml_file("double_queue_id_header.eml")
```
nguyenkims, line 752: 👌

tests/test_email_handler.py
```diff
@@ -61,3 +65,23 @@ def test_is_automatic_out_of_office():
 
     msg[headers.AUTO_SUBMITTED] = "auto-generated"
     assert is_automatic_out_of_office(msg)
+
+
+def test_process_spoofed(flask_client):
```
nguyenkims, line 70: Hope we can have a way to test email handler without side effect soon. Calling `handle` can accidentally call sendmail which requires a SMTP server.

email_handler.py
```diff
@@ -616,6 +665,11 @@ def handle_forward(envelope, msg: Message, rcpt_to: str) -> List[Tuple[bool, str
         # do not return 5** to allow user to receive emails later when alias is enabled or contact is unblocked
         return [(True, res_status)]
 
+    # Check if we need to reject or quarantine based on dmarc
```
nguyenkims, line 668: I think we can create a new config to turn on/off this dmarc anti-phishing check. We might need to collect the metrics about dmarc for several days before turning on this check.

nguyenkims (commented):
Hey the PR looks really good,  I just left some comments.

email_handler.py
acasajus: When sending spoofed gmail emails that's the dmarc result that rspamd generates. For dmarc the authentication failed (either spf or dkim).

app/email/headers.py
```diff
@@ -12,6 +12,8 @@
 MIME_VERSION = "Mime-Version"
 REPLY_TO = "Reply-To"
 RECEIVED = "Received"
+RSPAM_QUEUE_ID = "X-Rspamd-Queue-Id"
```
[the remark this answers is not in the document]
acasajus: :+1:

email_handler.py
nguyenkims: Yeah unfortunately gmail has a too relaxed policy but this would protect against phishing attack for other popular domains (facebook.com, airbnb.com, etc all have DMARC enforced)

nguyenkims: In the future we can create a list of known domain that have dmarc "none" policy but that we want to enforce DMARC on.

tests/utils.py
acasajus: It doesnt' change much for read only files. the file will be closed when the handle is dropped. For write files is more important because we'd might need to ensure the file is saved before doing other stuff. 
I'll change it to use with.

tests/test_email_handler.py
acasajus: Yup. As soon as we have a way to test it independently.

email_handler.py
acasajus: Will do.

tests/utils.py
acasajus: Ok. I'll use it.

email_handler.py
acasajus: Ok. For now I've just set to quarantine and reject.

acasajus: Behind an enable flag.

email_handler.py
acasajus: Done

acasajus: Added as a custom_event so we can see each dmarc result independently.

email_handler.py
acasajus: Applied same behaviour to quarantine and reject

app/config.py
```diff
@@ -422,6 +422,8 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 ZENDESK_API_TOKEN = os.environ.get("ZENDESK_API_TOKEN")
 ZENDESK_ENABLED = "ZENDESK_ENABLED" in os.environ
 
+ENABLE_DMARC_CHECK = os.environ.get("ENABLE_DMARC_CHECK")
```
nguyenkims, line 425: for flag, usually we use something like `ENABLE_DMARC_CHECK = "ENABLE_DMARC_CHECK" in os.environ` I think

app/email_utils.py
```diff
@@ -1427,3 +1433,18 @@ def save_email_for_debugging(msg: Message, file_name_prefix=None) -> str:
...
 
     return ""
+
+
+def get_dmarc_status(msg: Message) -> Optional[DmarcCheckResult]:
+    spam_result = msg.get_all(headers.SPAMD_RESULT)
+    if not spam_result:
+        return None
+    spam_entries = [entry.strip() for entry in spam_result[-1].split("\n")]
+    for iPos in range(len(spam_entries)):
```
nguyenkims, line 1443: nits: should we use underscore instead of camel case for the variable name here?

app/models.py
```diff
@@ -237,6 +237,26 @@ class AuditLogActionEnum(EnumE):
     extend_subscription = 7
 
 
+class DmarcCheckResult(EnumE):
```
nguyenkims, line 240: 👌

email_handler.py
```diff
@@ -536,6 +539,51 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
     )
 
 
+def apply_dmarc_policy(alias: Alias, contact: Contact, msg: Message) -> Optional[str]:
+    dmarc_result = get_dmarc_status(msg)
+    newrelic.agent.record_custom_event(
```
nguyenkims, line 544: Oh didn't know about `newrelic.agent.record_custom_event`, do you know if it allows us to know how many emails have a specific DMARC status in a given period? As the value is a string here, I don't know how NewRelic would handle or whether it'd increase the storage too much (NewRelic pricing is based on storage)

email_handler.py
nguyenkims: Re in case you haven't seen this comment :)

nguyenkims (commented):
Looks good, just left some comments.

app/email_utils.py
acasajus: arghh.. it gets confusing changing between projects with camelCase and snake_case.. I'll change it :+1:

app/config.py
acasajus: Changed.

email_handler.py
acasajus: In that case if there are multiple mailboxes for an alias we'd be doing the same thing once per mailbox. If the mail fails the dmarc check and we do it before the `forward_email_to_mailbox` we only need to to the check once.

email_handler.py
nguyenkims: Oh I see, thanks for the info.

app/email_utils.py
```diff
@@ -1427,3 +1433,18 @@ def save_email_for_debugging(msg: Message, file_name_prefix=None) -> str:
         return file_name
 
     return ""
+
+
+def get_dmarc_status(msg: Message) -> Optional[DmarcCheckResult]:
```
nguyenkims, line 1438: I wonder if we can have some tests for this method to make sure its parsing is correct?

email_handler.py
nguyenkims: Yeah if there are multiple mailboxes, the exception will abort everything at the end forward_email_to_mailbox on the first mailbox. I just want to reuse the header manipulation in `forward_email_to_mailbox`.

nguyenkims (commented):
Just added some new comments.

email_handler.py
acasajus: Added a notification if there isn't one already there.

tests/api/test_serializer.py
acasajus: ah, if I run the tests several times in my local installation, since the user has already been created the index fails.

app/email_utils.py
acasajus: :+1:

acasajus: added

PR #845, opened 2022-03-21
require password to use the api key page

nguyenkims:


nguyenkims:
> It would be nice to have a `next` validator to use in all the places we need to use next param as in the sudo page. This way clients can't be redirected somewhere outside our domain. Maybe for a future PR?

yeah we can do that in a future PR

PR #862, opened 2022-03-29
Properly validate //host.com urls 

acasajus:
Take also into account urls that do not specify a scheme

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
-xaddopts =
+addopts =
     --cov
     --cov-config coverage.ini
     --cov-report=html:htmlcov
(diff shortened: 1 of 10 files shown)


pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts =
+xaddopts =
```
nguyenkims, line 2: Hey what is the difference between these 2 options?

nguyenkims (approved):
LGTM! I just left a question about pytest.

pytest.ini
acasajus: OOps. Shouldn't have committed that. Thanks for spotting it! I disable the pytest coverage locally because pycharm's debugger can't work if the pytest coverage is enabled.

PR #867, opened 2022-03-30
Enable pre-commit in the repo

acasajus:
Devs can do `poetry run pre-commit install` and flake8 and black will be run automatically before each commit.

The migrations dir has been formatted also according to both linters.


nguyenkims (commented):
The pre-commit is only called if a developer has installed git hook right? Do you know if we can ignore the migrations/ folder in black? These are auto generated files and we don't have to worry about their formats.

acasajus:
I think it's better to close this and have each dev setup their env as they see fit as long as commits are black indented and flake8 validated

PR #873, opened 2022-04-06
Save original envelope for debugging

acasajus:
Instead of saving the modified email save the original envelope so we can try to reproduce the problem easily


nguyenkims (approved):
LGTM!

PR #884, opened 2022-04-11
Send newrelic events on login and register

acasajus:


app/auth/views/register.py
```diff
@@ -60,6 +61,7 @@ def register():
                     hcaptcha_res,
                 )
                 flash("Wrong Captcha", "error")
+                RegisterEvent(RegisterEvent.ActionType.regsiter_catpcha_failed).send()
```
nguyenkims, line 64: There seems to be a typo in `regsiter_catpcha_failed`

nguyenkims (commented):
Hey I wonder if we should also log these events in api endpoints too?

app/auth/views/register.py
acasajus: Good catch :ok_hand:

acasajus:
> Hey I wonder if we should also log these events in api endpoints too?

Totally forgot about the api routes. I've just added the events there too :+1:

nguyenkims (approved):
LGTM!

PR #885, opened 2022-04-11
Do not show an error if we receive an unsubscribe from a different address

acasajus:
If a user has multiple addresses in a web client it's easy to send an unsubscribe message from the wrong address. Log it but don't show an error.


nguyenkims (approved):
That makes sense to decrease the logging level of this message 👌

PR #886, opened 2022-04-11
Do not assume all parts in multipart messages are processed as messages

acasajus:
Some parts of a multipart message can be kept as strings. If they are a string do not modify them.

Also added a missing formatting place.


nguyenkims (approved):
Thanks for handling this case, LGTM!

PR #891, opened 2022-04-12
Only send enum name for events intead of the full class.enum

acasajus:
So in new relic we'll see `action=login` instead of `action=RegisterEvent.Action.login`


nguyenkims (approved):
👌

PR #894, opened 2022-04-13
Add login with proton

cquintana92:
This PR adds the following features:

1. OAuth support for connecting to Proton
2. New variables for configuring the OAuth access
3. A new button for Login with Proton
4. A new settings section for linking / unlinking your Proton account


app/auth/views/proton.py
```diff
@@ -0,0 +1,119 @@
...
+        authorization_response=request.url,
+        verify=PROTON_VALIDATE_CERTS,
+        method="GET",
+        include_client_id=True,
+    )
+    credentials = convert_access_token(token["access_token"])
+    action = get_action_from_state()
+
+    client_original_ip = get_client_ip()
+    proton_client = HttpProtonClient(
```
nguyenkims, line 103: I wonder if we can use a Procedural approach here instead (i.e. not based on classes)? I feel the logic can be better followed and it is the current approach used in the repo. 

For example, we can have a method that handles the "login" action and another for the "link" action. All the logic can happen in these methods, eventually with some helper method. We can always isolate and test the helper method without using dependency injection. What do you think?

app/proton/proton_callback_handler.py
```diff
@@ -0,0 +1,230 @@
...
+
+PROTON_PARTNER_ID = 1
+
+
+class Action(enum.Enum):
+    Login = 1
+    Link = 2
+
+
+@dataclass
```
nguyenkims, line 20: 👌 for using dataclass

app/proton/proton_callback_handler.py
```diff
@@ -0,0 +1,230 @@
...
+                partner_id=PROTON_PARTNER_ID,
+            )
+            if other_partner_user is not None:
+                PartnerUser.delete(other_partner_user.id)
+
+            return link_user(proton_user, current_user)
+
+
+def process_case(
+    proton_user: ProtonUser, action: Action, current_user: Optional[User]
```
nguyenkims, line 207: Nits: should `current_user` be renamed to `sl_user` as above methods use this naming?

app/utils.py
```diff
@@ -101,6 +101,10 @@ def query2str(query):
     return query.statement.compile(compile_kwargs={"literal_binds": True})
 
 
+def parse_querystring(query: str) -> dict:
```
nguyenkims, line 104: Nits: this method is small enough so I think we can use `urllib.parse.parse_qs(query)` directly :)

nguyenkims (commented):
The PR looks good! 

I just left some comments, some are just nitpickings. I'd insist on using a procedural style for handling the callback as it's more consistent with the existing code and the logic is easier to check.

app/auth/views/proton.py
```diff
@@ -0,0 +1,119 @@
...
+_authorization_base_url = PROTON_BASE_URL + "/oauth/authorize"
+_token_url = PROTON_BASE_URL + "/oauth/token"
+
+# need to set explicitly redirect_uri instead of leaving the lib to pre-fill redirect_uri
+# when served behind nginx, the redirect_uri is localhost... and not the real url
+_redirect_uri = URL + "/auth/proton/callback"
+
+
+def get_client_ip() -> Optional[str]:
+    if request.environ.get("HTTP_X_FORWARDED_FOR") is None:
```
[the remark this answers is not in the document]
acasajus: Ah, we didn't realize `Werkzeug` takes care of this. In that case we should indeed use `request.remote_address`

app/auth/views/proton.py
```diff
@@ -0,0 +1,119 @@
...
+        return Action.Link
+    raise Exception(f"Unknown action in state: {oauth_action}")
+
+
+@auth_bp.route("/proton/login")
+def proton_login():
+    if PROTON_CLIENT_ID is None or PROTON_CLIENT_SECRET is None:
+        return redirect(url_for("auth.login"))
+
+    next_url = request.args.get("next")
```
acasajus, line 61: To prevent redirecting outside SL:

```suggestion
    next_url = sanitize_next_url(request.args.get("next"))
```

app/auth/views/proton.py
```diff
@@ -0,0 +1,119 @@
...
+        redirect_uri = _redirect_uri + "?next=" + encode_url(next_url)
+    else:
+        redirect_uri = _redirect_uri
+
+    proton = OAuth2Session(PROTON_CLIENT_ID, redirect_uri=redirect_uri)
+    authorization_url, state = proton.authorization_url(_authorization_base_url)
+
+    # State is used to prevent CSRF, keep this for later.
+    session["oauth_state"] = state
+    session["oauth_flow_state"] = json.dumps({"action": extract_action().value})
```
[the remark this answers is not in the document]
acasajus: we could directly do 
```suggestion
    session["oauth_action"] = extract_action().value
```

app/models.py
```diff
@@ -2983,4 +2986,32 @@ def extend_subscription(
...
+class PartnerApiToken(Base, ModelMixin):
+    __tablename__ = "partner_api_token"
+
+    token = sa.Column(sa.String(32), unique=True, nullable=False)
+    partner_id = sa.Column(sa.BigInteger, unique=False, nullable=False)
+    expiration_time = sa.Column(ArrowType, unique=False, nullable=True)
+
+
+class PartnerUser(Base, ModelMixin):
+    __tablename__ = "partner_user"
```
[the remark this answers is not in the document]
acasajus: Isn't social auth deprecated?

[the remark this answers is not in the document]
nguyenkims: Yes this class is deprecated.

app/auth/views/proton.py
acasajus: I would keep it as it is to start homogenizing it with the rest of the proton codebase. Going back and forth between sl and proton gets confusing. In the rest of the codebase we use interfaces a lot. Afaik this is the only way to implement interfaces in python and have the type checks pass.

app/auth/views/proton.py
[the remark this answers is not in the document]
cquintana92: Changed it for using `flask_limit.utils.get_remote_address`

migrations/versions/2022_041211_5269e8c2c2c9_.py
```diff
@@ -0,0 +1,78 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '5269e8c2c2c9'
+down_revision = 'b500363567e3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
```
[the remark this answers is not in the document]
cquintana92: Changed it to alembic syntax

templates/dashboard/setting.html
```diff
@@ -208,6 +208,33 @@
     </div>
     <!-- END Change email -->
 
+    <!-- Connect with Proton -->
+    {% if connect_with_proton %}
```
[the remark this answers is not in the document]
cquintana92: Fixed the indentation

templates/auth/login.html
```diff
@@ -13,36 +13,44 @@
...
 
-  <form class="card" style="border-radius: 2%" method="post">
-    {{ form.csrf_token }}
-    <div class="card-body p-6">
-      <h1 class="card-title">Welcome back!</h1>
-      <div class="form-group">
-        <label class="form-label">Email address</label>
-        {{ form.email(class="form-control", type="email", autofocus="true") }}
-        {{ render_field_errors(form.email) }}
+  <div class="card" style="border-radius: 2%">
```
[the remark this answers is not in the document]
cquintana92: Fixed the indentation

templates/dashboard/setting.html
```diff
@@ -208,6 +208,33 @@
...
+                <a
+                    class="btn btn-primary mt-2"
+                    style="border-color:#6d4aff;background-color:white;color:#6d4aff"
+                    href="{{ url_for('dashboard.unlink_proton_account') }}"
+                >Unlink account</a>
+            {% else %}
+                <div class="mb-3">
+                    You can connect your Proton account with your SimpleLogin one. <br>
+                </div>
+                <a class="btn btn-primary mt-2" style="border-color:#6d4aff;background-color:white;color:#6d4aff" href="{{ url_for("auth.proton_login", action="link") }}">Connect with Proton</a>
```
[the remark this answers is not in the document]
cquintana92: Created a CSS class for the Proton button

app/models.py
```diff
@@ -2983,4 +2986,32 @@ def extend_subscription(
...
+
+    user_id = sa.Column(sa.BigInteger, unique=False, nullable=False)
+    partner_id = sa.Column(sa.BigInteger, unique=False, nullable=False)
+    partner_email = sa.Column(sa.String(255), unique=False, nullable=True)
+
+
+class PartnerAuditAction(EnumE):
```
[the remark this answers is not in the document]
cquintana92: No, it's not used yet. I created it in case we added the PartnerAudit in this PR, but it can be created when the need comes. Removed it.

app/proton/proton_client.py
```diff
@@ -0,0 +1,169 @@
...
+    plan: ProtonPlan
+
+
+@dataclasses.dataclass
+class AccessCredentials:
+    access_token: str
+    session_id: str
+
+
+def convert_access_token(access_token_response: str) -> AccessCredentials:
```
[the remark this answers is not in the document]
cquintana92: Documented and tested

app/auth/views/proton.py
acasajus: It's not php or js pattern, having interfaces is a quite common way of having extensible logic in almost all languages. It's not used in the backend but in all the big codebases in proton. backend, web, ios, android... we have this as a common way of doing things.

Even if switching codebases is less frequent, its good that they share a way of doing things. This allows people to read and follow other codes consistently.

app/proton/proton_callback_handler.py
cquintana92: In this case, the `current_user` references to the "currently logged-in user".

When the code from this function branches depending on the action, it either remains as `current_user` (because it represents the currently logged-in user, or it disappears and then the retrieving process from the DB for obtaining the `sl_user` happens.

app/auth/views/proton.py
cquintana92: Changed it, nice catch

app/models.py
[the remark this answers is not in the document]
cquintana92: Also, shouldn't the constraint be added to the migration rather than the model?

app/utils.py
cquintana92: Removed it

app/auth/views/proton.py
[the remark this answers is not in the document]
cquintana92: Changed it, now we store the action directly

app/auth/views/proton.py
```diff
@@ -0,0 +1,108 @@
...
+    authorization_url, state = proton.authorization_url(_authorization_base_url)
+
+    # State is used to prevent CSRF, keep this for later.
+    session["oauth_state"] = state
+    session["oauth_action"] = extract_action().value
+    return redirect(authorization_url)
+
+
+@auth_bp.route("/proton/callback")
+def proton_callback():
```
nguyenkims, line 67: As I understand, this callback is used for both logging user and linking user right? In this case, maybe the code can be easier to read if we have something like:

```
if current_user:
     handle_link_user()
else:
     handle_login()
```

what do you think?

migrations/versions/2022_041211_5269e8c2c2c9_.py
```diff
@@ -0,0 +1,82 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '5269e8c2c2c9'
+down_revision = 'b500363567e3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    partner_table = op.create_table(
```
nguyenkims, line 22: Do you know if this migration file is auto generated? In general we should try to auto generate it as much as possible and manually editing it should the last resort.

migrations/versions/2022_041211_5269e8c2c2c9_.py
```diff
@@ -0,0 +1,82 @@
...
+
+        sa.ForeignKeyConstraint(['partner_id'], ['partner.id'], ondelete='cascade'),
+        sa.ForeignKeyConstraint(['user_id'], ['users.id'], ondelete='cascade'),
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    op.add_column('users', sa.Column('partner_id', sa.BigInteger(), nullable=True, server_default=None))
+    op.add_column('users', sa.Column('partner_user_id', sa.String(128), nullable=True, server_default=None))
+
+    op.execute(
```
nguyenkims, line 68: Self hosting user wouldn't need proton so we can remove it here I think.

partner_manager.py
```diff
@@ -0,0 +1,89 @@
...
+import sys
+
+from app import config
+from app.log import LOG
+from app.models import PartnerApiTokenModel, PartnerModel
+from app.utils import random_string
+from tabulate import tabulate
+
+
+def create_partner(args):
```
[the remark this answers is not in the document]
nguyenkims: cc @cquintana92 

In general I try to limit the number of files present in the root folder so a newcomer can more easily focus on the most critical entry points. For utility cli, I think we can group them into a file in `app/`

app/auth/views/proton.py
nguyenkims: I tried to read the code again and it's really hard to follow the logic as the call stack is deep. I think a code should be optimised for simplicity as it'd allow greater flexibility and maintainability in the future. 

The debate about object oriented vs procedural is actually useless. A procedural code can be also be as complex as object-oriented one if done wrong.

At some point, it's up to the engineering team to maintain the code base and I won't have the time to do code review so the decision of what pattern used is up to you.

nguyenkims (commented):
I just left some thoughts, some are just suggestions. Let's wrap this one quickly so we can deploy it on staging and run some more thorough tests.

partner_manager.py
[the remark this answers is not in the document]
cquintana92: We're not using it, so for now I've removed it until we decide how to refactor the CLI utilities :+1:

app/models.py
```diff
@@ -2992,4 +2995,27 @@ def extend_subscription(
...
+    __tablename__ = "partner"
+
+    name = sa.Column(sa.String(128), unique=False, nullable=False)
+    contact_email = sa.Column(sa.String(128), unique=True, nullable=False)
+
+
+class PartnerApiToken(Base, ModelMixin):
+    __tablename__ = "partner_api_token"
+
+    token = sa.Column(sa.String(32), unique=True, nullable=False)
```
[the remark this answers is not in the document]
cquintana92: It's a token that we will be generating on the backend, so we will control the length. 32^62 (a-zA-Z0-9) provides quite a good entropy :+1:

app/auth/views/proton.py
cquintana92: The initial logic for handling the request is the same in both cases, until we call the `ProtonCallbackHandler`.

The code inside the `ProtonCallbackHandler` already determines which case we're dealing with (and performs the `if` by checking also the `action` that we have stored in the state):
https://github.com/cquintana92/app/blob/feature/add-login-with-proton/app/proton/proton_callback_handler.py#L209

migrations/versions/2022_041211_5269e8c2c2c9_.py
cquintana92: We will need to make sure we mark somehow the Proton entry so the partner id matches:
https://github.com/cquintana92/app/blob/feature/add-login-with-proton/app/proton/proton_callback_handler.py#L12

Or provide a deterministic method for retrieving that entry in the `proton_callback_handler.py` file. Do we have an implementation like that anywhere in the code?

app/models.py
```diff
@@ -2992,4 +2995,27 @@ def extend_subscription(
         )
 
 
+class Partner(Base, ModelMixin):
+    __tablename__ = "partner"
+
+    name = sa.Column(sa.String(128), unique=False, nullable=False)
```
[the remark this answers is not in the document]
cquintana92: Changed

migrations/versions/2022_041211_5269e8c2c2c9_.py
cquintana92: I've added a check that only adds the Proton partner if the `CONNECT_WITH_PROTON` variable is set

app/auth/views/proton.py
nguyenkims: Yeah I saw that too, I just think once we have proton user data, it's maybe easier to read if we separate these 2 cases in the endpoint code directly (and not inside the ProtonCallbackHandler.handle -> process_case which adds additional call stack). This is totally optional.

migrations/versions/2022_041211_5269e8c2c2c9_.py
nguyenkims: No no, we should never change the database by hand :D. Everytime we need to make a change, we modify models.py and let alembic generate the necessary migration files. The database generation is only called once we are sure about the model changes so the dev flow is usually like:

1. make necessary change in models.py to support the new feature
2. run new-migration.sh to generate the migration file and run the migration via `alembic upgrade head`
3. continue working on other parts of the app. If we realized that the model change isn't right here, we need to revert the change in the database by `alembic downgrade -1` and starts from step 1 over again.

migrations/versions/2022_041211_5269e8c2c2c9_.py
nguyenkims: The migration code is quite special and I suggest not to modify it and leave alembic to generate it. The migration should only focus on the database structure and not add/remove data. 

I think we'll add the partner manually when we deploy the code (this operation is to be done once anyway) so we don't need to do that in the migration.

app/auth/views/proton.py
acasajus: I agree, the sooner we separate the logic flows the easier is to read.

migrations/versions/2022_041211_5269e8c2c2c9_.py
acasajus: If models is the source of truth we need to add a fixture to the tests since the tests don't run the migrations.

nguyenkims: In tests, the database is created via https://github.com/simple-login/app/blob/b9d26d46f6d5e2f4ba823d5ecbf93ccfbb17a96f/tests/conftest.py#L37
 which is also based on models.py

acasajus: Yes, I meant that alembic migrations aren't run in tests. DB is derived from models as you say :)

nguyenkims: The database migration model comes from Django which comes from RoR I believe. It offers a great productivity as we don't have to write sql migration scripts.

As a side note, I wrote a migration library for mysql that's based on SQL files and the migration is actually the source of truth https://pypi.org/project/migration-sql/. Of course no one uses it :D.

acasajus: We should make sure the tests run the migrations just to make sure we don't have anything broken. We could use the models to generate the migrations, but then run the tests with the migrations to make sure we won't break production.

nguyenkims: That's indeed a good idea, I created this PR for using the migration everywhere and remove the shortcut `Base.metadata.create_all(engine)`

acasajus: :+1:

templates/auth/login.html
```diff
@@ -13,36 +13,44 @@
...
 
+        <div class="form-footer">
+          <button type="submit" class="btn btn-primary btn-block">Log in</button>
+        </div>
 
+      </form>
+      {% if connect_with_proton %}
+        <div class="text-center my-2 text-gray"><span>or</span></div>
+        <a class="btn btn-primary btn-block mt-2" style="border-color:#6d4aff;background-color:white;color:#6d4aff" href="{{ url_for("auth.proton_login") }}">Log in with Proton</a>
+        <div class="text-center my-2 text-gray"><span>SimpleLogin and ProtonMail have joined forces to protect your inbox from spam and phising. <br /><a href="https://simplelogin.io/blog/simplelogin-join-proton/">Read more</a></span></div>
```
nguyenkims, line 49: As seen with Andy, we can remove this section "SimpleLogin and ProtonMail have joined forces ..." to not overcharge the login screen :)

templates/auth/login.html
```diff
@@ -13,36 +13,44 @@
...
+            </a>
+          </div>
+        </div>
 
+        <div class="form-footer">
+          <button type="submit" class="btn btn-primary btn-block">Log in</button>
+        </div>
 
+      </form>
+      {% if connect_with_proton %}
```
nguyenkims, line 46: can we also add "Sign up with Proton" on the sign up page which works similarly to this?

migrations/versions/2022_050411_5269e8c2c2c9_.py
```diff
@@ -0,0 +1,86 @@
+"""Create partner tables and add the partner_id column to users
```
nguyenkims, line 1: Do you know if this file is generated using alembic?

migrations/versions/2022_050411_5269e8c2c2c9_.py
```diff
@@ -0,0 +1,86 @@
...
+
+        sa.ForeignKeyConstraint(['partner_id'], ['partner.id'], ondelete='cascade'),
+        sa.ForeignKeyConstraint(['user_id'], ['users.id'], ondelete='cascade'),
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    op.add_column('users', sa.Column('partner_id', sa.BigInteger(), nullable=True, server_default=None))
+    op.add_column('users', sa.Column('partner_user_id', sa.String(128), nullable=True, server_default=None))
+
+    if CONNECT_WITH_PROTON:
```
nguyenkims, line 71: I think we can remove this as a migration should only contain with DDL (CREATE, DROP, ALTER) and not DML (SELECT, INSERT, UPDATE, DELETE). We can manually create a partner when we enable login with proton.

nguyenkims (commented):

templates/auth/login.html
cquintana92: Text removed

templates/auth/login.html
cquintana92: Button added. It calls to the exact same code, as from the oauth perspective the logic is the same in both the login and create account (if you try to log in with proton and you don't have a SimpleLogin account, one will be created for you).

migrations/versions/2022_050411_5269e8c2c2c9_.py
cquintana92: Okay, I'll then remove the hardcoded Proton partner id and run a query on runtime

cquintana92: Removed the hardcoded Proton partner id. We will need to manually create the partner on the database

app/proton/proton_callback_handler.py
```diff
@@ -0,0 +1,230 @@
...
+        PartnerUser.create(
+            user_id=sl_user.id,
+            partner_id=PROTON_PARTNER_ID,
+            partner_email=proton_user.email,
+        )
+        Session.commit()
+
+
+class ClientMergeStrategy:
+    __metaclass__ = ABCMeta
```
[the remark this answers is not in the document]
cquintana92: IIRC, Having it like this allows us to have an abstract constructor, but not invokable directly so we cannot build instances of the abstract base class

app/proton/proton_callback_handler.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @abstractmethod
```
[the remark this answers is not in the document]
cquintana92: Because they use that same constructor (all have the same), but this init (the one of the base class) is not invokable directly

app/proton/proton_callback_handler.py
```diff
@@ -0,0 +1,230 @@
...
+
+class UnexistantSlClientStrategy(ClientMergeStrategy):
+    def process(self) -> ProtonCallbackResult:
+        # Will create a new SL User with a random password
+        new_user = User.create(
+            email=self.proton_user.email,
+            name=self.proton_user.name,
+            partner_user_id=self.proton_user.id,
+            partner_id=PROTON_PARTNER_ID,
+            password=random_string(20),
```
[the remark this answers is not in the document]
cquintana92: That would leave the door open to some user trying to log in with an empty password having a successful login process, right?

migrations/versions/2022_050411_5269e8c2c2c9_.py
cquintana92: This is the migration that was written by hand (writing raw SQL) and then moved to alembic

app/proton/proton_callback_handler.py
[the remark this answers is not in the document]
nguyenkims: I'm not sure to understand ... What do you mean by "invokable directly" ? We still create instance of children classes right?

app/proton/proton_callback_handler.py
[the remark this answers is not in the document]
nguyenkims: That's covered in PasswordOracle.check_password() but that's indeed an interesting point.

templates/auth/register.html
```diff
@@ -48,6 +48,11 @@ <h1 class="card-title">Create new account</h1>
       <div class="mt-2">
         <button type="submit" class="btn btn-primary btn-block">Create Account</button>
       </div>
+
+      {% if connect_with_proton %}
+        <div class="text-center my-2 text-gray"><span>or</span></div>
+        <a class="btn btn-primary btn-block mt-2 proton-button" href="{{ url_for("auth.proton_login") }}">Sign up with Proton</a>
+      {% endif %}
```
nguyenkims, line 55: In login.html, it seems that we need to move this button out of the <form> but we don't need to do it here, is this normal?

acasajus: If a user doesn't have a SL account but does have a proton one, they may think that they need to create an account first. We could also replace the button for a small text with an explanation like 'If you have a proton account just log in'. Or do would you prefer to remove this entirely?

app/proton/proton_callback_handler.py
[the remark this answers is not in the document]
acasajus: This prevents directly creating an instance of `ClientMergeStrategy` but not of the children.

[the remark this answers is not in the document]
cquintana92: I've updated the code and made the check more explicit

templates/auth/register.html
nguyenkims: I think we don't need to mention this to not overcharge the UI and both buttons have the same logic behind anw.

app/auth/views/proton.py
```diff
@@ -0,0 +1,114 @@
...
+        PROTON_BASE_URL, credentials, get_remote_address(), verify=PROTON_VALIDATE_CERTS
+    )
+    handler = ProtonCallbackHandler(proton_client)
+
+    if action == Action.Login:
+        res = handler.handle_login()
+    elif action == Action.Link:
+        res = handler.handle_link(current_user)
+    else:
+        raise RuntimeError(f"Unknown Action: {action.name}")
```
nguyenkims, line 102: Do you know why we use RuntimeError here? Maybe a normal exception is enough?

cquintana92: Changed it for an `Exception`

app/proton/proton_callback_handler.py
```diff
@@ -0,0 +1,230 @@
...
+from dataclasses import dataclass
+from flask import url_for
+from typing import Optional
+
+from app.db import Session
+from app.models import User, PartnerUser
+from app.proton.proton_client import ProtonClient, ProtonUser
+from app.utils import random_string
+
+PROTON_PARTNER_ID = 1
```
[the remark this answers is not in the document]
cquintana92: I've removed the global `PROTON_PARTNER_ID` and now perform a cached query whenever we need it

app/auth/views/proton.py
cquintana92: Pushed the changes to separate the cases on the controller level

nguyenkims (approved):
Well done with this important PR!

app/models.py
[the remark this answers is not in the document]
cquintana92: Constraint added

PR #897, opened 2022-04-14
return whether a domain is custom or primary in GET /api/v5/alias/options

nguyenkims:
This endpoint modification is necessary for the iOS app so it can display the domain info when user creates a new alias.


tests/api/test_alias_options.py
```diff
@@ -124,6 +124,8 @@ def test_different_scenarios_v5(flask_client):
             suffix_payload["signed_suffix"],
         )
         assert signed_suffix.startswith(suffix)
+        assert "is_custom" in suffix_payload
+        assert "is_premium" in suffix_payload
```
acasajus, line 128: Minor: the test could also validate that the returned value is what we expect.

nguyenkims: as the domain order isn't defined and the response contains the `signed_suffix` which can change, it isn't easy to assert the exact value ... I think this test is rather a dummy test to make sure we don't forget to include the field.

PR #899, opened 2022-04-15
add alias to To: header if it isn't included in To and Cc header

nguyenkims:
cc @ntnhon this will add the alias to the To: header if needed

### DIFF
--- a/tests/auth/test_login.py
+++ b/tests/auth/test_login.py
@@ -20,16 +20,3 @@ def test_unactivated_user_login(flask_client):
         b"Please check your inbox for the activation email. You can also have this email re-sent"
         in r.data
     )
-
-
-def test_activated_user_login(flask_client):
-    user = create_new_user()
-
-    r = flask_client.post(
-        url_for("auth.login"),
-        data={"email": user.email, "password": "password"},
-        follow_redirects=True,
-    )
-
-    assert r.status_code == 200
-    assert b"/auth/logout" in r.data
--- a/tests/test_email_handler.py
+++ b/tests/test_email_handler.py
@@ -191,7 +191,7 @@ def test_dmarc_reply_quarantine(flask_client, dmarc_result):
 
 def test_add_alias_to_header_if_needed():
     msg = EmailMessage()
-    create_random_user()
+    create_new_user()
     alias = Alias.first()
 
     assert msg[headers.TO] is None
--- a/tests/test_email_handler.py
+++ b/tests/test_email_handler.py
@@ -187,3 +187,15 @@ def test_dmarc_reply_quarantine(flask_client, dmarc_result):
         user_id=user.id, alert_type=ALERT_DMARC_FAILED_REPLY_PHASE
     ).all()
     assert len(alerts) == 1
+
+
+def test_add_alias_to_header_if_needed():
+    msg = EmailMessage()
+    create_random_user()
+    alias = Alias.first()
+
+    assert msg[headers.TO] is None
+
+    email_handler.add_alias_to_header_if_needed(msg, alias)
+
+    assert msg[headers.TO] == alias.email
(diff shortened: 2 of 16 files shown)


acasajus:
In case a message with BCC is received, maybe it would be better to add the alias to the BCC field so make it clear the message is BCC'd and not sent directly to the alias? It's a bit strange to receive a message with a BCC header but I haven't seen anything that prevents it.

tests/api/utils.py
```diff
@@ -0,0 +1,13 @@
...
+from tests.utils import create_new_user
+
+
+def get_new_user_and_api_key() -> Tuple[User, ApiKey]:
+    user = create_new_user()
+
+    # create api_key
+    api_key = ApiKey.create(user.id, "for test", commit=True)
+
+    return user, api_key
```
acasajus, line 13: :100:

tests/auth/test_login.py
```diff
@@ -25,15 +23,11 @@ def test_unactivated_user_login(flask_client):
...
-    User.create(email="a@b.c", password="password", name="Test User", activated=True)
-    Session.commit()
+    user = create_new_user()
 
     r = flask_client.post(
         url_for("auth.login"),
-        data={"email": "a@b.c", "password": "password"},
+        data={"email": user.email, "password": "password"},
         follow_redirects=True,
     )
```
acasajus, line 32: `login(flask_client)` could be used directly here, couldn't it?

tests/test_email_handler.py
```diff
@@ -194,3 +187,15 @@ def test_dmarc_reply_quarantine(flask_client, dmarc_result):
...
+def test_add_alias_to_header_if_needed():
+    msg = EmailMessage()
+    create_new_user()
+    alias = Alias.first()
+
+    assert msg[headers.TO] is None
+
+    email_handler.add_alias_to_header_if_needed(msg, alias)
+
+    assert msg[headers.TO] == alias.email
```
acasajus, line 201: Nits: I'd add a test case for appending to the `To` header

```suggestion
    assert msg[headers.TO] == alias.email
    
    
def test_append_alias_to_header_if_needed():
    msg = EmailMessage()
    original_to = 'noone@nowhere.no'
    msg[headers.To] = original_to
    create_new_user()
    alias = Alias.first()

    email_handler.add_alias_to_header_if_needed(msg, alias)

    assert msg[headers.TO] == f"{original_to},{alias.email}"
```

nguyenkims:
> In case a message with BCC is received, maybe it would be better to add the alias to the BCC field so make it clear the message is BCC'd and not sent directly to the alias? It's a bit strange to receive a message with a BCC header but I haven't seen anything that prevents it.

It seems BCC is only used when sending emails and not receiving emails. I've run tests with some email clients, Apple Mail, Gmail both show BCC header but Proton doesn't 😅. To be on the safe side, I think we can stay with the To: header here, what do you think?

tests/test_email_handler.py
nguyenkims: Good idea! Added in https://github.com/simple-login/app/pull/899/commits/01fd8809021e80f2f5613411dc60edbb55a959c0

acasajus:
Ah, proton shows it for the original sender but not for the receiving end. But maybe it's something we can discuss with the mail team. That would make it way clearer.

tests/auth/test_login.py
nguyenkims: yeah that's true, this test is removed in https://github.com/simple-login/app/pull/899/commits/b545ebaeb117e95ed5f16b497c16cb82a8a35414

nguyenkims:
> Ah, proton shows it for the original sender but not for the receiving end. But maybe it's something we can discuss with the mail team. That would make it way clearer.

As displaying the BCC field isn't required in the protocol so some email clients won't do it, I think we can use the To field here to be on the safe side.

acasajus:
Ok. let's add it to the To.

PR #900, opened 2022-04-15
refactor test: no hardcode a@b.c, make sure each test has a different user

nguyenkims:


nguyenkims:
@acasajus @cquintana92 this PR is about refactoring tests and shouldn't have any side effect, I go ahead and merge it :)

PR #901, opened 2022-04-15
use _ instead of . in reverse alias

nguyenkims:
to avoid AC_FROM_MANY_DOTS SpamAssassin rule

### DIFF
--- a/tests/test_email_utils.py
+++ b/tests/test_email_utils.py
@@ -500,18 +500,18 @@ def test_generate_reply_email_include_sender_in_reverse_alias(flask_client):
     user.include_sender_in_reverse_alias = True
 
     reply_email = generate_reply_email("test@example.org", user)
-    assert reply_email.startswith("test.at.example.org")
+    assert reply_email.startswith("test_at_example_org")
     assert reply_email.endswith(EMAIL_DOMAIN)
 
     reply_email = generate_reply_email("", user)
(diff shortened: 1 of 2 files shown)


tests/test_email_utils.py
```diff
@@ -500,18 +500,18 @@ def test_generate_reply_email_include_sender_in_reverse_alias(flask_client):
...
     assert reply_email.endswith(EMAIL_DOMAIN)
 
     reply_email = generate_reply_email("👌汉字@example.org", user)
-    assert reply_email.startswith("yizi.at.example.org")
+    assert reply_email.startswith("yizi_at_example_org")
 
     # make sure reply_email only contain lowercase
     reply_email = generate_reply_email("TEST@example.org", user)
-    assert reply_email.startswith("test.at.example.org")
+    assert reply_email.startswith("test_at_example_org")
```
acasajus, line 516: Nits: I would add a test with 
```suggestion
    reply_email = generate_reply_email("test.dot@example.org", user)
    assert reply_email.startswith("test_dot_at_example_org")
    
```

nguyenkims: That's a good idea, added in the https://github.com/simple-login/app/pull/901/commits/e136fc8c926f3029c7c3ba47cb16cafa9457843b :)

xyxzxyz:
Any chance we can get dot back? Changing dot to underscore causes filters that use sender address completely usesless.

PR #905, opened 2022-04-19
Calculate proper padding when decoding base32

acasajus:
Added missing `% 8` to the padding calculator and ensured in the tests that we're testing all possible paddings.


nguyenkims (approved):
LGTM!

PR #906, opened 2022-04-19
Support python>3.8 for verp emails and reduce size by truncating hmac and storing time in minutes since 2022-01-01

acasajus:
- For python > 3.9 `shake128` has a new required parameter when generating the digest that breaks the `hmac` module. So we use `sha3-224` and truncate the resulting hmac to 8 bytes.
- Store time in minutes from 2022-01-01 to reduce further the length of the generated email.


.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.7]
+        python-version: [3.7, 3.9]
```
nguyenkims, line 11: 👌

nguyenkims:
@acasajus does this mean we change the algorithm to generate envelope mail_from? Can that affect how we parse the previously generated mail_from? In this case maybe we need to handle the previous format as well? (the same way we did for handling the old `bounce+{email_log_id}@sl.co`)

acasajus:
Has the old one been in production already?

nguyenkims:
@acasajus yep it's been deployed yesterday.

acasajus:
Argh.. ok. I'll add backwards compat

app/email_utils.py
```diff
@@ -1513,3 +1521,38 @@ def get_verp_info_from_email(email: str) -> Optional[Tuple[VerpType, int]]:
     if data[2] > time.time() + VERP_MESSAGE_LIFETIME:
         return None
     return VerpType(data[0]), data[1]
+
+
+# This method processes the email address, checks if it's a signed verp email generated by us to receive bounces
```
nguyenkims, line 1526: Nits: can we move these comments to the method doc?

nguyenkims (approved):
LGTM! Just left a suggestion. Is the algo change validated by crypto team?

acasajus:
> LGTM! Just left a suggestion. Is the algo change validated by crypto team?

yes :)

app/email_utils.py
acasajus: Done

PR #914, opened 2022-04-20
Store bounces in the reply phase to prevent abuse

acasajus:
- Refactored yahoo and hotmail complaint handling to have a single codebase
 - Store complaints in the reply phase to prevent abuse
 - Add admin page to review stored complaints


app/handler/transactional_complaint.py
```diff
@@ -0,0 +1,229 @@
...
+        return None
+
+    @classmethod
+    def name(cls):
+        return "yahoo"
+
+
+class TransactionalHotmailOrigin(TransactionalComplaintOrigin):
+    @classmethod
+    def get_original_message(cls, message: Message) -> Optional[Message]:
```
nguyenkims, line 65: same comment as the one above, I think it'd be helpful to keep the original comment on why we need the part 3.

app/handler/transactional_complaint.py
```diff
@@ -0,0 +1,229 @@
...
+    try:
+        _, to_address = parse_full_address(
+            get_header_unicode(original_message[headers.TO])
+        )
+        _, from_address = parse_full_address(
+            get_header_unicode(original_message[headers.FROM])
+        )
+    except ValueError:
+        saved_file = save_email_for_debugging(message, "FromParseFailed")
+        LOG.w("Cannot parse from header. Saved to {}".format(saved_file or "nowhere"))
```
nguyenkims, line 106: Nits: you can shorten the code by not using "format"

```
LOG.w("Cannot parse from header. Saved to %s", saved_file)
```

app/handler/transactional_complaint.py
```diff
@@ -0,0 +1,229 @@
...
+        return True
```
nguyenkims, line 107: when there's an error, maybe we can return False so the email will be forwarded to postmaster and we can then take a look at the email content?

nguyenkims (commented):
The PR looks good. 

My biggest feedback is on the "transactional" word that's used in several locations in the PR. It's a bit confusing as we use the same word for transactional emails sent by SL.

Maybe we can just remove this prefix?

tests/utils.py
```diff
@@ -11,17 +11,11 @@
...
-
 
 def create_new_user() -> User:
-    global _nb_user
-    _nb_user += 1
-
     # new user has a different email address
     user = User.create(
-        email=f"{_nb_user}@mailbox.test",
+        email=f"user{random.random()}@mailbox.test",
```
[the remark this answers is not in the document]
acasajus: If we want to parallelize tests it has less contention to just use a random number. From the performance point of view it's just a read from `urandom` that's quite fast compared to all the other stuff that we do on the tests. 
In any case it's good to keep the tests as isolated as possible to reduce the chance of one test interfering with another.

app/admin_model.py
```diff
@@ -366,3 +373,78 @@ class AdminAuditLogAdmin(SLModelView):
...
+        Session.commit()
+        return redirect("/admin/transactionalcomplaint/")
+
+    @expose("/download_eml", methods=["GET"])
+    def download_eml(self):
+        complaint = self._get_complaint()
+        if not complaint:
+            return redirect("/admin/transactionalcomplaint/")
+        eml_path = complaint.refused_email.full_report_path
+        eml_data = s3.download_email(eml_path)
```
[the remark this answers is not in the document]
acasajus: From the security point of view it'd be better to do it through the server. This way we ensure that only admins current admins can download mails. If we revoke access to someone, that person immediately loses access and can't access emails any longer.

app/handler/transactional_complaint.py
acasajus: `AutomatedComplaint`? `ProviderComplaint`? I would put some prefix before complaint to distinguish it from a user complaint or some other type of complaint

app/handler/transactional_complaint.py
acasajus: :+1:

app/handler/transactional_complaint.py
acasajus: I'm trying to consistently use format string `f"something with {var}"` but old habits die hard. I'll change it.

app/handler/transactional_complaint.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @classmethod
+    @abstractmethod
+    def name(cls):
+        pass
+
+
+class TransactionalYahooOrigin(TransactionalComplaintOrigin):
+    @classmethod
+    def get_original_message(cls, message: Message) -> Optional[Message]:
```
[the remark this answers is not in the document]
acasajus: I've added the comment back and set the counter incremental instead of decremental

app/models.py
```diff
@@ -2967,6 +2974,7 @@ def logged_as_user(cls, admin_user_id: int, user_id: int):
             action=AuditLogActionEnum.logged_as_user.value,
             model="User",
             model_id=user_id,
+            data={},
```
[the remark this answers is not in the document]
acasajus: There's a mismatch between the db and the model. Data is non-nullable and in this case it wasn't being added to the db. I've also set in the model definition data to be non-nullable.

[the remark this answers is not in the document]
nguyenkims: Oh I see.

tests/utils.py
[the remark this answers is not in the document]
nguyenkims: I see, let's use the random  then.

app/admin_model.py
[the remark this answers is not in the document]
nguyenkims: Makes sense. In the future we might want to separate admin and api/webapp deployment to avoid admin from affecting the main webapp performance.

app/handler/transactional_complaint.py
nguyenkims: `ProviderComplaint` sounds good.

app/admin_model.py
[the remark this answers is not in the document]
acasajus: :+1: it would also provide an extra layer of security

app/models.py
```diff
@@ -2907,7 +2914,7 @@ class AdminAuditLog(Base):
     action = sa.Column(sa.Integer, nullable=False)
     model = sa.Column(sa.Text, nullable=False)
     model_id = sa.Column(sa.Integer, nullable=True)
-    data = sa.Column(sa.JSON, nullable=True)
+    data = sa.Column(sa.JSON, nullable=False)
```
nguyenkims, line 2917: I don't see a migration script for making the data non-nullable, should we generate one?

nguyenkims: In this case we also handle the existing rows that have `data=null`

app/models.py
```diff
@@ -2988,5 +2996,32 @@ def extend_subscription(
...
+            model_id=complaint_id,
+            data={},
+        )
+
+
+class ProviderComplaintState(EnumE):
+    new = 0
+    reviewed = 1
 
-# endregion
```
nguyenkims, line 2992: should we move this `endregion` to the top? It corresponds to the Phone region I think.

app/handler/provider_complaint.py
```diff
@@ -0,0 +1,233 @@
...
+    DomainDeletedAlias,
+    Contact,
+    ProviderComplaint,
+    Phase,
+    ProviderComplaintState,
+    RefusedEmail,
+)
+
+
+class ProviderComplaintOrigin(ABC):
```
nguyenkims, line 36: Nits: I'm not sure to understand the "Origin" suffix here? Maybe we can just remove it?

app/models.py
```diff
@@ -2988,5 +2996,32 @@ def extend_subscription(
...
+
+class ProviderComplaint(Base, ModelMixin):
+    __tablename__ = "provider_complaint"
+
+    user_id = sa.Column(sa.ForeignKey("users.id"), nullable=False)
+    state = sa.Column(sa.Integer, nullable=False)
+    phase = sa.Column(sa.Integer, nullable=False)
```
nguyenkims, line 3020: I wonder if we can set phase as an enum, something like the following?

```
 phase = sa.Column(
        sa.Enum(Phase),
        nullable=False,
        server_default=Phase.unknown.value,
    )
```

app/config.py
```diff
@@ -336,10 +336,9 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 # alert when a new alias is about to be created on a disabled directory
 ALERT_DIRECTORY_DISABLED_ALIAS_CREATION = "alert_directory_disabled_alias_creation"
 
-ALERT_HOTMAIL_COMPLAINT = "alert_hotmail_complaint"
-ALERT_HOTMAIL_COMPLAINT_REPLY_PHASE = "alert_hotmail_complaint_reply_phase"
-ALERT_HOTMAIL_COMPLAINT_TRANSACTIONAL = "alert_hotmail_complaint_transactional"
-ALERT_YAHOO_COMPLAINT = "alert_yahoo_complaint"
+ALERT_COMPLAINT_REPLY_PHASE = "alert_complaint_reply_phase"
+ALERT_COMPLAINT_FORWARD_PHASE = "alert_complaint_forward_phase"
+ALERT_COMPLAINT_TO_USER = "alert_complaint_to_user"
```
nguyenkims, line 341: Nits: should this be named `ALERT_COMPLAINT_TRANSACTIONAL_PHASE`?

app/handler/provider_complaint.py
```diff
@@ -0,0 +1,233 @@
...
+            alias=alias,
+            destination=to_address,
+            provider=capitalized_name,
+        ),
+        max_nb_alert=1,
+        nb_day=7,
+    )
+
+
+def report_complaint_to_user(user: User, origin: ProviderComplaintOrigin):
```
nguyenkims, line 168: Nits: can we rename this method to `report_complaint_to_user_in_transactional_phase`?

nguyenkims (commented):
The PR looks good! I just left some suggestions.

app/models.py
acasajus: the model in the db is non-nullable. This is bugfix. It's the same issue as the first comment in models.py where I add a `data={}` to the `logged_as_user` function

app/models.py
acasajus: :+1:

app/models.py
acasajus: :+1:

app/handler/provider_complaint.py
acasajus: Removed

app/handler/provider_complaint.py
acasajus: Changed

app/config.py
acasajus: Changed

app/models.py
```diff
@@ -2991,5 +3002,34 @@ def extend_subscription(
...
+class ProviderComplaintState(EnumE):
+    new = 0
+    reviewed = 1
+
+
+class ProviderComplaint(Base, ModelMixin):
+    __tablename__ = "provider_complaint"
+
+    user_id = sa.Column(sa.ForeignKey("users.id"), nullable=False)
+    state = sa.Column(sa.Integer, nullable=False)
```
nguyenkims, line 3025: should we use `ProviderComplaintState` here for the state column?

nguyenkims (approved):
LGTM!

app/models.py
acasajus: We're using `ProviderComplaintState` for the state. Like in `provider_complaint.py:230`

nguyenkims: We can also enforce this in the data model, like in https://github.com/simple-login/app/blob/eab7606f93b1dad21522ad319e2a5c2eff2bdce5/app/models.py#L464

This isn't the case everywhere though so this stays totally optional.

acasajus: This is only setting the default to the enum value isn't it? Nothing prevents it from doing `state=3` somewhere in the code afaik.

acasajus: I've added the default.

PR #918, opened 2022-04-21
handle the AttributeError that can also be raised by as_bytes()

nguyenkims:


acasajus:
We could go for a more generic approach to try to get all possible undelying errors with something like:
```python
def to_bytes(msg: Message):
    """replace Message.as_bytes() method by trying different policies"""
    for generator_policy in [None, policy.SMTP, policy.SMTPUTF8]:
        try:
            return msg.as_bytes(policy=generator_policy)
        except Exception as e:
            LOG.w(
                "Error while doing message.as_bytes() with {} policy with error {}".format(
                    generator_policy or "default", e
                )
            )
    msg_string = msg.as_string()
    try:
        return msg_string.encode()
    except Exception as e:
        LOG.w("Error while doing message.as_string().encode() with error {}".format(e))
    return msg_string.encode(error="replace")
```

nguyenkims:
@acasajus it's usually not recommended to use the exception catch-all :) but maybe in this case we don't really have a choice. Hope we can write our own `as_bytes()` in the future.

acasajus:
I would  certainly prefer not to use the catchall but `msg.as_bytes()` is quite unreliable. In this case I'd go for a `do what you can` approach :)

nguyenkims:
@acasajus catch-all approach used in https://github.com/simple-login/app/pull/918/commits/68ec159d9114e1f6440512f0b5a944ff82cfafb8 )

acasajus:
:+1:

acasajus (approved):

PR #921, opened 2022-04-21
Prevent free users from creating reverse-alias

acasajus:


app/dashboard/views/alias_contact_manager.py
```diff
@@ -150,6 +151,65 @@ def get_contact_infos(
     return ret
 
 
+def create_contact(alias: Alias, contact_address: str) -> Optional[Contact]:
+    if not current_user.is_premium():
+        flash(f"Please upgrade to premium to create reverse-alias")
```
nguyenkims, line 156: I think `flash` requires a second parameter for the message level (info, warning, etc.)

app/dashboard/views/alias_contact_manager.py
```diff
@@ -150,6 +151,65 @@ def get_contact_infos(
...
+        return None
+
+    try:
+        contact_name, contact_email = parse_full_address(contact_address)
+    except Exception:
```
nguyenkims, line 161: Nits: we can just do `except:` here

nguyenkims, line 161: Nits: we can just do `except:` here

nguyenkims (commented):
Looks good to me. Should we also add the check for `POST /aliases/<int:alias_id>/contacts` endpoint?

app/dashboard/views/alias_contact_manager.py
acasajus: Oops. Good catch.

app/dashboard/views/alias_contact_manager.py
acasajus: Should be `ValueError` instead of `Except`. I've replaced it.

app/api/views/alias.py
```diff
@@ -402,6 +402,9 @@ def create_contact_route(alias_id):
         return jsonify(error="request body cannot be empty"), 400
 
     user = g.user
+    if not user.is_premium():
```
nguyenkims, line 405: As this endpoint can be called for existing alias, I think we should only return an error when it's a new contact.

tests/api/test_alias.py
```diff
@@ -489,6 +490,29 @@ def test_create_contact_route(flask_client):
     assert r.json["existed"]
 
 
+def test_create_contact_route_free_users(flask_client):
+    user = User.create(
```
nguyenkims, line 494: Maybe we can use `get_new_user_and_api_key` here?

nguyenkims (commented):
I think it'd be great to have a feature flag to turn on/off this reverse alias creation limit so we can choose when to turn on this feature when the error is handled on all clients.

app/api/views/alias.py
acasajus: It will be called to create a contact for existing aliases. Creating an alias is not restricted, just creating a reverse one.

nguyenkims: Ah I mean this endpoint can be called for existing **contact**

acasajus: Why is it called for an existing one? It should be called only to create new ones.

nguyenkims: Ideally it should be the case but as there are several clients out there, we can take a safer road by only rejecting the request of a new contact is about to be created, what do you think?

acasajus: Since we want to review what the clients will do on error, I'd prefer to fix the client behavior (where possible) to prevent strange code flows creeping in and having to maintain legacy behaviours in the server.

I've looked into the browser extension and android clients and none seem to be calling this route. And there are no calls to this route logged in newrelic in the last 30 days so I don't think iOS implements it either.

nguyenkims: This endpoint is used by ios and the browser extension to create new contact as far as i know. You can see its calls on https://onenr.io/0bRmDbovNwy

acasajus: Argh.. Do you know why I don't I see it by url? https://onenr.io/01wZK7qGbR6

acasajus: I don't know how I missed the browser creation :P I've checked it and it works as expected showing the error that currently just shows `Forbidden` we should put something like. Free users cannot create reverse alias, please upgrade.

nguyenkims: That's strange, I can see logs when I clicked on your link ... I checked and your Newrelic account has full access and is an admin account.

tests/api/test_alias.py
acasajus: changed

acasajus:
Still missing adding a flag for already existing users. Will do that next week.

acasajus:
Added setting a global flag to disable this feature until it's we want to enable it.

app/api/views/alias.py
```diff
@@ -407,50 +406,26 @@ def create_contact_route(alias_id):
...
-            reply_email=generate_reply_email(contact_email, user),
-        )
-    except CannotCreateContactForReverseAlias:
-        return jsonify(error="You can't create contact for a reverse alias"), 400
-
-    LOG.d("create reverse-alias for %s %s", contact_addr, alias)
-    Session.commit()
+        contact = create_contact(g.user, alias, contact_address)
+    except ErrContactErrorUpgradeNeeded as err:
+        return jsonify(error=err.error_for_user()), 403
```
nguyenkims, line 424: I think we need to update the api doc to include the status code 403 so clients can handle it.

app/dashboard/views/alias_contact_manager.py
```diff
@@ -16,9 +16,27 @@
...
+    ErrContactErrorUpgradeNeeded,
+    ErrAddressInvalid,
+    ErrContactAlreadyExists,
+)
 from app.log import LOG
-from app.models import Alias, Contact, EmailLog
+from app.models import Alias, Contact, EmailLog, User
+
+
+# TODO: Move this to a scoped config once the global config gets scoped.
```
nguyenkims, line 29: I wonder if we can change the config in the test like that: https://github.com/simple-login/app/blob/ca2b177e02a717d170afdf0920d9fbc9a8b96d87/tests/auth/test_register.py#L20 ?

app/dashboard/views/alias_contact_manager.py
```diff
@@ -44,6 +62,49 @@ def _check(form, field):
     return _check
 
 
+def create_contact(
+    user: User, alias: Alias, contact_address: str, fail_if_already_exist: bool = True
```
nguyenkims, line 66: Do you know if the `fail_if_already_exist` is used by caller?

app/dashboard/views/alias_contact_manager.py
```diff
@@ -44,6 +62,49 @@ def _check(form, field):
...
+) -> Contact:
+    if not contact_address:
+        raise ErrAddressInvalid("Empty address")
+    try:
+        contact_name, contact_email = parse_full_address(contact_address)
+    except ValueError:
+        raise ErrAddressInvalid(contact_address)
+
+    if not is_valid_email(contact_email):
```
nguyenkims, line 75: should we need to sanitize the contact address here? This is necessary as sometimes this address can contain artefacts due to copy/paste.

app/errors.py
```diff
@@ -60,3 +65,31 @@ class MailSentFromReverseAlias(SLException):
...
+
+    def error_for_user(self) -> str:
+        return f"{self.address} is not a valid email address"
+
+
+class ErrContactAlreadyExists(SLException):
+    """raised when a contact already exists"""
+
+    # TODO: type-hint this as a contact when models are almost dataclasses and don't import errors
+    def __init__(self, contact):
```
nguyenkims, line 91: You can use `contact: "Contact"` as type annotation. pycharm is quite intelligent and can understand it.

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts =
+xaddopts =
```
nguyenkims, line 2: should we revert this file?

scripts/new-migration.sh
```diff
@@ -2,9 +2,17 @@
 # To run it:
 # sh scripts/new-migration.sh
 
+container_name=sl-db-new-migration
+
+if [ "$#" -lt "1" ]; then
+	echo "What is this migration for?"
```
nguyenkims, line 8: Does this mean we require the developer to give a name for a migration? Usually I never modify a migration file and leave alembic handle everything.

templates/dashboard/alias_contact_manager.html
```diff
@@ -64,7 +64,11 @@ <h1 class="h3"> {{ alias.email }} contacts
         <div class="small-text">
           Where do you want to send the email?
         </div>
-        <button class="btn btn-primary mt-2">Create reverse-alias</button>
+          {% if alias.user.is_premium() %}
```
nguyenkims, line 67: Should we use the global flag here so free users can still create reverse alias in the meantime?

nguyenkims (commented):

app/dashboard/views/alias_contact_manager.py
acasajus: Not used. I'll remove it. Callers react to the exceptions triggered.

app/dashboard/views/alias_contact_manager.py
acasajus: Good idea!

app/errors.py
acasajus: :+1:

app/models.py
```diff
@@ -490,6 +493,14 @@ class User(Base, ModelMixin, UserMixin, PasswordOracle):
         ),
     )
 
+    # bitwise flags. Allow for future expansion
+    flags = sa.Column(
```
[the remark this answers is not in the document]
acasajus: This way is more scalable. We won't need to add more columns in the future to add more flags so it will be easier on the db :)

pytest.ini
acasajus: ARGHSD! I need to find a way to have it disabled locally without messing with this file. Reverted.

scripts/new-migration.sh
acasajus: I'll make it optional.

templates/dashboard/alias_contact_manager.html
acasajus: Good catch. I've added a var in the template with the same condition we use in the logic.

app/api/views/alias.py
acasajus: updated

app/models.py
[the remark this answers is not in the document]
nguyenkims: A boolean flag has below pros and cons:

- cons: 
   + require a database migration for every new flag. 
   + increase the number of column on user table.

- pros: 
   + storage (boolean is cheap)
   + more explicit
   + consistency with the existing flags

Either approach is fine for me, you can take the call here :).

templates/dashboard/alias_contact_manager.html
```diff
@@ -64,7 +64,11 @@ <h1 class="h3"> {{ alias.email }} contacts
         <div class="small-text">
           Where do you want to send the email?
         </div>
-        <button class="btn btn-primary mt-2">Create reverse-alias</button>
+          {% if can_create_contacts %}
```
nguyenkims, line 67: Nits: the indentation doesn't look right here.

nguyenkims (approved):
LGTM! Just left some nits.

PR #923, opened 2022-04-22
Allow SimpleLogin to be used as a dependency

cquintana92:
This PR adds an attribute to the `pyproject.toml` file that allows simplelogin to be included as a dependency from other Python projects.


nguyenkims (approved):
Lgtm!

PR #927, opened 2022-04-25
add new parameter disabled in /GET /api/v2/aliases

nguyenkims:
cc @ntnhon


acasajus (approved):
It would be nice to use an enum for the filter, but we can do that in another refactor PR

PR #931, opened 2022-04-26
Fix mobile view

nguyenkims:
Currently on mobile the header is overflow and the logo takes too much space

image

This PR uses adapts the layout for mobile: 1) use smaller logo and 2) move help button to the menu instead of in the header

image

image


acasajus:
I'm not a css expert by any means. I've tried it and it seems to work so LGTM!

acasajus (approved):

PR #932, opened 2022-04-26
Do not send the transports to the js part since we have not stored them previously

### LINKED ISSUE #929: Webauthn: Client authentication flow defaults to USB security key
## Prerequisites
- [x ] I have searched open and closed issues to make sure that the bug has not yet been reported.

## Bug report

**Describe the bug**
User has registered an Android phone as  roaming authenticator in the SimpleLogin settings menu:
register

**Expected behavior**
On login, the user should be prompted  to pick an authenticator, in this case the Android phone.
Instead, the user only sees this:
login
That means there's no way for the user to pick the "correct" authenticator because the browser prompt defaults to USB key.

**Environment (If applicable):**
 - OS: Linux
 - Browser: Chrome 100
 
**Additional context**
If I were to guess, there's probably some misconfiguration in SimpleLogin's [Webauthn Client Authentication](https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/WebAuthn_Client_Authentication.html) flow. 

edit: this is the only authenticator I have registered.
edit2: The phone acts as roaming authenticator. Communication between the authenticator (phone) and the client I try to log in with (linux chrome) happens via [caBLE protocol](https://blog.millerti.me/2021/06/18/previewing-chromes-cable-v2-support-for-webauthn/).

**Also, thanks for your great product :smile:**

acasajus:
Closes #929 

Cable is still not accepted in the webauthn spec but safari and crhome implement it. We can just remove the requirement since we don't know what's the proper transport for each credential because we're not saving it on creation.

### DIFF
--- a/app/auth/views/fido.py
+++ b/app/auth/views/fido.py
@@ -155,7 +155,8 @@ def fido():
     webauthn_assertion_options = webauthn_assertion_options.assertion_dict
     try:
         # HACK: We need to upgrade to webauthn > 1 so it can support specifying the transports
-        del webauthn_assertion_options["allowCredentials"][0]["transports"]
+        for credential in webauthn_assertion_options["allowCredentials"]:
+            del credential["transports"]
     except KeyError:
         # Should never happen but...
         pass
--- a/app/auth/views/fido.py
+++ b/app/auth/views/fido.py
@@ -153,6 +153,12 @@ def fido():
         webauthn_users, challenge
     )
     webauthn_assertion_options = webauthn_assertion_options.assertion_dict
+    try:
+        # HACK: We need to upgrade to webauthn > 1 so it can support specifying the transports
+        del webauthn_assertion_options["allowCredentials"][0]["transports"]
+    except KeyError:
+        # Should never happen but...
+        pass
 
     return render_template(
         "auth/fido.html",


app/auth/views/fido.py
```diff
@@ -153,6 +153,13 @@ def fido():
         webauthn_users, challenge
     )
     webauthn_assertion_options = webauthn_assertion_options.assertion_dict
+    try:
+        # HACK: We need to upgrade to webauthn > 1 so it can support specifying the transports
+        for credential in webauthn_assertion_options["allowCredentials"]:
+            del credential["transports"]
+    except KeyError:
+        # Should never happen but...
```
nguyenkims, line 161: Should we log a warning message here?

nguyenkims (commented):
I'm not expert on this webauthn part so just left a suggestion there :). Do you know how to test this PR? I guess it solves the issue when the webauthn device is an android phone connected to a computer via USB cable?

app/auth/views/fido.py
acasajus: In principle it should never happen as the library always adds it independently of whatever we pass it to it. If we fail to delete the only issue will be that we will limit the webauthn device to be the ones in the spec. So I think it's quite safe.

acasajus:
It can be tested locally using chome with a google credential logged in, and then with an (android?) phone with the google cred also logged in. Just add a webauthn device for the 2FA and select your phone from the dropdown. And then when logging in also select your phone as the 2fa method. It should work also with safari and apple credentials (and an iPhone?) but I can't test since I don't have an iPhone.

acasajus:
We should keep the defined transport when registering a webauthn credential and use those in the auth phase, but we can add that to a later PR.

nguyenkims (approved):
LGTM!

PR #936, opened 2022-04-27
prevent disabled user from using the api

nguyenkims:


acasajus:
Unless it's taken care somewhere I don't see, we should add to the User model

```python
  def is_active(self):
    return not self.disabled
```

so flask-login doesn't allow to login if disabled.

nguyenkims:
@acasajus we don't use flask-login for the api I think. For the web app, we have this to disable login for disabled account:

https://github.com/simple-login/app/blob/d0860cd54db0be2eb7063ae183f52cb4bb92e7a4/server.py#L199

acasajus:
Ah, the user_loader is the part that I was missing for the web. :+1:

PR #947, opened 2022-05-03
Align db with models

acasajus:
The AdminAuditLog was created directly by hand insteaf of from alembic autogeneration. Models and migrations aren't aligned right now and this should fix it.


nguyenkims (approved):
👍

PR #950, opened 2022-05-04
Remove deprecated verp email validation

acasajus:
Remove old verp email validation. It breaks with python 3.10.


nguyenkims (approved):
👌 for remembering this!

PR #951, opened 2022-05-04
Send complaints in the forward phase to the mailbox that owns the alias

acasajus:
The email_sender can now keep the non sent messages in mem so the test can validate we're sending what we expect

### DIFF
--- a/app/handler/provider_complaint.py
+++ b/app/handler/provider_complaint.py
@@ -151,7 +151,7 @@ def report_complaint_to_user_in_reply_phase(
     send_email_with_rate_control(
         alias.user,
         f"{ALERT_COMPLAINT_REPLY_PHASE}_{origin.name()}",
-        alias.user.email,
+        alias.mailbox.email,
         f"Abuse report from {capitalized_name}",
         render(
             "transactional/provider-complaint-reply-phase.txt.jinja2",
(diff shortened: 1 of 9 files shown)


app/handler/provider_complaint.py
```diff
@@ -197,16 +197,16 @@ def report_complaint_to_user_in_forward_phase(
     send_email_with_rate_control(
         user,
         f"{ALERT_COMPLAINT_FORWARD_PHASE}_{origin.name()}",
-        user.email,
+        alias.mailbox.email,
```
nguyenkims, line 200: an alias can have several mailboxes so I think we might need to find a way to extract the right mailbox.

app/handler/provider_complaint.py
```diff
@@ -151,7 +151,7 @@ def report_complaint_to_user_in_reply_phase(
     send_email_with_rate_control(
         alias.user,
         f"{ALERT_COMPLAINT_REPLY_PHASE}_{origin.name()}",
-        alias.user.email,
+        alias.mailbox.email,
```
nguyenkims, line 154: A reply can be sent either from one of the alias's mailboxes or an "authorized address".

nguyenkims (commented):
Looks good, I just left some comments.

app/handler/provider_complaint.py
acasajus: How can we find what mailbox did it? I don't think the bounce message has any more info besides the alias

app/handler/provider_complaint.py
acasajus: We don't know which authorized mbox sent it afaik.

nguyenkims: Just forwarded you an example of a complaint email. I noticed the `Return-Path` field that is the original envelope from and that should contains info about the email log that allows us to get the mailbox.

app/handler/provider_complaint.py
nguyenkims: just sent you an example of the complaint.

app/handler/provider_complaint.py
acasajus: Ah, thats a good one. In #963 I fixed it by getting the info from the report attached by yahoo that contains all the info or by extracting it from the `X-simplelogin-envelope-to`.

acasajus:
Superseeded by #963

PR #953, opened 2022-05-05
handle case user doesn't have default domain for alias

nguyenkims:
when user doesn't have default domain for alias, display "Not Selected" to avoid confusion

### DIFF
--- a/templates/dashboard/setting.html
+++ b/templates/dashboard/setting.html
@@ -275,20 +275,15 @@
           <input type="hidden" name="form-name" value="change-random-alias-default-domain">
           <select class="form-control mr-sm-2" name="random-alias-default-domain">
             <option value="">Not Selected</option>
-            {% if current_user.default_alias_custom_domain_id or current_user.default_alias_public_domain_id %}
-              {% for is_public, domain in current_user.available_domains_for_random_alias() %}
-                <option value="{{ domain }}"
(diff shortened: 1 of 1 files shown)


templates/dashboard/setting.html
```diff
@@ -274,12 +274,21 @@
...
+                  {{ domain }} ({% if is_public %} SimpleLogin domain {% else %} your domain {% endif %})
+                </option>
+              {% endfor %}
+            {% else %}
+              {% for is_public, domain in current_user.available_domains_for_random_alias() %}
+                <option value="{{ domain }}">
+                  {{ domain }} ({% if is_public %} SimpleLogin domain {% else %} your domain {% endif %})
+                </option>
+              {% endfor %}
+            {% endif %}
```
acasajus, line 291: Nits: Unless I'm missing something this way imo it's a bit more evident what's the difference.

```suggestion
              {% for is_public, domain in current_user.available_domains_for_random_alias() %}
                <option value="{{ domain }}"
                 {% if current_user.default_alias_custom_domain_id or current_user.default_alias_public_domain_id %}
                    {% if current_user.default_random_alias_domain() == domain %} selected {% endif %} >
                 {% endif %}
                  {{ domain }} ({% if is_public %} SimpleLogin domain {% else %} your domain {% endif %})
                </option>
              {% endfor %}
```

nguyenkims: Good idea 👌. Updated in the last commit.

PR #959, opened 2022-05-09
Preserve next_url in oauth login

cquintana92:
This PR allows the "Login with Proton" feature to preserve the `next` parameter so the user is redirected after a successful login.

### DIFF
--- a/app/auth/views/proton.py
+++ b/app/auth/views/proton.py
@@ -15,7 +15,7 @@ from app.config import (
 )
 from app.proton.proton_client import HttpProtonClient, convert_access_token
 from app.proton.proton_callback_handler import ProtonCallbackHandler, Action
-from app.utils import encode_url, sanitize_next_url
+from app.utils import sanitize_next_url
 
 _authorization_base_url = PROTON_BASE_URL + "/oauth/authorize"
 _token_url = PROTON_BASE_URL + "/oauth/token"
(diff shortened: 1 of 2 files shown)


app/auth/views/proton.py
```diff
@@ -50,12 +50,8 @@ def proton_login():
         return redirect(url_for("auth.login"))
 
     next_url = sanitize_next_url(request.args.get("next"))
-    if next_url:
-        redirect_uri = _redirect_uri + "?next=" + encode_url(next_url)
-    else:
-        redirect_uri = _redirect_uri
-
-    proton = OAuth2Session(PROTON_CLIENT_ID, redirect_uri=redirect_uri)
+    session["oauth_next"] = next_url
```
acasajus, line 53: What if there's no `next`?

cquintana92: Then a `None` will be stored

app/auth/views/proton.py
```diff
@@ -120,5 +116,5 @@ def check_status_code(response: requests.Response) -> requests.Response:
     if res.redirect:
         return redirect(res.redirect)
 
-    next_url = request.args.get("next") if request.args else None
+    next_url = session.get("oauth_next")
```
acasajus, line 119: If it's not defined it returns `None`?

cquintana92: Yes:

```
λ python3
Python 3.8.10 (default, Mar 15 2022, 12:22:08)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> d = {"a": 1}
>>> print(d.get("a"))
1
>>> print(d.get("b"))
None
```

app/auth/views/proton.py
acasajus: If the session object expects a `str` and `None` is stored then `"None"` will be stored won't it? and that's not a valid url

cquintana92: Stored only if `next_url` is not None in order to make it more explicit

nguyenkims (approved):
LGTM!

PR #963, opened 2022-05-10
Handle complaints that have multiple recipients

acasajus:
- If there are multiple recipients in a complaint we try to extract the one that triggered the complaint from the headers in the original report. 
- We also try to find out the original mailbox from the return-path of the original message
- Also added tests using real complaints from the providers


reset_local_db.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+export DB_URI=postgresql://myuser:mypassword@localhost:15432/simplelogin
+echo 'drop schema public cascade; create schema public;' | psql  $DB_URI
```
nguyenkims, line 4: I wonder if we can use `dropdb simplelogin & createdb simplelogin` here from inside the container here?

reset_test_db.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
nguyenkims, line 1: Can we add some information about what the script is for here? Maybe we can update the CONTRIBUTING.md to mention about these scripts?

nguyenkims (commented):
The PR looks good to me, just left some comments :)

app/handler/provider_complaint.py
```diff
@@ -32,25 +33,44 @@
...
     @classmethod
-    def sanitize_addresses(
+    def _get_mailbox_id(cls, return_path: Optional[str]) -> Optional[Mailbox]:
+        if not return_path:
+            return None
+        _, return_path = parse_full_address(get_header_unicode(return_path))
+        verp_type, email_log_id = get_verp_info_from_email(return_path)
+        if verp_type == VerpType.transactional:
+            return None
+        email_log = EmailLog.get_by(id=email_log_id)
```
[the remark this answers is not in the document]
acasajus: Changed

reset_local_db.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
acasajus: Moved :+1:

reset_local_db.sh
acasajus: My local db users cannot create dbs. This way they don't need that permission while still being able to reset the db. 
What container do you mean?

reset_test_db.sh
acasajus: Added explanation info to the `CONTRIBUTING.md`

tests/test.env
```diff
@@ -60,3 +60,5 @@ DMARC_CHECK_ENABLED=true
 PROTON_CLIENT_ID=to_fill
 PROTON_CLIENT_SECRET=to_fill
 PROTON_BASE_URL=https://localhost/api
+
+POSTMASTER=postmaster@simplelogin.co
```
nguyenkims, line 64: Nits: should we use a dummy value for this param?

app/email_utils.py
```diff
@@ -1395,7 +1409,9 @@ def generate_verp_email(
     ).lower()
 
 
-def get_verp_info_from_email(email: str) -> Optional[Tuple[VerpType, int]]:
+def get_verp_info_from_email(
+    email: str, validate_time: bool = True
```
nguyenkims, line 1413: Do we have a caller that uses the `validate_time` param?

app/handler/provider_complaint.py
```diff
@@ -30,15 +33,72 @@
...
+            _, sender_address = parse_full_address(
+                get_header_unicode(message[headers.FROM])
+            )
+
+            return OriginalMessageInformation(
+                sender_address,
+                rcpt_address,
+                cls._get_mailbox_id(message[headers.RETURN_PATH]),
+            )
+        except ValueError:
```
nguyenkims, line 97: Nits: should we move the except closer to where this error can happen? It should only happen in `parse_address_list` right?

app/handler/provider_complaint.py
```diff
@@ -76,6 +162,17 @@ def get_original_message(cls, message: Message) -> Optional[Message]:
...
 
+    @classmethod
+    def get_original_addresses(
+        cls, message: Message
+    ) -> Optional[OriginalMessageInformation]:
+        """
+        Try to get the proper recipient from original x-simplelogin-envelope-to header we add on delivery.
+        If we can't find the header, use the first address in the original message from"""
+        original = cls.get_original_message(message)
+        rcpt_header = original["x-simplelogin-envelope-to"]
```
nguyenkims, line 173: We can use `headers.SL_ENVELOPE_TO` here instead of `x-simplelogin-envelope-to`

nguyenkims (approved):
LGTM! Just left some suggestions.

tests/test.env
acasajus: Changed

app/email_utils.py
acasajus: removed

app/handler/provider_complaint.py
acasajus: Yes, but `cls._get_mailbox_id` also calls `parse_address_list`.

app/handler/provider_complaint.py
acasajus: changed

app/handler/provider_complaint.py
```diff
@@ -58,6 +118,32 @@ def get_original_message(cls, message: Message) -> Optional[Message]:
...
+    def get_original_addresses(
+        cls, message: Message
+    ) -> Optional[OriginalMessageInformation]:
+        """
+        Try to get the proper recipient from the report that yahoo adds as a port of the complaint. If we cannot find
+        the rcpt in the report or we can't find the report, use the first address in the original message from
+        """
+        report = cls.get_feedback_report(message)
+        original = cls.get_original_message(message)
+        rcpt_header = report["original-rcpt-to"]
```
nguyenkims, line 144: should we create a constant for "original-rcpt-to" in headers.py?

nguyenkims (approved):
LGTM! Just a suggestion.

app/handler/provider_complaint.py
acasajus: I was a bit on the fence with this one since it's a yahoo complaint specific header.. but you're right. better to reduce magic strings. I've set it as a constant in headers.py.

PR #967, opened 2022-05-11
Fix open redirect

cquintana92:
This PR fixes a potential open redirect vulnerability.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -80,10 +80,11 @@ class NextUrlSanitizer:
     def sanitize(url: Optional[str], allowed_domains: List[str]) -> Optional[str]:
         if not url:
             return None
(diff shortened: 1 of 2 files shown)


acasajus (approved):

app/utils.py
```diff
@@ -80,10 +80,11 @@ class NextUrlSanitizer:
     def sanitize(url: Optional[str], allowed_domains: List[str]) -> Optional[str]:
         if not url:
             return None
-        result = urllib.parse.urlparse(url)
+        replaced = url.replace("\\", "/")
```
nguyenkims, line 83: I wonder if there's a library in python that can validate an url so we only need to call `urllib.parse.urlparse(replaced)` if the url is valid.

PR #969, opened 2022-05-11
GH Actions refactor

acasajus:
- Create docker container also on tag
- Only run the docker container creation once 
- Run the workflow on push and on PR creation so it's run only once when the PR is updated with new commits

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ name: Run tests & Publish to Docker Registry
 on: [push, pull_request]
 
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
     strategy:
       max-parallel: 4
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -93,8 +96,13 @@ jobs:
           name: code-coverage-report
           path: htmlcov
 
+  build:
+    runs-on: ubuntu-latest
+    needs: ['test']
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))
```
nguyenkims, line 102: Do you know which docker image tag will be used here when the image is pushed?

nguyenkims (approved):
LGTM, just left a question.

.github/workflows/main.yml
acasajus: Good question. I assumed that `tag_with_ref` would take the updated ref and use it. But looking at the docs there is no info about it and a quick google search seems it's deprecated [here](https://github.com/docker/build-push-action/issues/116). I'll update the action to use the newer data.

acasajus: Fixed updating to the lastest docker actions.

PR #970, opened 2022-05-11
use a grace period of 14 days for paddle subscription

nguyenkims:


app/models.py
```diff
@@ -751,8 +751,9 @@ def get_subscription(self) -> Optional["Subscription"]:
         sub = Subscription.get_by(user_id=self.id)
 
         if sub:
-            # sub is active until the next billing_date + 1
-            if sub.next_bill_date >= arrow.now().shift(days=-1).date():
+            # grace period is 14 days
+            # sub is active until the next billing_date + 14
+            if sub.next_bill_date >= arrow.now().shift(days=-14).date():
```
acasajus, line 756: Nits: I'd suggest setting 14 as a `SUBSCRIPTION_GRACE_DAYS` like the `_APPLE_GRACE_PERIOD_DAYS`. And use it in the test :)

nguyenkims: Good idea 👍. Updated in the latest commit.

acasajus (approved):
lgtm!

PR #977, opened 2022-05-12
Obtain git information from version file

cquintana92:

* Fixes the `/git` endpoint not working.
* Replaces how the `config.SHA1` is processed. Now it reads a `.version` file in the root of the repo.
* Adds the generation of the `.version` file to the docker image build process.
* Adds the `.git/` directory to the `.dockerignore` so it doesn't end up inside the image, and thus making it lighter.


acasajus:
Instead of having a .dev file. we could do a `git_version.py` with python variables inside so we don't have to read from the fs on request.

PR #987, opened 2022-05-13
Allow users to keep the original unsubscribe action in the headers but mask mailto unsubscribe actions

acasajus:
- Users can keep the original unsubscribe action
 - If the original unsub action contains a mailto message, encode it and replace the header so we can send the unsub mail on behalf of the original alias if the user clicks the unsub
 - If the original unsub contains http request keep it
- Allow to preserver the previous unsub message format 
- The encoded unsub message allows to encode the current actions (but we stiil use the old unsub subjects)


app/handler/unsubscribe.py
```diff
@@ -0,0 +1,196 @@
...
+from app.config import URL
+from app.db import Session
+from app.email import headers, status
+from app.email_utils import send_email, render, add_or_replace_header
+from app.log import LOG
+from app.models import Alias, Contact, User, Mailbox
+
+
+class UnsubscribeGenerator:
+    def _generate_header_with_original_behaviour(self, message: Message) -> Message:
```
nguyenkims, line 14: Maybe we can use `@staticmethod` here as this method doesn't depend on `self`? Same for other methods in this file.

app/handler/unsubscribe.py
```diff
@@ -0,0 +1,196 @@
...
+        return message
+
+    def add_header_to_message(
+        self, alias: Alias, contact: Contact, message: Message
+    ) -> Message:
+        """
+        Add List-Unsubscribe header
+        """
+        user = alias.user
+        if user.setting:
```
nguyenkims, line 44: is `user.setting` a new column?

nguyenkims (commented):
I just left some comments.

app/handler/unsubscribe.py
acasajus: it's just a placeholder for the new setting. The setting itself is not yet there. Any ideas how to call it? I'm really bad at naming..

app/handler/unsubscribe.py
acasajus: Yes, it'll probably end up being static but right now it's still in early stages of dev/test :+1:

acasajus:
Super seeded by https://github.com/simple-login/app/pull/1148

PR #988, opened 2022-05-13
Fix: Sanitize directory name before displaying it to the user

acasajus:
Prevent users from displaying unsanitized HTML code in the delete confirmation box


nguyenkims (approved):
LGTM!

PR #993, opened 2022-05-15
update the email wording

nguyenkims:


acasajus:
If I send a mail that has a To: with a reverse-alias and a normal address. Can simplelogin reject the delivery? I mean won't the normal address recipient receive the email directly from my smtp server without going through SL?

nguyenkims:
> If I send a mail that has a To: with a reverse-alias and a normal address. Can simplelogin reject the delivery? I mean won't the normal address recipient receive the email directly from my smtp server without going through SL?

Yes SL will reject the email to not mix up the reply phase code.
And yes, the email is still sent to the non reverse alias address as the email won't pass by SL in this case ... This is what I tried to explain in the warning email.

templates/emails/transactional/non-reverse-alias-reply-phase.txt.jinja2
```diff
@@ -1,14 +1,18 @@
...
-Your email sent to {{ destination }} from {{ alias }} ({{ subject }}) contains an address which isn't a reverse-alias in the To: or Cc: field.
+  Your email sent to {{ destination }} from {{ alias }} ({{ subject }}) contains an address
+  which isn't a reverse-alias in the To: or Cc: field.
 
-This will reveal your real mailbox address.
+  The email was then directly sent to this recipient and doesn't pass by SimpleLogin.
+  Unfortunately, it reveals your real mailbox address to the recipient.
 
-Please create a reverse alias for each recipient to make sure your mailbox stays hidden.
+  SimpleLogin doesn't deliver this email to the contacts (if any) to avoid any potential side effect.
```
acasajus, line 12:
```suggestion
  SimpleLogin won't deliver this email to the reverse-alias (if any) to avoid showing any real address to any recipient.
```
?

acasajus: > > If I send a mail that has a To: with a reverse-alias and a normal address. Can simplelogin reject the delivery? I mean won't the normal address recipient receive the email directly from my smtp server without going through SL?
> 
> Yes SL will reject the email to not mix up the reply phase code. And yes, the email is still sent to the non reverse alias address as the email won't pass by SL in this case ... This is what I tried to explain in the warning email.

:ok_hand:

acasajus (approved):

PR #996, opened 2022-05-17
Added TOTP-like alias generation and included some extra information on CONTRIBUTING.md

zafuru:
I like having short aliases, but UUIDs are very long, and I prefer using random characters instead of two random words, so I constantly find myself creating custom aliases like:

```
abw7iy@domain.tld
```

This pull request adds a mode that allows creating random aliases just like that one, which is very similar to the code we might get when using one-time-passwords.

I also modified `CONTRIBUTING.md` and explained that to copy the `example.env`, we first have to go back to the root folder. I spent a minute trying to find out why I could not copy that file 😅

Please do let me know if there is anything else I need to do before this can be merged 😊

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,6 +1,7 @@
 import enum
 import os
 import random
+import string
 import uuid
 from email.utils import formataddr
 from typing import List, Tuple, Optional
@@ -216,6 +217,7 @@ class SenderFormatEnum(EnumE):
 class AliasGeneratorEnum(EnumE):
     word = 1  # aliases are generated based on random words
     uuid = 2  # aliases are generated based on uuid
+    totp = 3  # aliases are generated based on random characters
 
 
 class AliasSuffixEnum(EnumE):
@@ -1188,8 +1190,11 @@ def generate_email(
     if scheme == AliasGeneratorEnum.uuid.value:
         name = uuid.uuid4().hex if in_hex else uuid.uuid4().__str__()
         random_email = name + "@" + alias_domain
-    else:
+    elif scheme == AliasGeneratorEnum.word.value:
         random_email = random_words() + "@" + alias_domain
+    else:
+        name = "".join(random.choices(string.digits + string.ascii_lowercase, k=6))
+        random_email = name + "@" + alias_domain
 
     random_email = random_email.lower().strip()
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,9 +64,10 @@ Install npm packages
 cd static && npm install
 ```
 
-To run the code locally, please create a local setting file based on `example.env`:
+To run the code locally, go back to the root folder, and please create a local setting file based on `example.env`:
 
 ```
+cd ..
 cp example.env .env
 ```
 
--- a/app/api/views/new_random_alias.py
+++ b/app/api/views/new_random_alias.py
@@ -98,8 +98,10 @@ def new_random_alias():
                 scheme = AliasGeneratorEnum.word.value
             elif mode == "uuid":
                 scheme = AliasGeneratorEnum.uuid.value
+            elif mode == "totp":
+                scheme = AliasGeneratorEnum.totp.value
             else:
-                return jsonify(error=f"{mode} must be either word or uuid"), 400
+                return jsonify(error=f"{mode} must be either word, uuid, or totp"), 400
 
         alias = Alias.create_new_random(user=user, scheme=scheme, note=note)
         Session.commit()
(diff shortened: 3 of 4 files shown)


acasajus:
Hey, thanks for the PR! It seems to be a nice feature. I'll leave some comments:
 - The keyspace is too short. If we get to a million aliases we will have a 1 in 2k chance of collision. Since we can't use upper and lower case characters, at least make it 9 chars long.
  - Please don't call it totp. It stands for time-based one time password. It's neither time based nor a password. Can we call it `short` or something like that?
  - Can you add a test to make sure the alias is generated as expected?

ntnhon:
> * Please don't call it totp. It stands for time-based one time password. It's neither time based nor a password. Can we call it `short` or something like that?

I'd say `short_uuid` to give a clearer context

zafuru:
> Hey, thanks for the PR! It seems to be a nice feature. I'll leave some comments:
> 
>     * The keyspace is too short. If we get to a million aliases we will have a 1 in 2k chance of collision. Since we can't use upper and lower case characters, at least make it 9 chars long.
> 
>     * Please don't call it totp. It stands for time-based one time password. It's neither time based nor a password. Can we call it `short` or something like that?
> 
>     * Can you add a test to make sure the alias is generated as expected?
@acasajus You're right, I'll increase the keyspace, but can we make this customizable when using custom domains? Maybe not in this PR as it will require adding a button to the front-end, too.
I know it's not a TOTP, but I'm very bad at naming and that's the first thing that came to mind when I saw `word` and `uuid` 😅 and yeah, I'll add a test case.

> > * Please don't call it totp. It stands for time-based one time password. It's neither time based nor a password. Can we call it `short` or something like that?
> 
> I'd say `short_uuid` to give a clearer context

@ntnhon That does sound But that can also be misleading because it would not be a `uuid`, maybe `alphanumeric`? `slid`? How about `short_code`?

zafuru:
I'm glad you told me to write some tests because the previous code had some bugs, but now it should be better. I ended up calling this alias generator `random_string` like the suffix generator

app/api/views/setting.py
```diff
@@ -15,11 +15,14 @@
 
 
 def setting_to_dict(user: User):
+    alias_generator = {
+        "word": AliasGeneratorEnum.word.value,
+        "uuid": AliasGeneratorEnum.uuid.value,
+        "random_string": AliasGeneratorEnum.random_string.value,
+    }
```
acasajus, line 22:
```suggestion
    try:
        alias_generator = AliasGeneratorEnum[user.alias_generator].value
    except KeyError:
        alias_generator = AliasGeneratorEnum.word.value
```

app/api/views/setting.py
```diff
@@ -59,13 +62,15 @@ def update_setting():
...
+        if alias_generator not in ["word", "uuid", "random_string"]:
             return jsonify(error="Invalid alias_generator"), 400
 
         if alias_generator == "word":
             user.alias_generator = AliasGeneratorEnum.word.value
-        else:
+        elif alias_generator == "uuid":
             user.alias_generator = AliasGeneratorEnum.uuid.value
+        else:
+            user.alias_generator = AliasGeneratorEnum.random_string.value
```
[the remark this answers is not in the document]
acasajus: That's more future proof since if there are modifications in enum you don't need to update this code.

zafuru:
Funny, I was thinking that it'd be great if we had a dictionary in the enum class for that to avoid modifying the same code in the future, but I didn't know that we could already do that. Pretty cool. Will update the code soon 😊

zafuru:
@acasajus Done! I think. I also worked a bit on the base enum class

therealmrmuffin:
This would be lite. I would love to see this be added.

kelszo:
Looks good! 

However I think "random characters" is a better name than "random string", technically all current email generating methods are "random string" methods. Just a small input though.

Additionally I think random character alias length should be defined in the environment variables, just like how the path to the words file is. This to easily propagate eventual changes to the alias length across the whole code base.

Nautman:
Is there any update on when this could be implemented? ☺️

nguyenkims:
@Nautman it seems that there are still some questions that are not answered/addressed yet in the PR.

leon1995:
@zafuru anything new to this? I am really hoping that this could be merged soon! thanks for your work!

app/models.py
```diff
@@ -1188,7 +1192,10 @@ def generate_email(
     if scheme == AliasGeneratorEnum.uuid.value:
         name = uuid.uuid4().hex if in_hex else uuid.uuid4().__str__()
         random_email = name + "@" + alias_domain
-    else:
+    elif scheme == AliasGeneratorEnum.random_string.value:
+        name = "".join(random.choices(string.digits + string.ascii_lowercase, k=9))
```
developStorm, line 1196: Could reuse exisiting util function here
https://github.com/simple-login/app/blob/e4d43179880c14b34ed92ea533be2a9acc2ad061/app/utils.py#L41

developStorm:
I personally like this feature, hope this can be merged soon. For reference, DuckDuckGo Email Protection currently uses 8 random lowercase letters and numbers for their address generation and it seems to work well for them.

Hulk667i:
Agree. It will be good if implemented

PR #997, opened 2022-05-17
Adapt extension setup

cquintana92:

1. Adds a new `setup_done` page for the onboarding process that will trigger the "extension setup" flow on the new extension (PR for the extension [can be found here](https://github.com/simple-login/browser-extension/pull/134))
2. Changes the `next` parameter for the buttons on the first onboarding page.

### DIFF
--- a/app/onboarding/views/index.py
+++ b/app/onboarding/views/index.py
@@ -1,9 +1,10 @@
 from app.onboarding.base import onboarding_bp
 from flask import render_template
+from flask_login import current_user
 
 
 @onboarding_bp.route("/", methods=["GET"])
 def index():
     return render_template(
-        "onboarding/index.html",
+        "onboarding/index.html", is_user_logged_in=current_user is not None
     )
--- a/app/onboarding/views/index.py
+++ b/app/onboarding/views/index.py
@@ -4,6 +4,4 @@ from flask import render_template
 
 @onboarding_bp.route("/", methods=["GET"])
 def index():
-    return render_template(
-        "onboarding/index.html"
-    )
+    return render_template("onboarding/index.html")
--- a/app/onboarding/views/index.py
+++ b/app/onboarding/views/index.py
@@ -1,10 +1,9 @@
 from app.onboarding.base import onboarding_bp
 from flask import render_template
-from flask_login import current_user
 
 
 @onboarding_bp.route("/", methods=["GET"])
 def index():
     return render_template(
-        "onboarding/index.html", is_user_logged_in=current_user is not None
+        "onboarding/index.html"
     )
--- a/app/onboarding/views/setup_done.py
+++ b/app/onboarding/views/setup_done.py
@@ -1,0 +1,23 @@
+import arrow
+from flask import make_response, render_template
+from flask_login import login_required
+
+from app.config import URL
+from app.onboarding.base import onboarding_bp
+
+
+@onboarding_bp.route("/setup_done", methods=["GET", "POST"])
+@login_required
+def setup_done():
+    response = make_response(render_template("onboarding/setup_done.html"))
+
+    response.set_cookie(
+        "setup_done",
+        value="true",
+        expires=arrow.now().shift(days=30).datetime,
+        secure=True if URL.startswith("https") else False,
+        httponly=True,
+        samesite="Lax",
+    )
+
+    return response
--- a/app/onboarding/views/setup_done.py
+++ b/app/onboarding/views/setup_done.py
@@ -11,6 +11,7 @@ from app.onboarding.base import onboarding_bp
 def setup_done():
     response = make_response(render_template("onboarding/setup_done.html"))
 
+    # TODO: Remove when the extension is updated everywhere
     response.set_cookie(
         "setup_done",
         value="true",
--- a/templates/onboarding/index.html
+++ b/templates/onboarding/index.html
@@ -30,7 +30,14 @@
 
   {% if current_user != None and current_user.is_authenticated %}
     <script type="text/javascript">
-        setInterval(function() {
+        let counterIterations = 5;
+        let extensionSetupIntervalId = setInterval(function() {
+            counterIterations--;
+            if (counterIterations === 0) {
+                clearInterval(extensionSetupIntervalId);
+                return;
+            }
+
             const data = { tag: "PERFORM_EXTENSION_SETUP" };
             window.postMessage(data, "/");
         }, 300); // Send it many times, in case the extension had not registered the listener yet
--- a/templates/onboarding/index.html
+++ b/templates/onboarding/index.html
@@ -26,4 +26,11 @@
 
 </div>
 
+<script type="text/javascript">
+    setTimeout(function() {
+        const data = { tag: "PERFORM_EXTENSION_SETUP" };
+        window.postMessage(data, "/");
+    }, 300);
+</script>
+
 {% endblock %}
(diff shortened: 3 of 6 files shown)


acasajus (approved):
lgtm!

templates/onboarding/setup_done.html
```diff
@@ -0,0 +1,29 @@
...
+    <div class="mt-8 mb-4 text-center">
+      <h2 class="text-black-50" style="font-size:2rem">Performing the extension setup...</h2>
+    </div>
+  </div>
+
+  <script type="text/javascript">
+      setTimeout(function() {
+          const data = { tag: "PERFORM_EXTENSION_SETUP" };
+          window.postMessage(data, "/");
+      }, 300);
```
nguyenkims, line 26: Do you know why do we need a timeout here?

templates/onboarding/index.html
```diff
@@ -13,17 +13,39 @@
...
+
+  <div id="loading-section" class="mt-8 mb-4 text-center">
+    <h3>Loading...</h3>
   </div>
 
-  <!-- Button container -->
-  <div class="mt-8 text-center">
-    <a class="mx-6 p-4 text-decoration-none" style="background:black;color:white;" href="{{ url_for('auth.register', next=url_for('onboarding.final')) }}">Create a new account</a>
-    <a class="mx-6 p-4 text-decoration-none" style="background:white;color:black;border-radius: 2px;border:1px solid black;" href="{{ url_for('auth.login', next=url_for('onboarding.final')) }}">I already have an account</a>
+  <div id="content-section" style="display: none">
```
nguyenkims, line 21: I wonder if we can test whether the user is authenticated here (ie `current_user is not None`) and either display the login/register button or send the PERFORM_EXTENSION_SETUP message? This way we don't need to show/hide the login/register button in JS.

nguyenkims (commented):
The PR looks good, I just left some comments.

templates/onboarding/index.html
acasajus: if the user is authenticated and we send the message, what should be the contents of the page? It could be that a user ends up here without having the extension installed. Do we want to wait for a couple secs and then redirect them to the dashboard?

nguyenkims: > if the user is authenticated and we send the message, what should be the contents of the page

I think we can show the message "Setting up SimpleLogin browser extension ..." in this case. 

> Do we want to wait for a couple secs and then redirect them to the dashboard?

No I think we should redirect them to the final onboarding page. This case happens for user who is already logged in in SL and now install the extension. We want them to start using the extension.

templates/onboarding/setup_done.html
nguyenkims: Oh I see. In that case should we send several times (with delay between each call) the `PERFORM_EXTENSION_SETUP` message to be sure it'll arrive to the extension content script? Currently I see the setup is done for every PERFORM_EXTENSION_SETUP message, maybe we can modify the extension to only run the setup one:
https://github.com/simple-login/browser-extension/blob/13a4836b9d05b045334ddbfd1ed629c82f054e33/src/content_script/input_tools.js#L255

cquintana92: I'll try to implement that

app/onboarding/views/setup_done.py
```diff
@@ -0,0 +1,23 @@
...
+from app.config import URL
+from app.onboarding.base import onboarding_bp
+
+
+@onboarding_bp.route("/setup_done", methods=["GET", "POST"])
+@login_required
+def setup_done():
+    response = make_response(render_template("onboarding/setup_done.html"))
+
+    response.set_cookie(
```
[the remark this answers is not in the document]
cquintana92: TODO added

templates/onboarding/index.html
cquintana92: Changed it.

* If the user is logged in, we show the same message as the `setup_done.html` template and send the event.
* If the user is not logged in, show the buttons.

templates/onboarding/setup_done.html
cquintana92: Changed the  `setTimeout` for `setInterval` and updated the extension code so it only processes the event once.

templates/onboarding/index.html
```diff
@@ -18,12 +18,23 @@
     <h2 class="text-dark" style="font-size:2rem">Let's take back control of your inbox!</h2>
   </div>
 
-  <!-- Button container -->
   <div class="mt-8 text-center">
-    <a class="mx-6 p-4 text-decoration-none" style="background:black;color:white;" href="{{ url_for('auth.register', next=url_for('onboarding.final')) }}">Create a new account</a>
-    <a class="mx-6 p-4 text-decoration-none" style="background:white;color:black;border-radius: 2px;border:1px solid black;" href="{{ url_for('auth.login', next=url_for('onboarding.final')) }}">I already have an account</a>
+  {% if is_user_logged_in %}
```
nguyenkims, line 22: Optional Nits: jinja2 has access to `current_user` so we can use it directly here.

app/onboarding/views/index.py
```diff
@@ -1,9 +1,10 @@
 from app.onboarding.base import onboarding_bp
 from flask import render_template
+from flask_login import current_user
 
 
 @onboarding_bp.route("/", methods=["GET"])
 def index():
     return render_template(
-        "onboarding/index.html",
+        "onboarding/index.html", is_user_logged_in=current_user is not None
```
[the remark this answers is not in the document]
acasajus: Ah, just came here to write this. With this fix it works for me.

templates/onboarding/index.html
```diff
@@ -18,12 +18,23 @@
...
+    <a class="mx-6 p-4 text-decoration-none" style="background:black;color:white;" href="{{ url_for('auth.register', next=url_for('onboarding.setup_done')) }}">Create a new account</a>
+    <a class="mx-6 p-4 text-decoration-none" style="background:white;color:black;border-radius: 2px;border:1px solid black;" href="{{ url_for('auth.login', next=url_for('onboarding.setup_done')) }}">I already have an account</a>
+  {% endif %}
   </div>
-
 </div>
 
+  {% if is_user_logged_in %}
+    <script type="text/javascript">
+        setInterval(function() {
```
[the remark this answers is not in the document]
acasajus: If the extension is installed it will handle the first one, the rest will be discarded. But makes sense just to just avoid having an endless loop.

templates/onboarding/index.html
cquintana92: Changed it and removed the `is_user_logged_in` from the controller

templates/onboarding/index.html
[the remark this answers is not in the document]
cquintana92: Added the counter for limiting it

templates/onboarding/setup_done.html
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+
+    <!-- Text container -->
+    <div class="mt-8 mb-4 text-center">
+      <h2 class="text-black-50" style="font-size:2rem">Performing the extension setup...</h2>
+    </div>
+  </div>
+
+  <script type="text/javascript">
+      setInterval(function() {
```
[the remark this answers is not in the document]
cquintana92: Added the counter for limiting it

nguyenkims (approved):
LGTM!

app/onboarding/views/index.py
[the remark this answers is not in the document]
cquintana92: Should be fixed with the change for https://github.com/simple-login/app/pull/997#discussion_r877794101

PR #998, opened 2022-05-17
take into account status.E518

nguyenkims:


acasajus:
what's status E518?

nguyenkims:
> E518

It's used for when a mailbox is disabled. This is a normal condition and shouldn't be considered as an error, hence this PR

acasajus:
ah. I thought it was coming from postfix insteaf of the response status of the processing messages in the handler.

PR #1004, opened 2022-05-19
Add new page for account activated

cquintana92:
This PR adds a new page for the "account activated" view, that prompts the user for installing the extension


app/onboarding/views/account_activated.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Browser(Enum):
+    Firefox = 1
+    Chrome = 2
+    Other = 3
+
+
+def get_browser() -> Browser:
+    user_agent = request.user_agent
```
nguyenkims, line 17: This request.user_agent seems to be deprecated ... https://werkzeug.palletsprojects.com/en/2.0.x/utils/#module-werkzeug.useragents

app/onboarding/views/account_activated.py
```diff
@@ -0,0 +1,41 @@
+from app.onboarding.base import onboarding_bp
+from enum import Enum
+from flask import redirect, render_template, request, url_for
+
+
+CHROME_EXTENSION_LINK = "https://chrome.google.com/webstore/detail/simpleloginreceive-send-e/dphilobhebphkdjbpfohgikllaljmgbn"
+FIREFOX_EXTENSION_LINK = "https://addons.mozilla.org/firefox/addon/simplelogin/"
+
+
+class Browser(Enum):
```
nguyenkims, line 10: Can we also add Edge? The store link for Edge is https://microsoftedge.microsoft.com/addons/detail/simpleloginreceive-sen/diacfpipniklenphgljfkmhinphjlfff

app/onboarding/views/account_activated.py
```diff
@@ -0,0 +1,41 @@
...
+    user_agent = request.user_agent
+    if user_agent.browser in ["chrome", "edge", "opera", "webkit"]:
+        return Browser.Chrome
+    elif user_agent.browser in ["mozilla", "firefox"]:
+        return Browser.Firefox
+    return Browser.Other
+
+
+@onboarding_bp.route("/account_activated", methods=["GET"])
+def account_activated():
```
nguyenkims, line 26: should we mark this route for logged user only?

app/auth/views/activate.py
```diff
@@ -64,4 +64,4 @@ def activate():
         return redirect(next_url)
     else:
         LOG.d("redirect user to dashboard")
-        return redirect(url_for("dashboard.index"))
+        return redirect(url_for("onboarding.account_activated"))
```
nguyenkims, line 67: can we serve the account activated page directly without a redirection?

nguyenkims (commented):
I tested with different browsers and all of them work well 👍. Do you think we can also detect if the page is opened on a phone to not show the extension install button?

app/auth/views/activate.py
acasajus: It's easier to debug if they are separate. We don't need to do the full flow to test each browser.

app/onboarding/views/account_activated.py
acasajus: What a way to break things.. Instead of `werkzeug.useragents.UserAgent` its' not `werkzeug.user_agent.UserAgent`...

cquintana92: In the version of `werkzeug` that we are using I can still see the `werkzeug.useragents.UserAgent`
Should we upgrade the dependency then? Because we cannot use the new method if we don't do so

app/onboarding/views/account_activated.py
cquintana92: Added Edge to the browser detection: https://github.com/simple-login/app/pull/1004/commits/f5d0142aba0cde1a577096e8fe70b708f7cab0d6

app/onboarding/views/account_activated.py
cquintana92: If a user is able to reach this page without being logged in, nothing would happen.

If they clicked on the "Install extension" button, nothing would break. Also, if they clicked on the "Go to web app" they would be prompted to log in anyways.

acasajus: IT's a bit weird that an unauth user can see the activated page. Better to add the decorator and this way they will be sent to login.

app/onboarding/views/account_activated.py
acasajus: I don't think we can upgrade it without upgrading flask. I'll do a little test.

acasajus: Upgrading flask is no easy task. Some deps will break and there are packages that haven't been updated in years and don't support `werkzeug` 2. So I think it's safe for now to leave as it is.

app/onboarding/views/account_activated.py
```diff
@@ -0,0 +1,63 @@
...
+
+def get_browser() -> Browser:
+    user_agent = request.user_agent
+    if user_agent.browser == "edge":
+        return Browser.Edge
+    elif user_agent.browser in ["chrome", "edge", "opera", "webkit"]:
+        return Browser.Chrome
+    elif user_agent.browser in ["mozilla", "firefox"]:
+        return Browser.Firefox
+    return Browser.Other
```
acasajus, line 27:
```suggestion
    if user_agent.platform in ["android", "iphone", "ipad']:
        return Browser.Other
    if user_agent.browser == "edge":
        return Browser.Edge
    elif user_agent.browser in ["chrome", "opera", "webkit"]:
        return Browser.Chrome
    elif user_agent.browser in ["mozilla", "firefox"]:
        return Browser.Firefox
    return Browser.Other
```

app/onboarding/views/account_activated.py
cquintana92: Thanks for testing it

nguyenkims (approved):
LGTM!

PR #1019, opened 2022-05-23
Receive partner as param in ProtonCallbackHandler

cquintana92:

1. Make the ProtonCallbackHandler receive the `Partner` instance as a parameter, so it can be passed from the outside
2. Remove the references to the `get_proton_partner_id` in the `ProtonCallbackHandler`
3. Retrieve the Proton partner on the HTTP handler and pass it to the `ProtonCallbackHandler`

### DIFF
--- a/app/proton/proton_callback_handler.py
+++ b/app/proton/proton_callback_handler.py
@@ -25,10 +25,10 @@ def get_proton_partner() -> Partner:
 
 
 def get_proton_partner_id() -> int:
-    global _PROTON_PARTNER_ID
-    if _PROTON_PARTNER_ID is None:
-        partner = get_proton_partner()
-        _PROTON_PARTNER_ID = partner.id
+    global _PROTON_PARTNER
+    if _PROTON_PARTNER is None:
+        get_proton_partner()
+    return _PROTON_PARTNER.id
 
     return _PROTON_PARTNER_ID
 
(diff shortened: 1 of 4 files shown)


app/proton/proton_callback_handler.py
```diff
@@ -12,14 +12,22 @@
...
+    return _PROTON_PARTNER
+
+
+def get_proton_partner_id() -> int:
+    global _PROTON_PARTNER_ID
+    if _PROTON_PARTNER_ID is None:
+        partner = get_proton_partner()
         _PROTON_PARTNER_ID = partner.id
 
     return _PROTON_PARTNER_ID
```
acasajus, line 33:
```suggestion
def get_proton_partner_id() -> int:
    global _PROTON_PARTNER
    if _PROTON_PARTNER is None:
        get_proton_partner()
    return _PROTON_PARTNER.id

    return _PROTON_PARTNER_ID
```

acasajus: It's simpler and no need to keep two globals

app/proton/proton_callback_handler.py
```diff
@@ -12,15 +12,23 @@
...
+    return _PROTON_PARTNER
+
+
+def get_proton_partner_id() -> int:
+    global _PROTON_PARTNER
+    if _PROTON_PARTNER is None:
+        get_proton_partner()
+    return _PROTON_PARTNER.id
 
     return _PROTON_PARTNER_ID
```
acasajus, line 33:
```suggestion
```

acasajus (approved):
LGTM

PR #1022, opened 2022-05-24
Save unsent emails to disk to be resent later

acasajus:
- Added a config SAVE_UNSENT_DIR to specify where to keep the unsent messages
- Added a method to save the messages to disk if sending fails for some reason
- Added method to load them from disk

### DIFF
--- a/app/mail_sender.py
+++ b/app/mail_sender.py
@@ -32,12 +32,8 @@ class SendRequest:
 
     def to_bytes(self) -> bytes:
         if not config.SAVE_UNSENT_DIR:
+            LOG.d("Skipping saving unsent message because SAVE_UNSENT_DIR is not set")
             return
-        if not os.path.isdir(config.SAVE_UNSENT_DIR):
-            try:
-                os.makedirs(config.SAVE_UNSENT_DIR)
-            except FileExistsError:
-                pass
         serialized_message = message_to_bytes(self.msg)
         data = {
             "envelope_from": self.envelope_from,
--- a/app/mail_sender.py
+++ b/app/mail_sender.py
@@ -6,6 +6,7 @@ import os
 import time
 import uuid
 from concurrent.futures import ThreadPoolExecutor
+from functools import wraps
 from mailbox import Message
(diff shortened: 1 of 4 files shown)


app/mail_sender.py
```diff
@@ -23,22 +30,72 @@ class SendRequest:
     is_forward: bool = False
     ignore_smtp_errors: bool = False
 
+    def to_bytes(self) -> bytes:
+        if not config.SAVE_UNSENT_DIR:
+            return
+        if not os.path.isdir(config.SAVE_UNSENT_DIR):
```
nguyenkims, line 36: I wonder if  we can do this directory creation in `config` instead? Similarly to https://github.com/simple-login/app/blob/15810c2382cc81e6312e7a8484cc47af1a278a78/app/config.py#L226

nguyenkims (commented):
Should we have a cronjob to resend emails that can't be sent?

acasajus:
Yeah, we should add a cronjob. But Id' prefer to do it in another PR to keep this one smallish.

acasajus:
Why the heck the big buttton is `close with comment`?

app/mail_sender.py
acasajus: Done

app/mail_sender.py
```diff
@@ -98,20 +151,30 @@ def _send_to_smtp(self, send_request: SendRequest, retries: int):
...
                 self._send_to_smtp(send_request, retries - 1)
             else:
-                if send_request.ignore_smtp_error:
-                    LOG.w("Ignore smtp error %s", e)
-                else:
-                    raise
+                if send_request.ignore_smtp_errors:
+                    LOG.w(f"Ignore smtp error {e}")
+                    return
+                LOG.e(
```
nguyenkims, line 166: I wonder if we can use LOG.exception here as it'll include all the stacktrace?

nguyenkims (approved):
LGTM! Just left a suggestion :)

app/mail_sender.py
acasajus: Changed :+1:

PR #1028, opened 2022-05-26
Store hmaced partner api tokens

cquintana92:

1. Add a migration for increasing the size of partner api tokens
2. Add a `PartnerApiToken.generate` method that takes care of generating api tokens, processing them and storing them into the database
3. Add a `Partner.find_by_token` method that allows retrieving a `Partner` instance given a raw `PartnerApiToken`.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,4 +1,9 @@
+from __future__ import annotations
+
+import base64
 import enum
+import hashlib
+import hmac
 import os
 import random
 import uuid
@@ -18,6 +23,7 @@ from sqlalchemy import text, desc, CheckConstraint, Index, Column
 from sqlalchemy.dialects.postgresql import TSVECTOR
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import deferred
+from sqlalchemy.sql import and_
 from sqlalchemy_utils import ArrowType
 
 from app import s3
@@ -37,6 +43,7 @@ from app.config import (
     MAX_NB_DIRECTORY,
     ROOT_DIR,
     NOREPLY,
+    PARTNER_API_TOKEN_SECRET,
 )
 from app.db import Session
 from app.errors import (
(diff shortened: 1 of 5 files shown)


app/models.py
```diff
@@ -3075,16 +3082,55 @@ class Partner(Base, ModelMixin):
...
+            Session.query(Partner, PartnerApiToken)
+            .filter(
+                and_(
+                    PartnerApiToken.token == hmaced,
+                    Partner.id == PartnerApiToken.partner_id,
+                )
+            )
+            .first()
+        )
+        if res is None or res.Partner is None:
```
nguyenkims, line 3098: Oh I didn't know we can use `res.Partner` here. Usually the result is unpacked with `partner, api_token=res`

nguyenkims (approved):
LGTM!

tests/models/test_partner_api_token.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_generate_partner_api_token(flask_client):
+    partner = Partner.create(
+        name=random_string(10),
+        contact_email="{s}@{s}.com".format(s=random_string(10)),
+        commit=True,
+    )
+
+    (partner_api_token, token) = PartnerApiToken.generate(partner.id, None)
```
nguyenkims, line 12: Nits: we can remove parenthesis `(` here :)

cquintana92: Removed

app/models.py
cquintana92: Changed it

cquintana92: Sorry, re-reading the comment I have noticed I misunderstood. Fixed the `None` crash and replaced it with the destructuring syntax

acasajus: Either keep it as it is or check for res not to be None before unpack. Otherwise it will fail :)

PR #1029, opened 2022-05-26
Send full user report asynchronously on request

acasajus:
User can request to be sent a full report of all the info known by SL to his maibox

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -263,6 +263,7 @@ JOB_BATCH_IMPORT = "batch-import"
 JOB_DELETE_ACCOUNT = "delete-account"
 JOB_DELETE_MAILBOX = "delete-mailbox"
 JOB_DELETE_DOMAIN = "delete-domain"
+JOB_SEND_USER_REPORT = "send-user-report"
 
 # for pagination
 PAGE_LIMIT = 20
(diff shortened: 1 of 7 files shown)


nguyenkims (commented):
I tested the PR and it works well 👌

Some comments:
1) I wonder if we can ignore foreign key columns as they don't contain readable info for users. For example user can't understand `Alias.mailbox_id, Alias.original_owner_id` . For some foreign keys like directory_id, it's be better if we have export the directory name.

2) For email log table, can we export the following info which is more user friendly?
- contact address
- alias address
- timestamp
- email direction (forward, reply)

3) Apart from the foreign key columns, I think the following columns can be removed from the export:

User:
- otp_secret

Alias:
- ts_vector
- transfer_token
- hibp_last_check

Custom Domain
- ownership_txt_token

acasajus:
What do you mean `Alias.mailbox_id` don't contain relevant info? they have the id of the mailbox for that alias. I'm not following..

nguyenkims:
@acasajus ah I see what you meant, it's indeed easier for us to just export the foreign key in this case 👌

acasajus:
I've filtered out 
```python
    REMOVE_FIELDS = {
        "User": ("otp_secret",),
        "Alias": ("ts_vector", "transfer_token", "hibp_last_check"),
        "Domain": ("ownership_txt_token",),
    }
```

nguyenkims (approved):
LGTM! I've pushed some commits to update the wording and remove the current export alias.

PR #1035, opened 2022-05-27
Notify user cycle email

nguyenkims:
Create a notification when there's a cycle email.

### DIFF
--- a/email_handler.py
+++ b/email_handler.py
@@ -550,6 +550,18 @@ def handle_email_sent_to_ourself(alias, from_addr: str, msg: Message, user):
     # link available for 6 days as it gets deleted in 7 days
     refused_email_url = refused_email.get_url(expires_in=518400)
 
+    Notification.create(
+        user_id=user.id,
+        title=f"Email sent to {alias.email} from its own mailbox {from_addr}",
+        message=Notification.render(
+            "notification/cycle-email.html",
+            alias=alias,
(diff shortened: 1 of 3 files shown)


acasajus:
If people send themselves a message to their aliases, they'll expect something in their mailbox. If they don't see the mail we send them telling that they're doing a loop, I don't think they'll notice a notification in the dashboard... We could send them the mail with the from as the alias and to to their mailbox adding a warning to the message as we do for spam probable emails

nguyenkims:
> If people send themselves a message to their aliases, they'll expect something in their mailbox. If they don't see the mail we send them telling that they're doing a loop, I don't think they'll notice a notification in the dashboard... We could send them the mail with the from as the alias and to to their mailbox adding a warning to the message as we do for spam probable emails

@acasajus we already do that actually via this call https://github.com/simple-login/app/blob/125a2fc33761f673fd1d3dc26683cf5fb140cb02/email_handler.py#L565

The problem is this call is rate limited (in case an attacker bombs a user mailbox by spoofing an email sent from user's mailbox to a user'a alias) so this warning email might not reach users. The notification is rather a backup for this case. Our todo list includes a ticket to handle this cycle email scenario.

acasajus:
:ok_hand:  makes sense

acasajus (approved):

PR #1040, opened 2022-06-01
Separate code for proton callback handler

cquintana92:

1. Separates the account linking code used in the ProtonCallbackHandler into an `account_linking` module that can be accessed independently from the Proton partner, so we can use it from other places when importing the SimpleLogin project as a dependency.
2. It also introduces a `PartnerSubscription` table (with the corresponding migration) for Partner-managed subscriptions.
3. It provides a `SLPlan` class that we can use for refering to plans from partners, so we can sync the subscriptions from partners into SimpleLogin.


acasajus:
I think we're missing the case where the user does not have a full account. In that case we should reject the login/link with a message.

cquintana92:
I have now implemented the call using the new SL-specific endpoint, and updated the tests accordingly in order to check these new cases

app/models.py
```diff
@@ -3152,4 +3156,41 @@ class PartnerUser(Base, ModelMixin):
...
+            )
+            .first()
+        )
+        if res:
+            subscription, partner_user = res
+            return subscription
+        return None
+
+    def is_active(self):
+        return self.end_at > arrow.now()
```
nguyenkims, line 3193: should we take into account the grace period here? For example this is used for AppleSubscription:
https://github.com/simple-login/app/blob/dba56f0dae89861b64b9873626828b9ba4094bc3/app/models.py#L1976

app/account_linking.py
```diff
@@ -0,0 +1,256 @@
...
+    plan: SLPlan
+
+
+@dataclass
+class LinkResult:
+    user: User
+    strategy: str
+
+
+class LinkException(Exception):
```
nguyenkims, line 37: I think all exceptions are centralized in errors.py

app/account_linking.py
```diff
@@ -0,0 +1,256 @@
...
+        self.message = message
+
+
+class AccountAlreadyLinkedToAnotherPartnerException(LinkException):
+    def __init__(self):
+        super().__init__("This account is already linked to another Partner")
+
+
+def set_plan_for_user(user: User, plan: SLPlan, partner: Partner):
+    subs = PartnerSubscription.find_by_user_id(user.id)
```
nguyenkims, line 48: Nits: I'd rename "subs" to "sub" here to avoid plural confusion :). sub as abbreviation for subscription is used in several places in the code.

app/account_linking.py
```diff
@@ -0,0 +1,256 @@
...
+    if plan.type == SLPlanType.Free:
+        # If the plan is Free, check if there is a subscription. If there is, remove it
+        if subs is None:
+            # There was no subscription. We are done
+            return
+        else:
+            # There was a subscription. This is a downgrade. Remove it
+            PartnerSubscription.delete(subs.id)
```
nguyenkims, line 56: For sensitive operation as this one, I think we can add a log which can be handy when investigating later.

nguyenkims (commented):
I haven't been able to test all the code but overall everything looks good to me 👍. I just left some comments.

app/account_linking.py
cquintana92: Moved exceptions to `errors.py`

app/models.py
cquintana92: Added a grace period

app/account_linking.py
cquintana92: Renamed the variable

app/account_linking.py
cquintana92: Added logs to the different branches of this function

app/models.py
acasajus: Good catch :ok_hand:

PR #1059, opened 2022-06-06
only allow to copy the api key when it is created

nguyenkims:


templates/dashboard/new_api_key.html
```diff
@@ -0,0 +1,60 @@
...
+
+{% set active_page = "api_key" %}
+
+{% block default_content %}
+  <div class="card">
+    <div class="card-body">
+      <h1 class="h3">New API Key {{ api_key.name }} is created </h1>
+
+      <div class="alert alert-warning">
+        For security reason, API Key is only visible when it is created.
```
cquintana92, line 15: `reason` => `reasons`

PR #1065, opened 2022-06-08
Add missing rate limits

acasajus:
- Fix: rate limit marker was never called if successful in forget password route
- Fix: missing rate limit in coinbase_checkout


nguyenkims (approved):
LGTM!

PR #1066, opened 2022-06-08
use responseBody.Latest_receipt_info and not responseBody.Receipt.In_app

nguyenkims:
This bug is reported by users not having their subscription renewal taken into account in SL.
As recommended in https://developer.apple.com/documentation/appstorereceipts/responsebody/receipt/in_app


acasajus:
Is there a way to test this besides directly in prod?

nguyenkims:
> Is there a way to test this besides directly in prod?

We can mock the apple request but I don't think this test is really useful.

PR #1067, opened 2022-06-09
Do not report complaints for deleted aliases

acasajus:
- In the case of deleted aliases, we no longer have the user so skip reporting the complaint back to the user
- For domain deleted ones, use the user orm relationship


nguyenkims (approved):
LGTM!

PR #1075, opened 2022-06-12
handle subscription_payment_refunded event

nguyenkims:
The corresponding subscription will be updated to reflect the refund.


acasajus (approved):
Does paddle offer some mock service to test it out?

nguyenkims:
@acasajus yes they have a log of all events they sent.

PR #1079, opened 2022-06-13
Fix alias token

nguyenkims:


acasajus:
Replaced by https://github.com/simple-login/app/pull/1080

PR #1090, opened 2022-06-14
Refactor unsubscribe handling

acasajus:


app/handler/unsubscribe_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+    def encode(unsub: UnsubscribeData) -> str:
+        if unsub.action == UnsubscribeAction.DisableAlias:
+            return f"{unsub.data}="
+        if unsub.action == UnsubscribeAction.DisableContact:
+            return f"{unsub.data}_"
+        if unsub.action == UnsubscribeAction.UnsubscribeNewsletter:
+            return f"{unsub.data}*"
+
+    @classmethod
+    def decode(cls, data: str) -> Optional[UnsubscribeData]:
```
nguyenkims, line 30: can this method be `staticmethod` as well?

app/models.py
```diff
@@ -238,6 +220,11 @@ class BlockBehaviourEnum(EnumE):
     return_5xx = 1
 
 
+class UnsubscribeBehaviourEnum(EnumE):
```
nguyenkims, line 223: This should rather be in the next PR right?

app/models.py
```diff
@@ -1495,15 +1488,15 @@ def unsubscribe_link(self, contact: Optional["Contact"] = None) -> (str, bool):
         The mailto: method is preferred
         """
         if contact:
-            if UNSUBSCRIBER:
-                return f"mailto:{UNSUBSCRIBER}?subject={contact.id}_", True
+            if config.UNSUBSCRIBER:
+                return f"mailto:{config.UNSUBSCRIBER}?subject={contact.id}_", True
```
nguyenkims, line 1492: Should we use the UnsubscribeEncoder.encode here?

app/handler/unsubscribe_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+
+@dataclass
+class UnsubscribeData:
+    action: UnsubscribeAction
+    data: int
+
+
+class UnsubscribeEncoder:
+    @staticmethod
+    def encode(unsub: UnsubscribeData) -> str:
```
nguyenkims, line 21: I don't see this method used anywhere.

templates/dashboard/setting.html
```diff
@@ -504,6 +504,38 @@
     </div>
     <!-- END One-click subscribe -->
 
+      <div class="card" id="unsub-behaviour">
```
nguyenkims, line 507: should this be part of the next PR?

templates/emails/transactional/unsubscribe-block-contact.txt.jinja2
```diff
@@ -5,7 +5,7 @@ Hi
 
 {{ contact.website_email }} can no longer send emails to {{ alias.email }}
 
-{{ contact.website_email }} is blocked thanks to the "One-click unsubscribe” provided by your mailbox service.
+{{ contact.website_email }} is blocked thanks to the "One-click unsubscribe" provided by your mailbox service.
```
nguyenkims, line 8: 👌

templates/dashboard/setting.html
acasajus: Good catch. Removed.

app/handler/unsubscribe_encoder.py
acasajus: Removed for the next PR

app/models.py
acasajus: :+1:

app/models.py
acasajus: I was leaving it for the next but this can simplified for this one. I've removed this function and moved the logic to the encoder.

nguyenkims (approved):
LGTM!

PR #1095, opened 2022-06-16
Allow to create users from partner

cquintana92:

1. Adds a new optional parameter `from_partner` for the user creation. When this parameter is set, it will not send any notification to the user nor give the user any trial period.
2. Set the parameter explicitly to False when logging in with Proton (the user has started the flow, so even if it's linked to a partner, it's not created by a partner)

### DIFF
--- a/tests/test_account_linking.py
+++ b/tests/test_account_linking.py
@@ -77,10 +77,9 @@ def test_get_strategy_unexistant_sl_user():
 
 def test_login_case_from_partner():
     partner = get_proton_partner()
-    external_user_id = random_string()
     res = process_login_case(
         random_link_request(
-            external_user_id=external_user_id,
+            external_user_id=random_string(),
             from_partner=True,
         ),
         partner,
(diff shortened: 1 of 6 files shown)


tests/test_account_linking.py
```diff
@@ -72,6 +75,24 @@ def test_get_strategy_unexistant_sl_user():
...
+            from_partner=True,
+        ),
+        partner,
+    )
+
+    assert res.strategy == NewUserStrategy.__name__
+    assert res.user is not None
+    assert User.FLAG_CREATED_FROM_PARTNER == (
+        res.user.flags & User.FLAG_CREATED_FROM_PARTNER
+    )
```
acasajus, line 93:
```suggestion
def test_login_case_from_partner():
    partner = get_proton_partner()
    external_user_id = random_string()
    res = process_login_case(
        random_link_request(
            external_user_id=external_user_id,
            from_partner=True,
        ),
        partner,
    )

    assert res.strategy == NewUserStrategy.__name__
    assert res.user is not None
    assert User.FLAG_CREATED_FROM_PARTNER == (
        res.user.flags & User.FLAG_CREATED_FROM_PARTNER
    )
    
def test_login_case_from_web():
    partner = get_proton_partner()
    external_user_id = random_string()
    res = process_login_case(
        random_link_request(
            external_user_id=external_user_id,
            from_partner=False,
        ),
        partner,
    )

    assert res.strategy == NewUserStrategy.__name__
    assert res.user is not None
    assert 0 == (
        res.user.flags & User.FLAG_CREATED_FROM_PARTNER
    )
```

acasajus (approved):

PR #1100, opened 2022-06-17
Fix hover color in login with proton button

cquintana92:
Updated background on hover for the "Login with Proton" buttons in order to use the same as the Proton background


acasajus (approved):
lgmt

PR #1101, opened 2022-06-17
Send email to users with a subscription and a partner plan upgrade

acasajus:
Send an email once to users that have a subscription and get upgraded by a partner.

### DIFF
--- a/app/account_linking.py
+++ b/app/account_linking.py
@@ -2,13 +2,25 @@ from abc import ABC, abstractmethod
 from arrow import Arrow
 from dataclasses import dataclass
 from enum import Enum
+
+from flask import url_for
 from newrelic import agent
 from typing import Optional
 
+from app import config
 from app.db import Session
+from app.email_utils import send_email_at_most_times, render
 from app.errors import AccountAlreadyLinkedToAnotherPartnerException
 from app.log import LOG
-from app.models import PartnerSubscription, Partner, PartnerUser, User
+from app.models import (
+    PartnerSubscription,
+    Partner,
+    PartnerUser,
+    User,
+    AppleSubscription,
+    Subscription,
+)
+from app.proton.utils import is_proton_partner
 from app.utils import random_string
 
 
@@ -36,6 +48,36 @@ class PartnerLinkRequest:
 class LinkResult:
     user: User
     strategy: str
+
+
+def send_double_subscription_if_needed(partner_user: PartnerUser):
+    sub = partner_user.user.get_active_subscription()
+    if isinstance(sub, AppleSubscription):
+        channel = "Apple"
+    elif isinstance(sub, Subscription):
+        channel = "Paddle"
+    else:
+        return
+    send_email_at_most_times(
+        partner_user.user,
+        config.ALERT_DUAL_SUBSCRIPTION_WITH_PARTNER,
+        partner_user.user.email,
+        f"You have two subscriptions in SimpleLogin",
+        render(
+            "transactional/double-subscription-partner.txt.jinja2",
+            is_proton=is_proton_partner(partner_user.partner),
+            partner=partner_user.partner,
+            subscription_channel=channel,
+            cancel_link=url_for("dashboard.billing"),
+        ),
+        render(
+            "transactional/double-subscription-partner.html",
+            is_proton=is_proton_partner(partner_user.partner),
+            partner=partner_user.partner,
+            subscription_channel=channel,
+            cancel_link=url_for("dashboard.billing"),
+        ),
+    )
 
 
 def set_plan_for_partner_user(partner_user: PartnerUser, plan: SLPlan):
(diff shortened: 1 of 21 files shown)


app/account_linking.py
```diff
@@ -38,6 +51,38 @@ class LinkResult:
     strategy: str
 
 
+def send_double_subscription_if_needed(partner_user: PartnerUser):
+    sub = partner_user.user.get_active_subscription()
+    if isinstance(sub, CoinbaseSubscription):
+        channel = "Coinbase"
```
nguyenkims, line 57: Coinbase subscription isn't renewable so users can't cancel it.

app/models.py
```diff
@@ -591,57 +591,62 @@ def create(cls, email, name="", password=None, from_partner=False, **kwargs):
...
-        if apple_sub and apple_sub.is_valid():
-            return True
+        sub = self.get_active_subscription()
+        if sub is None:
+            return False
 
-        manual_sub: ManualSubscription = ManualSubscription.get_by(user_id=self.id)
-        if manual_sub and not manual_sub.is_giveaway and manual_sub.is_active():
-            return True
+        if isinstance(sub, PartnerSubscription):
```
nguyenkims, line 643: ah why do we return False for partner subscription?

app/account_linking.py
```diff
@@ -38,6 +51,38 @@ class LinkResult:
...
+            partner=partner_user.partner,
+            subscription_channel=channel,
+            cancel_link=url_for("dashboard.billing"),
+        ),
+        render(
+            "transactional/double-subscription-partner.html",
+            is_proton=is_proton_partner(partner_user.partner),
+            partner=partner_user.partner,
+            subscription_channel=channel,
+            cancel_link=url_for("dashboard.billing"),
```
nguyenkims, line 81: the cancel link here is only for paddle subscription actually. For apple subscription, it's to be done in user icloud account.

app/account_linking.py
```diff
@@ -38,6 +51,38 @@ class LinkResult:
...
+    sub = partner_user.user.get_active_subscription()
+    if isinstance(sub, CoinbaseSubscription):
+        channel = "Coinbase"
+    elif isinstance(sub, AppleSubscription):
+        channel = "Apple"
+    elif isinstance(sub, Subscription):
+        channel = "Paddle"
+    else:
+        return
+    send_email_at_most_times(
```
nguyenkims, line 64: do you know why do we use `send_email_at_most_times` and not just `send_email` here?

app/models.py
```diff
@@ -591,57 +591,62 @@ def create(cls, email, name="", password=None, from_partner=False, **kwargs):
 
         return user
 
-    # region Billing
-    def lifetime_or_active_subscription(self) -> bool:
-        """True if user has lifetime licence or active subscription"""
-        if self.lifetime:
-            return True
-
+    def get_active_subscription(
```
nguyenkims, line 594: not related to the double subscription email but I think we can rename User.get_subscription to something like User.get_paddle_subscription to avoid confusion with this new method.

nguyenkims (commented):
The PR looks good, I just left some comments.

app/account_linking.py
acasajus: I'll remove it.

app/models.py
acasajus: It's not a paid subscription in SL. Do we want to consider it a paid one? We cannot manage it from SL's side.

app/account_linking.py
acasajus: Could me show that info in the billing template? If you have apple put a notice that to modify the billing you have to go to apple. And if you've got coinbase then you can't do jack. But we can show until when the sub is active in any case, even if you have a partner sub.

app/account_linking.py
acasajus: We said that we only wanted to send it once even if the user re-linked. Do you prefer to send a mail at each update? Keep in mind that a renewal in the partner side will also trigger this code.

app/models.py
acasajus: :+1:

app/models.py
nguyenkims: I think we can consider this subscription as paid in SL too. Btw `User.is_paid` doesn't seem to be used anywhere (rather in the metrics which can be removed) so I think whether a subscription is paid or not doesn't have any effect in SL.

app/account_linking.py
nguyenkims: I think it might be easier to distinguish this in the email. For apple user, we can tell them to unsubscribe in their iCloud account. We can also optionally send them this link https://support.apple.com/en-us/HT202039

app/account_linking.py
nguyenkims: ah ok, let's send it only once then.

app/models.py
acasajus: Just to confim: It's also used for showing paid referrals. If we want to count them i'll change it.

app/account_linking.py
acasajus: Ok. I'll change it.

app/account_linking.py
acasajus: :+1:

app/models.py
acasajus: changed to count as paid

nguyenkims: Yeah it makes sense to count them in referral too.

nguyenkims (approved):
LGTM!

PR #1107, opened 2022-06-22
allow to hide some public domains and set their order

nguyenkims:


acasajus (approved):
LGTM. Just a minor comment. I would add an extra test with premium plan and hidden domain to test that multiple `filter_by` don't override one on top of the other.

nguyenkims:
> LGTM. Just a minor comment. I would add an extra test with premium plan and hidden domain to test that multiple `filter_by` don't override one on top of the other.

I think we don't have to test filter_by as it's meant to be combined. If not, it's going to be a huge bug in sqlalchemy :)

PR #1109, opened 2022-06-22
Updated version numbers on doc

Spitfireap:
People should upgrade from 2.X.X (supposing some of them still have this version) to 3.4.0, then migrate to 4.X.X. It's more incremental and easier to follow for people upgrading from 3.4.0 (which should be the majority)


nguyenkims (approved):
Thanks for making this PR!

PR #1113, opened 2022-06-27
Add state management to job

acasajus:
In order to be able to retry jobs we need to have a state for them. This will allow to retry jobs if they have been taken for some time and haven't finished.


acasajus:
Yes, this one is just to start adding the state for all jobs. So when we merge the next one all jobs have the proper state. It's the comment on the runner 
```
            # TODO: Change This condition after deploying this MR
            # to Job.state == ready or (Job.state == taken and job.taken_at < arrow.now.shift(minutes=-10))
```

nguyenkims (approved):

nguyenkims:
Yea 30 mins should be enough.

PR #1115, opened 2022-06-28
Send welcome email when user created by login with proton

cquintana92:


app/account_linking.py
```diff
@@ -182,6 +182,9 @@ def process(self) -> LinkResult:
         )
         Session.commit()
 
+        if not new_user.created_by_partner:
```
nguyenkims, line 185: do you know in what case `new_user.created_by_partner` can be True here?

tests/models/test_user.py
```diff
@@ -8,6 +8,7 @@ def test_available_sl_domains(flask_client):
     user = create_new_user()
 
     assert set(user.available_sl_domains()) == {"d1.test", "d2.test", "sl.local"}
+    assert user.created_by_partner is False
```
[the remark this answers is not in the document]
cquintana92: Moved it into a different test

app/account_linking.py
cquintana92: When called from the partner_api this flag will be True

PR #1122, opened 2022-06-29
Add djlint

cquintana92:

1. Add `djlint` as a HTML format checker. We cannot use the latest version due to a dependency conflict in `importlib` with `flake8` , so we're kinda stuck on `0.7.3`. But that should be OK because we only need it on development time for HTML formatting.
2. Add `djlint --check` to CI.
3. Add the explanation to `CONTRIBUTING.md`
4. Fix many of the warnings reported by `djlint`.

Keep in mind that quite a few errors/warnings have been ignored in `pyproject.toml` as they were either false-positives or would have made the changes even harder to review.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,6 +77,7 @@ jobs:
         run: |
           poetry run black --check .
           poetry run flake8
+          poetry run djlint --check templates
 
       - name: Run db migration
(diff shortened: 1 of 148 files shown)


nguyenkims (approved):
The formatting looks really nice and black-like 👌

PR #1130, opened 2022-06-30
Feature: Use new job status to retry killed jobs

acasajus:
It also changes the `run_at` condition. If we set a window for it it could be that the job is killed because something fails consistent and when it's fixed the job is out of the valid `run_at` window. This MR only sets that `run_at` has to be in the past + 10 mins


job_runner.py
```diff
@@ -220,19 +222,33 @@ def process_job(job: Job):
         LOG.e("Unknown job name %s", job.name)
 
 
+def get_jobs_to_run() -> List[Job]:
+    # Get jobs that match all conditions:
+    #  - Job.state == ready OR (Job.state == taken AND Job.taken_at < now - 30 mins AND Job.attempts < 5)
+    #  - Job.run_at is Null OR Job.run_at < now + 10 mins
```
nguyenkims, line 228: Currently we run jobs that have now - 1h < run_at < now + 1h, should we use the same condition here?

acasajus: That's why this MR also changes the run_at condition. If we set a window for it it could be that the job is killed because something fails   consistent and when it's fixed the job is out of the valid run_at window. This MR only sets that run_at has to be in the past + 10 mins to prevent that issue.

acasajus: The case that this fixes is:
 - Create a job with run_at 6AM
 - At 4AM there's an issue and the jobs stop being processed (for instnance a DB issue)
 - It's fixed at 8AM
 
 This job will never run with the previous condition

nguyenkims: That's indeed a possible scenario.

nguyenkims (approved):
LGTM!

PR #1138, opened 2022-07-04
Update pre-commit

acasajus:


app/email_utils.py
```diff
@@ -19,7 +19,7 @@
 from typing import Tuple, List, Optional, Union
 
 import arrow
-import dkim
+import dkim  # pylint: disable=syntax-error
```
nguyenkims, line 22: why do we need to add `disable=syntax-error` here?

nguyenkims (approved):
LGTM, just left a question.

nguyenkims:
Btw now that pre-commit is used, should we upgrade CONTRIBUTING.md to ask users to use pre-commit instead?

app/email_utils.py
cquintana92: Because `dkimpy` has a `# type:` comment that should be placed before the docstring. Otherwise pylint detects it as an error unless we add this annotation.

```
def verify(message, logger=None, dnsfunc=get_txt, minkey=1024,
        timeout=5, tlsrpt=False):
    """Verify the first (topmost) DKIM signature on an RFC822 formatted message.
    @param message: an RFC822 formatted message (with either \\n or \\r\\n line endings)
    @param logger: a logger to which debug info will be written (default None)
    @param timeout: number of seconds for DNS lookup timeout (default = 5)
    @param tlsrpt: message is an RFC 8460 TLS report (default False)
     False: Not a tlsrpt, True: Is a tlsrpt, 'strict': tlsrpt, invalid if
     service type is missing. For signing, if True, length is never used.
    @return: True if signature verifies or False otherwise
    """
    # type: (bytes, any, function, int) -> bool
```

acasajus:
> Btw now that pre-commit is used, should we upgrade CONTRIBUTING.md to ask users to use pre-commit instead?

Yes we should.

app/email_utils.py
nguyenkims: ah ok I see

.pre-commit-config.yaml
```diff
@@ -1,10 +1,26 @@
...
-    rev: 4.0.1
+      - id: check-yaml
+      - id: trailing-whitespace
+  - repo: https://github.com/psf/black
+    rev: 22.3.0
     hooks:
-    -   id: flake8
+      - id: black
+        language_version: "3.10"
+  - repo: https://gitlab.com/pycqa/flake8
```
nguyenkims, line 13: should flake8 be replaced by pylint?

acasajus: I've seen cases where flake8 catches stuff that pyling doesn't and viceversa. I would keep them both for some time and see how it goes.

acasajus:
> I run pre-commit on all files using this command `pre-commit run --all-files` but got this error, does that mean I need to upgrade my python to 3.10?

Yes. It will require 3.10. We've been using 3.7 and 3.10 for some time in the tests.

acasajus:
I've set python3.9 and python3.10. It will work with both for now :+1:

PR #1145, opened 2022-07-05
Fix: Check if required session headers exist

acasajus:


app/auth/views/proton.py
```diff
@@ -65,13 +68,16 @@ def proton_login():
...
-    session["oauth_action"] = extract_action().value
+    session[SESSION_STATE_KEY] = state
+    session[SESSION_ACTION_KEY] = extract_action().value
     return redirect(authorization_url)
 
 
 @auth_bp.route("/proton/callback")
 def proton_callback():
+    if SESSION_STATE_KEY not in session or SESSION_STATE_KEY not in session:
+        flash("Invalid OAuth state", "error")
```
nguyenkims, line 79: should we show a more user-friendly error message here? Something like "unknown error, please retry"?

nguyenkims (approved):
Just left a suggestion

PR #1148, opened 2022-07-06
Allow users to keep the original unsub behaviour

acasajus:
Users can choose if the unsub link has to disable the alias/block the contact or if it has to preserve the original unsubscribe behaviour and send back the original unsubscribe


templates/dashboard/setting.html
```diff
@@ -584,6 +584,42 @@
...
+              <b>Disable alias or block contact</b> means that the unsubscribe action will disable that alias and you will stop receiving emails
+              to it, or the contact will be disabled based on your preferences.
+              <br>
+            </div>
+            <form method="post"
+                  action="#change-unsubscribe-behaviour"
+                  class="form-inline">
+              <input type="hidden" name="form-name" value="change-unsubscribe-behaviour">
+              <select class="form-control mr-sm-2" name="unsubscribe-behaviour">
+                <option value="{{ UnsubscribeBehaviourEnum.PreserveOriginal.name }}" {% if current_user.unsub_behaviour.value == UnsubscribeBehaviourEnum.PreserveOriginal.value %}
```
nguyenkims, line 608: I have this error `jinja2.exceptions.UndefinedError: 'UnsubscribeBehaviourEnum' is undefined` when opening this page

templates/dashboard/setting.html
```diff
@@ -584,6 +584,42 @@
           </div>
         </div>
         <!-- END One-click subscribe -->
+        <div class="card" id="unsub-behaviour">
```
nguyenkims, line 587: can we add the new option into this existing settings instead?
image

app/dashboard/views/unsubscribe.py
```diff
@@ -68,3 +70,46 @@ def block_contact(contact_id):
         )
     else:  # ask user confirmation
         return render_template("dashboard/block_contact.html", contact=contact)
+
+
+@dashboard_bp.route("/encoded", methods=["GET"])
+@login_required
+def encoded_unsubscribe():
```
nguyenkims, line 77: can we add a small doc on what to expect as input for this route?

app/dashboard/views/unsubscribe.py
```diff
@@ -68,3 +70,46 @@ def block_contact(contact_id):
         )
     else:  # ask user confirmation
         return render_template("dashboard/block_contact.html", contact=contact)
+
+
+@dashboard_bp.route("/encoded", methods=["GET"])
```
nguyenkims, line 75: can we make the url more explicit, for example `/encoded-unsubscribe` instead of just `/encoded`?

app/handler/unsubscribe_encoder.py
```diff
@@ -25,52 +40,106 @@ class UnsubscribeLink:
...
+            # Initial 0 is the version number. If we need to add support for extra use-cases we can bump up this number
+            data = (0, data.alias_id, data.recipient, data.subject)
+        payload = (action.value, data)
+        serialized_data = (
+            base64.urlsafe_b64encode(json.dumps(payload).encode("utf-8"))
+            .rstrip(b"=")
+            .decode("utf-8")
+        )
+        signed_data = cls._get_signer().sign(serialized_data).decode("utf-8")
+        return f"{UNSUB_PREFIX}.{signed_data}"
```
nguyenkims, line 73: can we run into the limit of email subject length?

migrations/versions/2022_070619_b0101a66bb77_add_unsubscribe_behaviour.py
```diff
@@ -0,0 +1,36 @@
...
+    unsub_enum.create(op.get_bind())
+    op.add_column('users', sa.Column('unsub_behaviour', unsub_enum, default='PreserveOriginal', server_default='DisableAlias', nullable=False))
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('users', 'unsub_behaviour')
+    unsub_enum = get_enum()
+    unsub_enum.drop(op.get_bind())
```
nguyenkims, line 35: Nice! Do you know if the enum is indeed removed when a rollback is run?

app/models.py
```diff
@@ -252,6 +252,11 @@ class JobState(EnumE):
     error = 3
 
 
+class UnsubscribeBehaviourEnum(EnumE):
+    DisableAlias = 0
+    PreserveOriginal = 1
```
nguyenkims, line 257: should we also add BlockSender here?

app/handler/unsubscribe_generator.py
```diff
@@ -1,34 +1,90 @@
...
+            unsub = UnsubscribeData(UnsubscribeAction.DisableContact, contact.id)
         else:
-            unsub_link = UnsubscribeEncoder.encode(
-                UnsubscribeAction.DisableAlias, alias.id
+            unsub = UnsubscribeData(UnsubscribeAction.DisableAlias, alias.id)
+        return self._add_unsubscribe_header(message, unsub)
+
+    def _generate_header_with_original_behaviour(
+        self, alias: Alias, message: Message
+    ) -> Message:
```
nguyenkims, line 31: Do you think we can add some doc for some steps in this method? It's a little bit hard to know what's going on in the implementation.

templates/dashboard/setting.html
cquintana92: Fixed it

acasajus: Thanks :+1:

templates/dashboard/setting.html
acasajus: Ok. Will do.

app/handler/unsubscribe_encoder.py
acasajus: We potentially can. The limit set by the rfc is almost 1000 chars if I remember correctly but I'm sure some clients won't allow more than 256/512 chars. I'm trying to encode it as tight as I can.

app/dashboard/views/unsubscribe.py
acasajus: Will do.

app/handler/unsubscribe_generator.py
acasajus: Will do.

app/dashboard/views/unsubscribe.py
acasajus: Will do.

app/dashboard/views/unsubscribe.py
acasajus: On second thought. The route for this will be `/dashboard/unsubscribe/encoded` as it is. Maybe we don't need to add another `unsubscribe` at the end?

templates/dashboard/setting.html
acasajus: image

app/handler/unsubscribe_generator.py
acasajus: Done

app/models.py
acasajus: Done

app/dashboard/views/unsubscribe.py
acasajus: Done

templates/dashboard/setting.html
acasajus: Done

app/handler/unsubscribe_encoder.py
```diff
@@ -25,52 +40,106 @@ class UnsubscribeLink:
...
+            and type(data) is not int
+        ):
+            raise ValueError(f"Data has to be an int for an action of type {action}")
+        if action == UnsubscribeAction.OriginalUnsubscribeMailto:
+            if type(data) is not UnsubscribeOriginalData:
+                raise ValueError(
+                    f"Data has to be an UnsubscribeOriginalData for an action of type {action}"
+                )
+            # Initial 0 is the version number. If we need to add support for extra use-cases we can bump up this number
+            data = (0, data.alias_id, data.recipient, data.subject)
```
nguyenkims, line 65: I wonder if `version` should be part of the UnsubscribeOriginalData?

app/handler/unsubscribe_encoder.py
nguyenkims: For now maybe we can have a `LOG.e` in case the encoded data is too long so we know when this issue will ever happen?

app/models.py
```diff
@@ -493,6 +513,14 @@ class User(Base, ModelMixin, UserMixin, PasswordOracle):
         nullable=False,
     )
 
+    # Keep original unsub behaviour
+    unsub_behaviour = sa.Column(
```
nguyenkims, line 517: should we have a migration to convert `one_click_unsubscribe_block_sender` to this new column?

app/dashboard/views/unsubscribe.py
```diff
@@ -68,3 +70,50 @@ def block_contact(contact_id):
...
+     - data: This parameter should contain the encoded unsubscribe request.
+    """
+    if "data" not in request.args:
+        return redirect(url_for("dashboard.index"))
+
+    unsub_data = UnsubscribeHandler().handle_unsubscribe_from_request(
+        current_user, request.args.get("data")
+    )
+    if not unsub_data:
+        flash(f"Invalid unsuscribe request", "error")
```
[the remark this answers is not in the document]
acasajus: Fixed typo and the route. I don't know why I thought that the `/unsubscribe` was added by the module. Now its `/dashboard/unsubscribe/encoded/<encoded_request>` so we have a shorter url since we skip the `data` param name.

app/handler/unsubscribe_handler.py
```diff
@@ -173,3 +199,20 @@ def _check_email_is_authorized_for_alias(
...
+
+    def _unsubscribe_original_behaviour(
+        self, original_unsub_data: UnsubscribeOriginalData, user: User
+    ) -> str:
+        alias = Alias.get(original_unsub_data.alias_id)
+        if not alias:
+            return status.E508
+        if alias.user_id != user.id:
+            return status.E509
+        send_email(
```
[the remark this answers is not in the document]
acasajus: We're sending the email from the alias afaics aren't we?

```python
        send_email(
            original_unsub_data.recipient,
            original_unsub_data.subject,
            "",
            from_name=alias.email,
            from_addr=alias.email,
        )
```
We still want to add dkim signing and keep track of it for bounces don't we?

app/models.py
acasajus: I was thinking of  running a migration script just after this is deployed.

app/handler/unsubscribe_encoder.py
acasajus: If we need it in the future we can add it. For not it's not needed. It's a fail-safe in the encoding to be able to be backwards compatible if we need it since this info will be encoded in emails and can potentially be stored forever in user emails.

acasajus:
There are quite some tests in `tests/handler/test_unsubscribe*` I've added all that I come up with.

app/handler/unsubscribe_handler.py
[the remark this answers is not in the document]
nguyenkims: `send_email` always send the email from `EMAIL_DOMAIN` which isn't the same as alias domain so this email can fail DMARC alignment.

app/models.py
nguyenkims: I think we can add a sql query into the migration file, something like https://github.com/simple-login/app/blob/02a005d07647fa75b4f5271a4e56f8f1ce412fd7/migrations/versions/2021_102715_0b9150eb309d_.py#L22

```
op.execute("update users set unsub_behaviour=0 where one_click_unsubscribe_block_sender=false;")
op.execute("update users set unsub_behaviour=1 where one_click_unsubscribe_block_sender=true;")

```

nguyenkims:
> There are quite some tests in `tests/handler/test_unsubscribe*` I've added all that I come up with.

ah I mean end to end test and not automatic test.

app/dashboard/views/unsubscribe.py
acasajus: Yep. I was wrong. Fixed. Not its `/dashboard/unsubscribe/encoded/<request>`

app/handler/unsubscribe_encoder.py
acasajus: Added

app/models.py
acasajus: Ah. I though you didn't want non auto-generated alembic ops in the migrations. That's how we wanted to do it for proton partner and did a manual migration. Will add this to the migration :+1:

PR #1155, opened 2022-07-13
Restrict the number of free alias for new free users

acasajus:
- Old accounts will keep the old value (pending config change)
- To enable this change in prod, we need to:
  - Add the new config
  - Add the flag to current users in prod
  - Redeploy

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -295,6 +295,7 @@ class User(Base, ModelMixin, UserMixin, PasswordOracle):
 
     FLAG_FREE_DISABLE_CREATE_ALIAS = 1 << 0
     FLAG_CREATED_FROM_PARTNER = 1 << 1
+    FLAG_FREE_NEW_ALIAS_LIMIT = 1 << 2
 
     email = sa.Column(sa.String(256), unique=True, nullable=False)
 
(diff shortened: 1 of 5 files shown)


app/models.py
```diff
@@ -488,7 +489,7 @@ class User(Base, ModelMixin, UserMixin, PasswordOracle):
     # bitwise flags. Allow for future expansion
     flags = sa.Column(
         sa.BigInteger,
-        default=FLAG_FREE_DISABLE_CREATE_ALIAS,
+        default=FLAG_FREE_DISABLE_CREATE_ALIAS | FLAG_FREE_NEW_ALIAS_LIMIT,
```
cquintana92, line 492: Should we also update the code that sets the flag when a user is created from a partner? https://github.com/simple-login/app/blob/408e3544410bdea4752ade645a95e06f6c9d5105/app/models.py#L562

app/models.py
```diff
@@ -295,6 +295,7 @@ class User(Base, ModelMixin, UserMixin, PasswordOracle):
 
     FLAG_FREE_DISABLE_CREATE_ALIAS = 1 << 0
     FLAG_CREATED_FROM_PARTNER = 1 << 1
+    FLAG_FREE_OLD_ALIAS_LIMIT = 1 << 2
```
nguyenkims, line 298: If I understand correctly, after deployment of this PR, we're going to run a migration on `users` table to add this flag for all existing users right? In this case we can indeed merge this PR now and choose to run the query when the  change is public.

nguyenkims (approved):

app/models.py
acasajus: :+1:

PR #1164, opened 2022-07-17
remove 15 hardcoding

nguyenkims:


acasajus (approved):
Nits: I added a function to get the number of free aliases per user. Do we want to use it? On the other hand, if we leave it without specifying the number of aliases they can go to the web and check the offer and won't have outdated info in their mailboxes.

nguyenkims:
@acasajus I think we can add use this function in https://github.com/simple-login/app/pull/1155 or after it's being merged.

PR #1189, opened 2022-07-26
remove envs used for proton beta link

nguyenkims:


app/proton/utils.py
```diff
@@ -29,8 +27,4 @@ def is_proton_partner(partner: Partner) -> str:
 def is_connect_with_proton_enabled() -> bool:
     if CONNECT_WITH_PROTON:
```
cquintana92, line 28:
```
return CONNECT_WITH_PROTON is not None
```

acasajus: `CONNECT_WITH_PROTON` is boolean 
```python
CONNECT_WITH_PROTON = "CONNECT_WITH_PROTON" in os.environ
```

acasajus (approved):

PR #1202, opened 2022-07-29
Mitigate dns issue with dnsbl

Spitfireap:
Spamhaus requires dns with rdns to work. To be sure that emails are not reject this commit change the postfix configuration to only take into account right response from dnsbl provider.  

This should be back ported to 4.6.2-beta also :)


nguyenkims:
@Spitfireap Thanks for making the PR! Can you please cherry pick it for the branch `4.6.2-beta` instead? 4.6.2-beta will be renamed 4.6.2 when it's stable for a release and after that merged back to master.

PR #1213, opened 2022-08-02
fix paddle refund

nguyenkims:


server.py
```diff
@@ -601,9 +601,9 @@ def paddle():
                 )
                 return "No such subscription"
 
-            plan_id = request.form["subscription_plan_id"]
+            plan_id = int(request.form["subscription_plan_id"])
             if request.form["refund_type"] == "full":
-                if plan_id == PADDLE_MONTHLY_PRODUCT_ID:
+                if plan_id in PADDLE_MONTHLY_PRODUCT_IDS:
```
acasajus, line 606: Where is this defined? in config.py I can only see `PADDLE_MONTHLY_PRODUCT_ID`

nguyenkims: It's defined here https://github.com/simple-login/app/blob/97805173cbfa6359b0ab5c85acc531deb39ffa94/app/config.py#L192-L191

acasajus: Ah. My bad. I missed it. Then what does the `PADDLE_MONTHLY_PRODUCT_ID` do? Can we remove it in a future PR?

nguyenkims: PADDLE_MONTHLY_PRODUCT_ID is the currently used product ID whereas PADDLE_MONTHLY_PRODUCT_IDS (with a `S` at the end) contains all product Ids SL has used. We have changed the product ID several times in the past and it's there for compatibility reason (people who bought SL before should be able to continue using it).

PR #1221, opened 2022-08-08
Rate limiting depending on user authenticated status

cquintana92:
This PR changes how we are obtaining a the key for the rate limiting process.

- If the user is not authenticated, the request source IP is used.
- If the user is authenticated, the user ID is used.

### DIFF
--- a/app/extensions.py
+++ b/app/extensions.py
@@ -1,13 +1,23 @@
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
-from flask_login import LoginManager
+from flask_login import current_user, LoginManager
 
 login_manager = LoginManager()
 login_manager.session_protection = "strong"
 
-# Setup rate limit facility
-limiter = Limiter(key_func=get_remote_address)
 
+# We want to rate limit based on:
+# - If the user is not logged in: request source IP
+# - If the user is logged in: user_id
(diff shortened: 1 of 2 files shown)


nguyenkims (approved):
I wonder if we can have a test for this change?

app/extensions.py
```diff
@@ -1,13 +1,23 @@
...
-limiter = Limiter(key_func=get_remote_address)
 
+# We want to rate limit based on:
+# - If the user is not logged in: request source IP
+# - If the user is logged in: user_id
+def __key_func():
+    if current_user.is_authenticated:
+        return current_user.id
+    else:
+        return get_remote_address()
```
acasajus, line 16:
```suggestion
    if current_user.is_authenticated:
        return f"userid:{current_user.id}"
    else:
        ip_addr = get_remote_address()
        return f"ip:{ip_addr}"
```

acasajus: Just to scope a bit the field by which we're rate limiting.

cquintana92:
> I wonder if we can have a test for this change?

I managed to add a few tests that prove the behaviour is the one we want.
It's a bit hacky dealing with the rate_limiter in test time, but I think I managed to add a meaningful explanation to the test file.

PR #1224, opened 2022-08-09
Generate a web session from an api key

acasajus:


app/auth/views/api_to_cookie.py
```diff
@@ -0,0 +1,33 @@
...
+from app.models import ApiToCookieToken
+from app.utils import sanitize_next_url
+
+
+@auth_bp.route("/api_to_cookie", methods=["POST"])
+def api_to_cookie():
+    if current_user.is_authenticated:
+        LOG.d("user is already authenticated, redirect to dashboard")
+        return redirect(url_for("dashboard.index"))
+    body = request.json
```
nguyenkims, line 16: This is the url that app should open in the webview right? Do you know if the app can open a webview with a POST request? This URL is also used by the web extension and opening an URL with a POST request requires workaround like https://stackoverflow.com/a/17089124

app/models.py
```diff
@@ -3296,3 +3296,12 @@ class NewsletterUser(Base, ModelMixin):
 
     user = orm.relationship(User)
     newsletter = orm.relationship(Newsletter)
+
+
+class ApiToCookieToken(Base, ModelMixin):
+
+    __tablename__ = "api_cookie_token"
+    code = sa.Column(sa.String(128), unique=True, nullable=False)
+    user_id = sa.Column(sa.ForeignKey(User.id, ondelete="cascade"), nullable=True)
```
nguyenkims, line 3305: can we add a foreign key to ApiKey to know which api key has been used to create a token?

acasajus: Added

app/auth/views/api_to_cookie.py
acasajus: Ah. Good catch. I've changed it to a `GET` request :+1:

app/auth/views/api_to_cookie.py
```diff
@@ -0,0 +1,33 @@
...
+def api_to_cookie():
+    if current_user.is_authenticated:
+        LOG.d("user is already authenticated, redirect to dashboard")
+        return redirect(url_for("dashboard.index"))
+    code = request.args.get("token")
+    if not code:
+        flash("Missing token", "error")
+        return redirect(url_for("auth.login"))
+
+    token = ApiToCookieToken.get_by(code=code)
```
nguyenkims, line 21: Nits: I wonder if ApiToCookieToken.code should be renamed to ApiToCookieToken.token to be more consistent?

nguyenkims (approved):
LGTM, just left a nits :)

app/auth/views/api_to_cookie.py
acasajus: ApiKeys also use `code` in the db for the token. That's why I went for code here also.

PR #1225, opened 2022-08-10
Refine activation code generation.

weakish:
Previously, activation code genaration invoked a redudant `str()` conversion
(`secrets.cohice(string.digits)` already returns a string).
Also, secrets.choice was called six times, which is slower than just one call.
I hope this refinement make the code more readable and faster.

### DIFF
--- a/app/api/views/auth.py
+++ b/app/api/views/auth.py
@@ -1,5 +1,4 @@
 import secrets
-import string
 
 import facebook
 import google.oauth2.credentials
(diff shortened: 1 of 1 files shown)


acasajus:
Hey, thanks for contributing! May I suggest a small improvement. 

At the top define
```python
ACTIVATION_CODE_NUM_DIGITS=6
```

and then we can do 
```python
str(secrets.randbelow(10**ACTIVATION_CODE_NUM_DIGITS)).zfill(ACTIVATION_CODE_NUM_DIGITS)
```

:+1:

weakish:
I applied suggestion in e2b147f (I defined the const in config.py instead of the top of auth.py since it seems most constants are defined there.)

PR #1226, opened 2022-08-11
Support next with Proton Link

cquintana92:

- Support for `next` in the "Link Proton account" OAuth flow
- Support for passing a `scheme` to the Proton login/link
- Fix a server error when re-linking a Proton account (user already had the same Proton account linked)

### DIFF
--- a/app/auth/views/proton.py
+++ b/app/auth/views/proton.py
@@ -146,10 +146,11 @@ def proton_callback():
     handler = ProtonCallbackHandler(proton_client)
     proton_partner = get_proton_partner()
 
+    next_url = session.get("oauth_next")
     if action == Action.Login:
         res = handler.handle_login(proton_partner)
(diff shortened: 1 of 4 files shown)


app/auth/views/proton.py
```diff
@@ -166,5 +167,4 @@ def check_status_code(response: requests.Response) -> requests.Response:
     if res.redirect:
         return after_login(res.user, res.redirect, login_from_proton=True)
 
-    next_url = session.get("oauth_next")
     return after_login(res.user, next_url, login_from_proton=True)
```
acasajus, line 170: To avoid having to pass the next_url to handle_link we could do:

```python
redirect_url = next_url or res.redirect
after_login(res.user, redirect_url, login_from_proton=True)
```

acasajus (approved):

PR #1230, opened 2022-08-12
Fixes for connect with proton on mobile

cquintana92:
This PR adds the following changes for making `Connect with Proton` available on mobile:

- `DELETE /api/setting/unlink_proton_account` for unlinking the Proton account from the API
- Add `connected_proton_address: Optional[str]` on the response of `GET /api/user_info`
- Fixes on the `/api_to_cookie` endpoint for ensuring we redirect the user even if they're already authenticated
- Fixes in the Proton OAuth callback for invalid flows
-

### DIFF
--- a/app/api/views/setting.py
+++ b/app/api/views/setting.py
@@ -12,6 +12,7 @@ from app.models import (
     SenderFormatEnum,
     AliasSuffixEnum,
 )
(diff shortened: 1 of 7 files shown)


acasajus (approved):
Added an extra test.

PR #1259, opened 2022-09-01
Fix: Spelling

PurpleSn0w:
Minor change to fix spelling.


templates/dashboard/index.html
```diff
@@ -398,7 +398,7 @@
       {% endif %}
       <span class="alias-activity">{{ alias_info.nb_forward }}</span> forwards,
```
nguyenkims, line 399: should we change `forwards -> forwarded, blocks -> blocked` as well?

PurpleSn0w: Yes, good catch! Resolved in latest commit.

nguyenkims (approved):
Thanks for making the PR!

PR #1273, opened 2022-09-07
Fix: Use email directly for DomainDeletedAlias

acasajus:
If the complaint is for a `DomainDeletedAlias` use `alias.email` directly

### DIFF
--- a/app/handler/provider_complaint.py
+++ b/app/handler/provider_complaint.py
@@ -245,16 +245,22 @@ def handle_complaint(message: Message, origin: ProviderComplaintOrigin) -> bool:
 
 
 def report_complaint_to_user_in_reply_phase(
-    alias: Alias,
+    alias: Union[Alias, DomainDeletedAlias],
     to_address: str,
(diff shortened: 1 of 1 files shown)


app/handler/provider_complaint.py
```diff
@@ -245,16 +245,22 @@ def handle_complaint(message: Message, origin: ProviderComplaintOrigin) -> bool:
...
     origin: ProviderComplaintOrigin,
     msg_info: OriginalMessageInformation,
 ):
     capitalized_name = origin.name().capitalize()
+    mailbox_email = msg_info.mailbox_address
+    if not mailbox_email:
+        if type(alias) is Alias:
+            mailbox_email = alias.mailbox.email
+        else:
+            mailbox_email = alias.email
```
nguyenkims, line 259: should we use alias.domain.mailboxes[0] if alias is a DomainDeletedAlias here?

nguyenkims (approved):

PR #1288, opened 2022-09-15
Store session in redis if redis is enabled

acasajus:
This prevents somebody from storing an old cookie and being able to access the user data.

### DIFF
--- a/app/session.py
+++ b/app/session.py
@@ -3,6 +3,8 @@ from typing import Optional
 
 import flask
 import redis
+from flask import current_app, session
+from flask_login import logout_user
 
 try:
     import cPickle as pickle
@@ -17,12 +19,12 @@ SESSION_PREFIX = "session"
 
 
 class ServerSession(CallbackDict, SessionMixin):
-    def __init__(self, initial=None, session_Id=None):
+    def __init__(self, initial=None, session_id=None):
         def on_update(self):
             self.modified = True
 
         super(ServerSession, self).__init__(initial, on_update)
-        self.session_id = session_Id
+        self.session_id = session_id
         self.modified = False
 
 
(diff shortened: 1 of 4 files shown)


app/session.py
```diff
@@ -0,0 +1,90 @@
...
+        domain = self.get_cookie_domain(app)
+        path = self.get_cookie_path(app)
+        httponly = self.get_cookie_httponly(app)
+        secure = self.get_cookie_secure(app)
+        expires = self.get_expiration_time(app, session)
+        val = pickle.dumps(dict(session))
+        self._redis.setex(
+            name=self._get_key(session.sid),
+            value=val,
+            time=int(app.permanent_session_lifetime.total_seconds()),
```
nguyenkims, line 75: I wonder if we can store `app.permanent_session_lifetime.total_seconds()` at init time so it doesn't need to be computed each time?

app/session.py
```diff
@@ -0,0 +1,90 @@
+import uuid
```
nguyenkims, line 1: I wonder if this is is based on an existing implementation or written from scratch?

app/session.py
```diff
@@ -0,0 +1,90 @@
...
+
+SESSION_PREFIX = "session"
+
+
+class ServerSession(CallbackDict, SessionMixin):
+    def __init__(self, initial=None, sid=None):
+        def on_update(self):
+            self.modified = True
+
+        CallbackDict.__init__(self, initial, on_update)
```
nguyenkims, line 23: I compare this implementation with `SecureCookieSession` provided by default and they don't seem to use `CallbackDict`, instead they just call `super(SecureCookieSession, self).__init__(initial, on_update)`,  do you know why we need to use `CallbackDict` here?

app/session.py
```diff
@@ -0,0 +1,90 @@
...
+
+import itsdangerous
+from flask.sessions import SessionMixin, SessionInterface
+from werkzeug.datastructures import CallbackDict
+
+SESSION_PREFIX = "session"
+
+
+class ServerSession(CallbackDict, SessionMixin):
+    def __init__(self, initial=None, sid=None):
```
nguyenkims, line 19: what does `sid` stand for?

app/session.py
```diff
@@ -0,0 +1,90 @@
...
+        val = self._redis.get(self._get_key(sid))
+        if val is not None:
+            try:
+                data = pickle.loads(val)
+                return ServerSession(data, sid=sid)
+            except:
+                pass
+        return ServerSession(sid=unverified_sid)
+
+    def save_session(
```
nguyenkims, line 63: do you know where is the part a session is removed when user logs out?

app/session.py
cquintana92: I'd say is `session_id`

app/session.py
cquintana92: Based on this implementation, I'd say that if the session is not in Redis, a `ServerSession` is returned anyways, but without the `data` parameter.

app/session.py
acasajus: It's a timedelta method. Just makes an arithmetic sum and multiply. Should be fast enough.

app/session.py
acasajus: `uuid` is a standard python module... Not sure I follow..

app/session.py
acasajus: Yes. It's `session_id`

acasajus: I'll rename it to make it more clear.

app/session.py
acasajus: `SecureCookieSession` uses `CallbackDict` behind the scenes. We just skip `SecureCookieSession` because we've got a different implementation

app/session.py
acasajus: Done

app/session.py
acasajus: I've added an explicit deletion of the session data from redis just to be on the safe side

app/session.py
nguyenkims: ah this comment is rather for the whole file, I just don't know where to put it :)

Issue #1249, opened 2022-08-26
sl installed and working but no verificationmail sent


swoop124:
Hello everyone,

I installed Simple-Login, following the doc, on this github, step by step.
I'am able to send Mails from the Dockerhost via `swaks` but simple-login doesn't send the verifivation mail from registering. In the mail.info log i can only see this:
```
 /opt/sl# tail /var/log/mail.info 
Aug 26 15:21:21 SL postfix/smtpd[6412]: connect from unknown[10.0.0.4]
Aug 26 15:21:28 SL postfix/smtpd[6412]: disconnect from unknown[10.0.0.4] quit=1 commands=1
Aug 26 15:33:21 SL postfix/postfix-script[6619]: stopping the Postfix mail system
Aug 26 15:33:21 SL postfix/master[5660]: terminating on signal 15
Aug 26 15:33:21 SL postfix/postfix-script[6771]: starting the Postfix mail system
Aug 26 15:33:21 SL postfix/master[6773]: daemon started -- version 3.5.13, configuration /etc/postfix
Aug 26 15:33:26 SL postfix/postfix-script[6804]: stopping the Postfix mail system
Aug 26 15:33:26 SL postfix/master[6773]: terminating on signal 15
Aug 26 15:33:26 SL postfix/postfix-script[6956]: starting the Postfix mail system
Aug 26 15:33:26 SL postfix/master[6958]: daemon started -- version 3.5.13, configuration /etc/postfix
```
```
docker ps
CONTAINER ID   IMAGE                    COMMAND                  CREATED       STATUS          PORTS                                                                                            NAMES
6ea920bd7701   postgres:12.1            "docker-entrypoint.s…"   3 hours ago   Up 3 hours      127.0.0.1:5432->5432/tcp                                                                         sl-db
27ba5671a77f   simplelogin/app:3.4.0    "python job_runner.py"   4 hours ago   Up 17 minutes   7777/tcp                                                                                         sl-job-runner
f2983a07ea87   simplelogin/app:3.4.0    "python email_handle…"   4 hours ago   Up 16 minutes   7777/tcp, 0.0.0.0:20381->20381/tcp, :::20381->20381/tcp                                          sl-email
7067d78d466e   simplelogin/app:3.4.0    "gunicorn wsgi:app -…"   4 hours ago   Up 17 minutes   0.0.0.0:7777->7777/tcp, :::7777->7777/tcp                                                        sl-app
```
[...]

PR #1294, opened 2022-09-19
Handle unset and false variables the same way

ymage:
Some variables set with a False value are handled as True because the test is only on existence.
cf #1249


acasajus:
We cannot change this config like that. It will break already existing configs.

ymage:
What do you think about the opposite default ? Something like this :
```COLOR_LOG = os.environ.get("COLOR_LOG", 'False').lower() not in ('false', '0', 'f')```

ymage:
Can you give it a try and take a look ?

PR #1307, opened 2022-09-22
Allow RedisSessionStore to connect to sentinel

cquintana92:
This PR adds support for Redis Sentinel to the RedisSessionStore.

The code for extracting the master and replica from Sentinel has been extracted from here: https://github.com/alisaifee/limits/blob/106e74f2b239c2fe4622fb15e2fa5594c242b9fb/limits/storage/redis_sentinel.py

### DIFF
--- a/app/session.py
+++ b/app/session.py
@@ -2,7 +2,7 @@ import uuid
 from typing import Optional
 
 import flask
-import redis
+import limits.storage
 from flask import current_app, session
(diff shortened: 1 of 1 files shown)


acasajus:
I've updated the MR to use the same code as `flask_limiter` does to have both use the same code to parse the URI.

cquintana92:
The change looks good, less code to mantain!

PR #1334, opened 2022-10-05
Have custom domains set up multiple dkim records to be able to rotate keys

acasajus:
Have custom domains set up:

 - dkim._domainkey.domain.com
 - dkim02._domainkey.domain.com
 - dkim03._domainkey.domain.com

So we can rotate dkim keys if needed.

### DIFF
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -28,6 +28,6 @@ class CustomDomainValidation:
             dkim_record = get_cname_record(custom_record)
             if dkim_record != expected_record:
                 invalid_records[custom_record] = dkim_record or "empty"
-        custom_domain.verified = len(invalid_records) == 0
+        custom_domain.dkim_verified = len(invalid_records) == 0
         Session.commit()
         return invalid_records
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -6,28 +6,34 @@ from app.models import CustomDomain
 class CustomDomainValidation:
     def __init__(self, dkim_domain: str):
         self.dkim_domain = dkim_domain
-
-    def get_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
-        """
-        Get a list of dkim records to set up. It will be
-
-        """
-        return {
+        self._dkim_records = {
             (f"{key}._domainkey", f"{key}._domainkey.{self.dkim_domain}")
             for key in ("dkim", "dkim02", "dkim03")
         }
 
+    def get_dkim_records(self) -> {str: str}:
+        """
+        Get a list of dkim records to set up. It will be
+
+        """
+        return self._dkim_records
+
     def validate_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
         """
         Check if dkim records are properly set for this custom domain.
-        Returns empty list if all records are ok. Otherwise return the records that aren't properly configured
+        Returns empty list if all records are ok. Other-wise return the records that aren't properly configured
         """
         invalid_records = {}
-        for prefix, expected_record in self.get_dkim_records(custom_domain):
+        for prefix, expected_record in self.get_dkim_records():
             custom_record = f"{prefix}.{custom_domain.domain}"
             dkim_record = get_cname_record(custom_record)
             if dkim_record != expected_record:
                 invalid_records[custom_record] = dkim_record or "empty"
-        custom_domain.dkim_verified = len(invalid_records) == 0
+        if len(invalid_records) == 0:
+            custom_domain.verified = True
+        # HACK: IF DMARC is enabled do not update the dkim verification to avoid breaking it for domains that have it
+        # already set up with only one dkim record
+        elif not custom_domain.dmarc_verified:
+            custom_domain.verified = False
         Session.commit()
         return invalid_records
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -1,0 +1,33 @@
+from app.db import Session
+from app.dns_utils import get_cname_record
+from app.models import CustomDomain
+
+
+class CustomDomainValidation:
+    def __init__(self, dkim_domain: str):
+        self.dkim_domain = dkim_domain
+
+    def get_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
+        """
+        Get a list of dkim records to set up. It will be
+
+        """
+        return {
+            (f"{key}._domainkey", f"{key}._domainkey.{self.dkim_domain}")
+            for key in ("dkim", "dkim02", "dkim03")
+        }
+
+    def validate_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
+        """
+        Check if dkim records are properly set for this custom domain.
+        Returns empty list if all records are ok. Otherwise return the records that aren't properly configured
+        """
+        invalid_records = {}
+        for prefix, expected_record in self.get_dkim_records(custom_domain):
+            custom_record = f"{prefix}.{custom_domain.domain}"
+            dkim_record = get_cname_record(custom_record)
+            if dkim_record != expected_record:
+                invalid_records[custom_record] = dkim_record or "empty"
+        custom_domain.verified = len(invalid_records) == 0
+        Session.commit()
+        return invalid_records
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -18,7 +18,7 @@ class CustomDomainValidation:
         """
         return self._dkim_records
 
-    def validate_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
+    def validate_dkim_records(self, custom_domain: CustomDomain) -> dict[str, str]:
         """
         Check if dkim records are properly set for this custom domain.
         Returns empty list if all records are ok. Other-wise return the records that aren't properly configured
@@ -32,6 +32,6 @@ class CustomDomainValidation:
         # HACK: If dkim is enabled, don't disable it to give users time to update their CNAMES
         if custom_domain.dkim_verified:
             return invalid_records
-        custom_domain.verified = len(invalid_records) == 0
+        custom_domain.dkim_verified = len(invalid_records) == 0
         Session.commit()
         return invalid_records
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -29,11 +29,9 @@ class CustomDomainValidation:
             dkim_record = get_cname_record(custom_record)
             if dkim_record != expected_record:
                 invalid_records[custom_record] = dkim_record or "empty"
-        if len(invalid_records) == 0:
-            custom_domain.verified = True
-        # HACK: IF DMARC is enabled do not update the dkim verification to avoid breaking it for domains that have it
-        # already set up with only one dkim record
-        elif not custom_domain.dmarc_verified:
-            custom_domain.verified = False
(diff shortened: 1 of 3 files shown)


app/custom_domain_validation.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, dkim_domain: str):
+        self.dkim_domain = dkim_domain
+
+    def get_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
+        """
+        Get a list of dkim records to set up. It will be
+
+        """
+        return {
+            (f"{key}._domainkey", f"{key}._domainkey.{self.dkim_domain}")
```
nguyenkims, line 16: I don't see `custom_domain` is used here, is this expected?

app/custom_domain_validation.py
```diff
@@ -0,0 +1,33 @@
+from app.db import Session
+from app.dns_utils import get_cname_record
+from app.models import CustomDomain
+
+
+class CustomDomainValidation:
```
nguyenkims, line 6: I wonder if we can move this dkim checking to `dns_utils`? Maybe we don't need the whole object `CustomDomain ` and just the `domain` (`str`)?

app/custom_domain_validation.py
```diff
@@ -0,0 +1,33 @@
...
+        Check if dkim records are properly set for this custom domain.
+        Returns empty list if all records are ok. Otherwise return the records that aren't properly configured
+        """
+        invalid_records = {}
+        for prefix, expected_record in self.get_dkim_records(custom_domain):
+            custom_record = f"{prefix}.{custom_domain.domain}"
+            dkim_record = get_cname_record(custom_record)
+            if dkim_record != expected_record:
+                invalid_records[custom_record] = dkim_record or "empty"
+        custom_domain.verified = len(invalid_records) == 0
```
nguyenkims, line 31: this should rather be `custom_domain.dkim_verified` right?

app/custom_domain_validation.py
```diff
@@ -0,0 +1,33 @@
+from app.db import Session
+from app.dns_utils import get_cname_record
+from app.models import CustomDomain
+
+
+class CustomDomainValidation:
+    def __init__(self, dkim_domain: str):
+        self.dkim_domain = dkim_domain
+
+    def get_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
```
nguyenkims, line 10: this method seems to always return the same result, should we create a constant for it instead?

app/custom_domain_validation.py
nguyenkims: Something like 

```
DKIM_SELECTOR = ("dkim", "dkim02", "dkim03")
def validate_dkim_records(domain: str) -> bool:
    """
    Return true if dkim records are properly set for this domain.
    """
    invalid_records = {}
    prefix_and_records = {
            (f"{key}._domainkey", f"{key}._domainkey.{EMAIL_DOMAIN}")
            for key in DKIM_SELECTOR
        }
    for prefix, expected_record in prefix_and_records:
        custom_record = f"{prefix}.{domain}"
        dkim_record = get_cname_record(custom_record)
        if dkim_record != expected_record:
            invalid_records[custom_record] = dkim_record or "empty"
    return len(invalid_records) == 0
```

app/custom_domain_validation.py
```diff
@@ -0,0 +1,33 @@
...
+        Returns empty list if all records are ok. Otherwise return the records that aren't properly configured
+        """
+        invalid_records = {}
+        for prefix, expected_record in self.get_dkim_records(custom_domain):
+            custom_record = f"{prefix}.{custom_domain.domain}"
+            dkim_record = get_cname_record(custom_record)
+            if dkim_record != expected_record:
+                invalid_records[custom_record] = dkim_record or "empty"
+        custom_domain.verified = len(invalid_records) == 0
+        Session.commit()
```
nguyenkims, line 32: I think we shouldn't invalidate DKIM right away as if user sets a enforced DMARC, emails from their aliases will fail DMARC. We should rather just display a warning and keep the `dkim_verified` unchanged.

app/custom_domain_validation.py
acasajus: Yes, The template gives only the record to the user, not the full domain. As in 
`Set dkim._domainkey record in your domain`

app/custom_domain_validation.py
acasajus: I want to eventually move all domain validations out of the controller logic to be able to reuse it in other parts (like via api) if needed. So this is just the first one that will go into the Validator. But I'll add the refactor in another MR.

app/custom_domain_validation.py
acasajus: I'll move it to be generated on instantiation.

app/custom_domain_validation.py
acasajus: Good catch :ok_hand:  
```suggestion
        custom_domain.dkim_verified = len(invalid_records) == 0
```

app/custom_domain_validation.py
acasajus: Hm.. it could happen now couldn't it? If somebody sets DMARC but not DKIM they'll end up in the same situation. 

We could disable dkim if they don't have DMARC. Otherwise show a warning.

nguyenkims: The issue is for users who have current DKIM (with 1 CNAME) correctly set up and have DKIM enforced in their DMARC. Let's say we send an email to let them know about the change, they go the DNS page, validate their domain again and see the DKIM error but don't have time to make the change or maybe their DNS provider takes a while to propagate the change. In both cases, emails sent from their alias in the meantime won't have DKIM signature added and will be rejected because of their DMARC policy.

app/custom_domain_validation.py
nguyenkims: Ah I think we don't need OOP to have code reusuable right? Currently `dns_utils` doesn't have much dependency and can be used everywhere.

app/custom_domain_validation.py
nguyenkims: it isn't used, should it be removed from the method params?

app/custom_domain_validation.py
acasajus: Yep, This is what I propose: 
 - User has dkim enabled with one cname and dmarc
 - Validates dkim, fails because now we require 3
 - If DMARC is validated, we don't disable dkim but show a warning
 - if DMARC is not validated we disable dkim and show the warning

nguyenkims: To be on the safe side, I think we shouldn't disable DKIM if it's currently enabled. User can have their own DMARC setting (to have DMARC reporting for ex) and we don't know that. We'll leave a bit of time (1-3 months) for users to update DKIM and after that, we can use a stricter policy.

app/custom_domain_validation.py
acasajus: Removed

app/custom_domain_validation.py
acasajus: It moves us closer to removing global vars. If we have a function in a module we still need to have something global to that module and can't avoid having global variables.

app/custom_domain_validation.py
acasajus: Done

app/custom_domain_validation.py
acasajus: Ok.

app/custom_domain_validation.py
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+    def get_dkim_records(self) -> {str: str}:
+        """
+        Get a list of dkim records to set up. It will be
+
+        """
+        return self._dkim_records
+
+    def validate_dkim_records(self, custom_domain: CustomDomain) -> {str: str}:
```
nguyenkims, line 21: I think the annotation for `{str: str}` is `Dict[str,str]`

app/custom_domain_validation.py
```diff
@@ -0,0 +1,37 @@
...
+        invalid_records = {}
+        for prefix, expected_record in self.get_dkim_records():
+            custom_record = f"{prefix}.{custom_domain.domain}"
+            dkim_record = get_cname_record(custom_record)
+            if dkim_record != expected_record:
+                invalid_records[custom_record] = dkim_record or "empty"
+        # HACK: If dkim is enabled, don't disable it to give users time to update their CNAMES
+        if custom_domain.dkim_verified:
+            return invalid_records
+        custom_domain.verified = len(invalid_records) == 0
```
nguyenkims, line 35: ah this needs to be changed to `custom_domain.dkim_verified`

app/dashboard/views/domain_detail.py
```diff
@@ -47,7 +47,7 @@ def domain_detail_dns(custom_domain_id):
     spf_record = f"v=spf1 include:{EMAIL_DOMAIN} ~all"
 
     # hardcode the DKIM selector here
```
nguyenkims, line 49: maybe we can remove this comment?

app/dashboard/views/domain_detail.py
```diff
@@ -122,23 +122,17 @@ def domain_detail_dns(custom_domain_id):
                 spf_errors = get_txt_record(custom_domain.domain)
 
         elif request.form.get("form-name") == "check-dkim":
-            dkim_record = get_cname_record("dkim._domainkey." + custom_domain.domain)
-            if dkim_record == dkim_cname:
+            dkim_errors = domain_validator.validate_dkim_records(custom_domain)
+            if len(dkim_errors) == 0:
                 flash("DKIM is setup correctly.", "success")
-                custom_domain.dkim_verified = True
```
nguyenkims, line 128: We should set `custom_domain.dkim_verified` here right?

templates/dashboard/domain_detail/dns.html
```diff
@@ -285,24 +288,28 @@ <h1 class="h2">{{ custom_domain.domain }}</h1>
...
+                  The CNAME record we obtain for <em>{{ custom_record }}</em> is {{ retrieved_cname }}
+                </li>
+              {% endfor %}
+            </ul>
             {% if custom_domain.dkim_verified %}
 
               Without DKIM setup, emails you sent from your alias might end up in Spam/Junk folder.
             {% endif %}
           </div>
+          {% if custom_domain.dkim_verified %}
```
nguyenkims, line 307: this is shown even if the domain has all the DKIM records correctly set up right?

app/custom_domain_validation.py
acasajus: Argh. I've changed it twice already. Good catch!

app/custom_domain_validation.py
acasajus: Changed to `dict[str,str]`.

app/dashboard/views/domain_detail.py
acasajus: Done :+1:

app/dashboard/views/domain_detail.py
acasajus: The `DomainValidator` does it. If it's already set it doesn't change it. If it's not set it will set it to the result of the test.

templates/dashboard/domain_detail/dns.html
acasajus: It should not because it's under `if not dkim_ok`. So it should only show up if there are errors when validating dkim and it's still verified.

nguyenkims (approved):

PR #1335, opened 2022-10-06
Replace reverse alias

nguyenkims:


app/config.py
```diff
@@ -503,3 +503,14 @@ def setup_nameservers():
...
+# run over all reverse alias for an alias and replace them with sender address
+ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT = (
+    "ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT" in os.environ
+)
+
+if ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT:
+    # max number of reverse alias that can be replaced
+    MAX_NB_REVERSE_ALIAS_REPLACEMENT = int(
+        os.environ["MAX_NB_REVERSE_ALIAS_REPLACEMENT"]
+    )
```
acasajus, line 516:
```suggestion
    MAX_NB_REVERSE_ALIAS_REPLACEMENT = int(
        os.environ.get("MAX_NB_REVERSE_ALIAS_REPLACEMENT",5)
    )
```

nguyenkims: I don't put a default value here to avoid us from forgetting to set `MAX_NB_REVERSE_ALIAS_REPLACEMENT` if `ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT` is enabled actually :)

acasajus: Better to add an error then, but avoid stopping production from running if we make a config mistake.

nguyenkims: Not sure to understand, not setting it will trigger an error and stop the container from starting which is what we wanted right?

acasajus: Yeah. I mean instead of having the container from starting. Show an error but allow the container to start to avoid breaking production.

nguyenkims: prod shouldn't be broken as the existing containers will continue to run right. I just run a test with `ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT` set but not `MAX_NB_REVERSE_ALIAS_REPLACEMENT` and the existing container continues to run.

acasajus: What? How does it run? It should fail becase that should throw an exception and abort.

acasajus: If `ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT` is set but not `MAX_NB_REVERSE_ALIAS_REPLACEMENT` that throws an `KeyException` and the app won't start.

acasajus:
Added tests + batched DB queries

email_handler.py
```diff
@@ -1132,8 +1132,38 @@ def handle_reply(envelope, msg: Message, rcpt_to: str) -> (bool, str):
...
     if user.replace_reverse_alias:
         LOG.d("Replace reverse-alias %s by contact email %s", reply_email, contact)
+
         msg = replace(msg, reply_email, contact.website_email)
 
+        if config.ENABLE_ALL_REVERSE_ALIAS_REPLACEMENT:
+            start = time.time()
+            last_contact_id = 0
+            # MAX_NB_REVERSE_ALIAS_REPLACEMENT is there to limit potential attack
+            for i in range(0, config.MAX_NB_REVERSE_ALIAS_REPLACEMENT, 100):
```
nguyenkims, line 1142: Oh I thought by batch replacement, you meant something like `bulk_replace(msg, reverse_aliases, sender_addresses)` where `reverse_aliases` is a list of reverse alias and `sender_addresses` is list of sender address. This changes all occurrences of reverse aliases by sender instead of iterating through reverse alias 1 by 1. When an email body is complex, the current `replace(msg, old, new)` would take a lot of time as it does a tree traversal through the message body. 

Using `range(0, config.MAX_NB_REVERSE_ALIAS_REPLACEMENT, 100)` is actually going to increase the processing time because of database overhead. I've run a test with config.MAX_NB_REVERSE_ALIAS_REPLACEMENT set to 50000, this takes 25 seconds compared to 0.4s  where we use `.limit(config.MAX_NB_REVERSE_ALIAS_REPLACEMENT)` directly on an alias that has 25 contacts. I recommend to not fetch contacts by batch but rather than fetching the whole Contact object, only fetch what's needed which is contact reply_email and website_email.

acasajus: With the `range(0,config.MAX_...,100)` it certainly takes 4s, but with 1k it takes 0.4s in the production hosts. With the batch size set to that it allows more breathing room to the db.

I haven't found how to do `bulk_replace` in python. But we could try to do something even if it's looping the replace itself, the payload encoding will only be done once.

acasajus: I've reverted the bulk change and added that it only requests the fields it's using

PR #1337, opened 2022-10-06
do not quarantine an email if fails DMARC but has a small rspamd score

nguyenkims:


app/handler/spamd_result.py
```diff
@@ -101,6 +103,14 @@ def extract_from_headers(
                 spamd_result.set_spf_result(spf_result)
                 break
 
+        # parse the rspamd score
+        score_line = spam_entries[0]  # e.g. "default: False [2.30 / 13.00];"
+        spamd_result.rspamd_score = float(
+            score_line[(score_line.find("[") + 1) : score_line.find("]")]
+            .split("/")[0]
+            .strip()
+        )
```
acasajus, line 112:
```suggestion
        try:
            score_line = spam_entries[0]  # e.g. "default: False [2.30 / 13.00];"
            spamd_result.rspamd_score = float(
                score_line[(score_line.find("[") + 1) : score_line.find("]")]
                .split("/")[0]
                .strip()
            )
        except (IndexError, ValueError):
            LOG.d("Could not parse rspamd score")
```

app/config.py
```diff
@@ -503,3 +503,11 @@ def setup_nameservers():
...
         "Please define RECOVERY_CODE_HMAC_SECRET in your configuration with a random string at least 16 chars long"
     )
+
+# the minimum rspamd spam score above which emails that fail DMARC should be quarantined
+if "MIN_RSPAMD_SCORE_FOR_FAILED_DMARC" in os.environ:
+    MIN_RSPAMD_SCORE_FOR_FAILED_DMARC = float(
+        os.environ["MIN_RSPAMD_SCORE_FOR_FAILED_DMARC"]
+    )
+else:
+    MIN_RSPAMD_SCORE_FOR_FAILED_DMARC = None
```
acasajus, line 513: MIN_RSPAMD_SCORE_FOR_FAILED_DMARC = float(os.environ.get("MIN_RSPAMD_SCORE_FOR_FAILED_DMARC",0))

app/handler/dmarc.py
```diff
@@ -34,22 +34,46 @@ def apply_dmarc_policy_for_forward_phase(
...
+        <p style="color:red">
+            This email failed anti-phishing checks when it was received by SimpleLogin, be careful with its content.
+            More info on <a href="https://simplelogin.io/docs/getting-started/anti-phishing/">anti-phishing measure</a>
+        </p>
+        """
+
+    # do not quarantine an email if fails DMARC but has a small rspamd score
+    if (
+        config.MIN_RSPAMD_SCORE_FOR_FAILED_DMARC is not None
+        and spam_result.rspamd_score < config.MIN_RSPAMD_SCORE_FOR_FAILED_DMARC
```
acasajus, line 50:
```suggestion
        spam_result.rspamd_score < config.MIN_RSPAMD_SCORE_FOR_FAILED_DMARC
```

app/config.py
nguyenkims: This is different actually, I don't want to default `MIN_RSPAMD_SCORE_FOR_FAILED_DMARC` to 0 but None instead. The comparison with None is then used in `dmarc.py`

app/handler/dmarc.py
nguyenkims: I want to make sure ` config.MIN_RSPAMD_SCORE_FOR_FAILED_DMARC` is set and not use the implicit default value.

app/handler/spamd_result.py
nguyenkims: I think it's better to raise exception so we get notified by sentry. (fail fast principle https://en.wikipedia.org/wiki/Fail-fast

acasajus: LOG.e then and we won't fail the delivery of the message if rspamd misbehaves.

app/config.py
acasajus: If it's 0 by default and the comparison is `<` then it'will effectively be disabled won't it?

app/handler/dmarc.py
nguyenkims: I prefer this to be explicit actually.

app/config.py
nguyenkims: I prefer to be explicit and not use 0 as default value.

PR #1360, opened 2022-10-21
Feat: Added parallel limiter to prevent sqlalchemy transaction issues

acasajus:


app/parallel_limiter.py
```diff
@@ -0,0 +1,80 @@
...
+
+        return decorated
+
+
+def lock(
+    name: Optional[str] = None,
+    max_secs: int = 5,
+    only_when: Optional[Callable[..., bool]] = None,
+):
+    return __InnerLock(name, max_secs, only_when)
```
nguyenkims, line 80: Nits: I wonder if we can put the implementation directly here without using the class?

app/dashboard/views/alias_contact_manager.py
```diff
@@ -231,6 +231,7 @@ def delete_contact(alias: Alias, contact_id: int):
 
 @dashboard_bp.route("/alias_contact_manager/<int:alias_id>/", methods=["GET", "POST"])
 @login_required
+@parallel_limiter.lock(name="contact_creation")
```
nguyenkims, line 234: Do you think is there a way to test this in unit test? We can run a redis container when running test.

app/parallel_limiter.py
```diff
@@ -0,0 +1,80 @@
...
+    def acquire_lock(self, lock_name: str, lock_value: str) -> bool:
+        retries = 0
+
+        while not lock_redis.storage.set(
+            lock_name, lock_value, ex=self.max_wait_secs, nx=True
+        ):
+            retries += 1
+            if retries > self.max_wait_secs / self.wait_loop_secs:
+                return False
+            time.sleep(self.wait_loop_secs)
```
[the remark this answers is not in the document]
acasajus: I'd prefer to leave it generic. Right now it's used in alias and coupon, but we may need to use in routes where there are legitimate use-cases. We could return 429 in line 66 though.

app/dashboard/views/custom_domain.py
```diff
@@ -35,7 +35,7 @@ def custom_domain():
                 return redirect(url_for("dashboard.custom_domain"))
 
             if new_custom_domain_form.validate():
-                new_domain = new_custom_domain_form.domain.data.lower().strip()
+                new_domain = new_custom_domain_form.domain.data.ower().strip()
```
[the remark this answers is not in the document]
acasajus: Good catch. I don't know how that got changed.

app/parallel_limiter.py
acasajus: I didn't manage. We need to keep the state (name, max secs and so on) and with nested functions as soon as they were wrapped the state could not be grabbed implicitly. In any case doing it in a class makes the state easier to follow.

app/dashboard/views/alias_contact_manager.py
acasajus: Added a basic unit test.

PR #1375, opened 2022-10-27
Change: Do not sleep on exclusive zones

acasajus:


app/parallel_limiter.py
```diff
@@ -22,24 +21,16 @@ def __init__(
         lock_suffix: Optional[str] = None,
         max_wait_secs: int = 5,
```
nguyenkims, line 22: we can also remove `max_wait_secs` right?

tests/test_prarallel_limiter.py
```diff
@@ -8,6 +8,6 @@ def test_parallel_limiter(flask_app):
     user = create_new_user()
     with flask_app.test_request_context():
         login_user(user)
-        pl = _InnerLock("test", max_wait_secs=1, wait_loop_secs=0.1)
+        pl = _InnerLock("test", max_wait_secs=1)
```
nguyenkims, line 11: can we test a case that raises the TooManyRequest exception? For example by having a slow operation in the wrapped function.

app/parallel_limiter.py
acasajus: That defines what's the max time redis will hold the lock.

tests/test_prarallel_limiter.py
acasajus: I tried to but didn't manage to have two parallel threads running with pytest. It seems to make threads work serially. Either that or `time.sleep` doesn't release the GIL.

PR #1383, opened 2022-10-28
Simpler csv export

Spitfireap:
Hi,

I wanted a simpler way to export alias csv as a backup, so I simply reused the api function. I also added it to the GDPR export archive. 

I also added some tweaks for development environment :).

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,19 +1,15 @@
-import csv
 import email
 import json
 import os
 import random
 import string
 from email.message import EmailMessage
-from io import StringIO
 from typing import Optional, Dict
 
 import jinja2
 from flask import url_for
 
-from app import alias_utils
-from app.db import Session
-from app.models import User, Alias, CustomDomain, Mailbox, AliasMailbox
+from app.models import User
 from app.utils import random_string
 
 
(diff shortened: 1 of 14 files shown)


acasajus:
The aliases are already exported in the job exporter. No need to duplicate that. We need to export in json to add all the info that we have which is not contained in the CSV format. Can you please remove that part?

Also, can you add a test to make sure the routes you added work as expected?

Spitfireap:
Thanks for the review.

I deleted the csv from the export archive. As for the test, I'm not sure what you mean. I've simply reused API's function to export the csv, so the feature is [already tested here](https://github.com/simple-login/app/blob/989358af3435daef884909660b27432577c57aca/tests/api/test_import_export.py#L20), no ?

acasajus:
Yes, but not the new route you just added `/dashboard/alias_export/

tests/utils.py
```diff
@@ -41,6 +45,91 @@ def login(flask_client) -> User:
     return user
 
 
+def alias_export(flask_client, target_url):
```
acasajus, line 48: We keep this file for small utils re-used across lots of tests. Not only two. Can you move this function to a different file like `alias_test_utils.py` ?

Spitfireap: done !

acasajus:
Merged, thanks!

PR #1394, opened 2022-11-02
Feat: Re-deliver mails 

acasajus:
If we fail to contact postfix, store the mails locally and try to re-deliver them.

### DIFF
--- a/app/mail_sender.py
+++ b/app/mail_sender.py
@@ -22,6 +22,9 @@ from app.message_utils import message_to_bytes
 
 @dataclass
 class SendRequest:
+
+    SAVE_EXTENSION = "sendrequest"
+
     envelope_from: str
     envelope_to: str
     msg: Message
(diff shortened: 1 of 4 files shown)


app/mail_sender.py
```diff
@@ -180,6 +183,27 @@ def _save_request_to_unsent_dir(self, send_request: SendRequest):
 mail_sender = MailSender()
 
 
+def load_unsent_mails_from_fs_and_resend():
+    if not config.SAVE_UNSENT_DIR:
+        return
+    for filename in os.listdir(config.SAVE_UNSENT_DIR):
+        (_, extension) = os.path.splitext(filename)
+        if extension[1:] != SendRequest.SAVE_EXTENSION:
+            LOG.i(f"Skipping {filename} does not have the proper extension")
```
nguyenkims, line 192: as this should never happen, should we use a LOG.e here?

app/mail_sender.py
```diff
@@ -180,6 +183,27 @@ def _save_request_to_unsent_dir(self, send_request: SendRequest):
...
+            continue
+        full_file_path = os.path.join(config.SAVE_UNSENT_DIR, filename)
+        if not os.path.isfile(full_file_path):
+            LOG.i(f"Skipping {filename} as it's not a file")
+            continue
+        LOG.i(f"Trying to re-deliver email {filename}")
+        try:
+            send_request = SendRequest.load_from_file(full_file_path)
+        except Exception as e:
+            LOG.error(f"Could not load file {filename}: {e}")
```
nguyenkims, line 202: we can use LOG.exception here, it'll automatically include the exception

PR #1407, opened 2022-11-09
remove the code which is never called

nguyenkims:


.github/workflows/main.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.9", "3.10"]
```
acasajus, line 35: We still run the ci in python3.9 because updating macs to 3.10 was complicated. If that's no longer true and we can update to 3.10 we also need to update the Dockerfile to use 3.10 in the containers.

nguyenkims: I just upgrade to Python 3.10 on Mac and everything seems to work well so far. We are now using python 3.10 in our dockerfile so I think it's safe to remove python 3.9 here.

acasajus: Let's do this then! python 3.10 from now on :+1:

PR #1458, opened 2022-11-30
Use canonical email when registering users

acasajus:
For old users allow them to login with non canonicalized emails.
New registrations will use canonical emails always.


cquintana92 (approved):

julitapy:
This PR makes it impossible to register a new account, if you have a dot in the username part of you email address and your email provider does care about dots (basically all except gmail). Thanks for merging it just in time for me to realize that I don't want to spend $30/year for such poorly maintained service.

acasajus:
Hello, we had to deploy a quick fix due to some abusers generating many accounts with one mail account. We will deploy an improvement next week.

PR #1459, opened 2022-11-30
Keep dirty email after registration

acasajus:


app/models.py
```diff
@@ -320,6 +320,7 @@ class User(Base, ModelMixin, UserMixin, PasswordOracle):
     FLAG_FREE_OLD_ALIAS_LIMIT = 1 << 2
 
     email = sa.Column(sa.String(256), unique=True, nullable=False)
+    dirty_email = sa.Column(sa.String(256), unique=False, nullable=True)
```
nguyenkims, line 323: It's probably too late now but I'd prefer using other word rather than "dirty" :), for ex "original_email" or "non_canonical_email"

PR #1464, opened 2022-12-02
Update docs to the same port as the reset script + remove pre-commit pylint

acasajus:


scripts/reset_local_db.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-export DB_URI=postgresql://myuser:mypassword@localhost:35432/simplelogin
+export DB_URI=postgresql://myuser:mypassword@localhost:15432/simplelogin
```
nguyenkims, line 3: just curious, why do we need to change the port here?

PR #1503, opened 2022-12-21
Set the proper link for unsub newsletter

acasajus:


app/newsletter_utils.py
```diff
@@ -27,13 +27,15 @@ def send_newsletter_to_user(newsletter, user) -> (bool, str):
             comm_alias_id = comm_alias.id
 
         unsubscribe_oneclick = unsubscribe_link
-        if via_email:
+        if via_email and comm_alias_id > -1:
```
nguyenkims, line 30: In the last newsletter, the unsubscribe link is something like https://app.simplelogin.io/dashboard/unsubscribe/3210948 so user still needs to log in in order to unsubscribe, do you know if this is expected?

acasajus: Yes, the unsub link requires auth because we need to make sure that the alias belongs to the user. I've got a plan to use the signed unsub payload with another path that wouldn't require auth. But it's not in this MR.

PR #1515, opened 2023-01-08
Update docu for /api/alias/random/new

adrian-schnell:
the param `mode` has to be passed in the query


nguyenkims (approved):
Thanks for improving the doc!

PR #1524, opened 2023-01-10
Fix: Use timed signers to avoid leaving permanent links in mailbox change and verification

acasajus:


cquintana92 (approved):

app/dashboard/views/mailbox_detail.py
```diff
@@ -262,11 +262,11 @@ def cancel_mailbox_change_route(mailbox_id):
 
 @dashboard_bp.route("/mailbox/confirm_change")
 def mailbox_confirm_change_route():
-    s = Signer(MAILBOX_SECRET)
+    s = TimestampSigner(MAILBOX_SECRET)
     signed_mailbox_id = request.args.get("mailbox_id")
 
     try:
-        mailbox_id = int(s.unsign(signed_mailbox_id))
+        mailbox_id = int(s.unsign(signed_mailbox_id, max_age=900))
```
nguyenkims, line 269: should we display an error when the link is unexpired so user can retry?

PR #1457, opened 2022-11-28
Transfer alias when deleting a mailbox

mlec1:
This MR tries to address the following [ticket](https://github.com/simple-login/app/discussions/410).

It give the possibility to transfer aliases to a new mailbox when a mailbox is deleted.

Let me know if I have missed something.

PR #1534, opened 2023-01-13
Tranfer aliases to a new mailbox when deleting mailboxes

acasajus:
Improvement on https://github.com/simple-login/app/pull/1457 
I had to do it in a different MR because I can't push to @mlec1 's branch

### DIFF
--- a/app/api/views/mailbox.py
+++ b/app/api/views/mailbox.py
@@ -16,7 +16,7 @@ from app.email_utils import (
     is_valid_email,
 )
 from app.log import LOG
-from app.models import AliasMailbox, Mailbox, Job
+from app.models import Mailbox, Job
 from app.utils import sanitize_email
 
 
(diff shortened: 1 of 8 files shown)


mlec1:
Hello, 

Sorry for my late reaction and that you had to take over this.

Just a small comment concerning the `npm ci` in the Dockerfile, because there were the question about it on my branch:
- `npm install` may write to **package-lock.json** if a dependency of **package.json** is missing or a newer version is available
- `npm ci` will install all the packages directly from the **package-lock.json** file

Here are some better explanation: https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

I had the issue once in another project where the **npm install** was installing newer dependencies which were breaking the project. After searching about the issue, we discovered the **npm ci** command line, and we went for it.

I thought it might be a safer install method for you too, but I let you the decision to keep it or not, I am far from being a npm expert

acasajus:
I've added your fix to the Dockerfile in a different MR https://github.com/simple-login/app/pull/1543 . Thanks!

cquintana92 (approved):

app/api/views/mailbox.py
```diff
@@ -78,24 +78,52 @@ def delete_mailbox(mailbox_id):
...
+        (optional) transfer_aliases_to: in body. Id of the new mailbox for the aliases.
+                                        If omitted or the value is set to -1,
+                                        the aliases of the mailbox will be deleted too.
     Output:
         200 if deleted successfully
 
     """
     user = g.user
-    mailbox = Mailbox.get(mailbox_id)
+    mailbox = Mailbox.get(id=mailbox_id)
```
nguyenkims, line 89: Nits: we can remove the `id=` part here

app/dashboard/views/mailbox.py
```diff
@@ -40,28 +47,53 @@ def mailbox_route():
...
             if mailbox.id == current_user.default_mailbox_id:
                 flash("You cannot delete default mailbox", "error")
                 return redirect(url_for("dashboard.mailbox_route"))
 
+            transfer_mailbox_id = delete_mailbox_form.transfer_mailbox_id.data
+            if transfer_mailbox_id and transfer_mailbox_id > 0:
+                transfer_mailbox = Mailbox.get(transfer_mailbox_id)
+
+                if not transfer_mailbox or transfer_mailbox.user_id != current_user.id:
+                    flash("You must transfer the aliases to a mailbox you own.")
```
nguyenkims, line 72: `flash` requires a second argument for the message level (warning, info, etc)

app/dashboard/views/mailbox.py
```diff
@@ -112,12 +147,12 @@ def mailbox_route():
                 elif not email_can_be_used_as_mailbox(mailbox_email):
                     flash(f"You cannot use {mailbox_email}.", "error")
                 else:
-                    new_mailbox = Mailbox.create(
+                    transfer_mailbox = Mailbox.create(
```
nguyenkims, line 150: We shouldn't rename `new_mailbox` to `transfer_mailbox` here right?

job_runner.py
```diff
@@ -124,6 +124,58 @@ def welcome_proton(user):
...
+            alias_transferred_to = transfer_mailbox.email
+
+            for alias in mailbox.aliases:
+                if alias.mailbox_id == mailbox.id:
+                    alias.mailbox_id = transfer_mailbox.id
+                    if transfer_mailbox in alias._mailboxes:
+                        alias._mailboxes.remove(transfer_mailbox)
+                else:
+                    alias._mailboxes.remove(mailbox)
+                    if transfer_mailbox not in alias._mailboxes:
```
nguyenkims, line 147: can that happen that `alias.mailbox_id = transfer_mailbox_id`?

app/api/views/mailbox.py
mlec1: Done here https://github.com/simple-login/app/pull/1547

app/dashboard/views/mailbox.py
mlec1: Done here https://github.com/simple-login/app/pull/1547

app/dashboard/views/mailbox.py
mlec1: I agree. Done here https://github.com/simple-login/app/pull/1547

job_runner.py
mlec1: I am not sure to understand your question:

If the mailbox we want to delete, is the "main" mailbox of the alias, we set the "main" mailbox to the **transfer_mailbox** and the value is saved with Session.commit()

acasajus: It's validated in the request before creating the job that they aren't the same.

mlec1: Ha ok, I understand the question now. Yeah, I agree with @acasajus.

PR #1552, opened 2023-01-23
Fix delete alias mlec

mlec1:
Fix linting to pass test 🎨


nguyenkims (approved):
Thanks for fixing the formatting!

PR #1609, opened 2023-02-27
Add sender to PGP message body

guzlewski:
I'm using SimpleLogin with PGP encryption enabled for a while, and I'm struggling with one issue. As FROM header is not encrypted, I have selected option **No Name** in **Sender Address Format** for better security. However, I'm losing information who sent these emails, I think the best option is to modify the header to something like this:
 
Forwarded by SimpleLogin to <simplelogin_recipient> **from <sender_addess>** with \<subject> as subject

This PR implements this feature :)


acasajus:
@nguyenkims I think it's a good idea.

nguyenkims (approved):
LGTM!

PR #1623, opened 2023-03-08
Beautify sender formatting in html template

guzlewski:
To be consistent with subject formatting, it should bold instead of in quotes.


guzlewski:
Changes from my last PR don't work as expected, sender name is present but email is missing. I don't have setup to run tests and couldn't figure out why this is happening. Is it due to some anonymisation or FROM header is medified before entering forward_email_to_mailbox function? I'd be over the moon if you could look into it.

Currenct behaviour:
image
Expected behaviour:
image

PR #1624, opened 2023-03-09
Break using an alias as a mailbox loop in the email_handler.py

acasajus:
We don't allow to use aliases as mailbox. If that happens break the loop in the email_handler

### DIFF
--- a/email_handler.py
+++ b/email_handler.py
@@ -701,6 +701,26 @@ def handle_forward(envelope, msg: Message, rcpt_to: str) -> List[Tuple[bool, str
                 )
                 mailbox.verified = False
                 Session.commit()
+                mailbox_url = f"{URL}/dashboard/mailbox/{mailbox.id}/"
+                send_email_with_rate_control(
+                    user,
+                    ALERT_MAILBOX_IS_ALIAS,
+                    user.email,
+                    f"Your mailbox {mailbox.email} is an alias",
+                    render(
+                        "transactional/mailbox-invalid.txt.jinja2",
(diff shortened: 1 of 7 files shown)


email_handler.py
```diff
@@ -693,12 +693,48 @@ def handle_forward(envelope, msg: Message, rcpt_to: str) -> List[Tuple[bool, str
...
+                    ),
+                    render(
+                        "transactional/mailbox-invalid.html",
+                        mailbox=mailbox,
+                        mailbox_url=mailbox_url,
+                        alias=alias,
+                    ),
+                    max_nb_alert=1,
+                )
+                ret.append((False, status.E525))
```
nguyenkims, line 724: Nits: maybe we can use `continue` here so we don't need the `else` below which reduce the number of indentation 😅

email_handler.py
```diff
@@ -1000,6 +1036,13 @@ def handle_reply(envelope, msg: Message, rcpt_to: str) -> (bool, str):
         LOG.w(f"No contact with {reply_email} as reverse alias")
         return False, status.E502
 
+    contact_alias_check = Alias.get_by(email=contact.website_email)
```
nguyenkims, line 1039: I think we allow a contact to be an alias. For ex, a user can send an email from his alias to another user's alias.

acasajus: You're right. I've removed this condition

email_handler.py
acasajus: I prefer explicit conditions over loop updates but fair enough. Added the continue and reduced the indent

PR #1625, opened 2023-03-10
Feat: Use only sfw words with a number suffix

acasajus:


nguyenkims:
@acasajus Hey I only see the word in suffix but not numbers, is this expected?
image

acasajus:
I thought we only wanted to update random aliases. But it does make sense to also add it for the custom ones since we're reducing the number of words. Added to the MR

nguyenkims (approved):

PR #1658, opened 2023-03-23
Fix drag and drop to upload PGP public key not working on Firefox and Chromium (but working on Safari)

D-Bao:
The drag and drop of a PGP public key file only worked on Safari. I added the missing event listeners to also make it work on Firefox and Chromium-based browsers.

For the client side validation of the PGP key file I also tried to verify the file based on its extension, because file.type can give different results based on different browsers' implementations.
For example for a file with the extension `.asc`:
- file.type returns "text/plain" on Safari
- file.type returns "" (empty string) on Firefox and Chromium

So if I didn't change this code:
https://github.com/simple-login/app/blob/66039c526bfeeb4ca60c834fe82337cff6eb9df7/static/js/utils/drag-drop-into-text.js#L11-L12
a user could upload an `.asc` file properly in Safari but would get an error notification on Firefox/Chromium.

Bonus: after a short discussion with @nguyenkims about the UI, I also slightly changed the style of the textarea to make it more explicit that the user can drag and drop their PGP key file there.

### DIFF
--- a/static/js/utils/drag-drop-into-text.js
+++ b/static/js/utils/drag-drop-into-text.js
@@ -17,6 +17,7 @@ function enableDragDropForPGPKeys(inputID) {
             reader.onloadend = onFileLoaded;
             reader.readAsBinaryString(file);
         }
+        dropArea.classList.remove("dashed-outline");
     }
 
     function onFileLoaded(event) {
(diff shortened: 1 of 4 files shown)


nguyenkims (approved):
LGTM!

PR #1665, opened 2023-03-24
Create Partner only domains

acasajus:
Domains that will only be shown to users that come from a partner and have enabled that flag.

Still WIP


app/models.py
```diff
@@ -2763,6 +2774,31 @@ def render(template_name, **kwargs) -> str:
         )
 
 
+class Partner(Base, ModelMixin):
```
nguyenkims, line 2777: i wonder if we can keep this class at the same location to reduce code diff?

app/models.py
```diff
@@ -967,13 +967,24 @@ def available_sl_domains(self) -> [str]:
         """
         return [sl_domain.domain for sl_domain in self.get_sl_domains()]
 
-    def get_sl_domains(self) -> List["SLDomain"]:
-        query = SLDomain.filter_by(hidden=False).order_by(SLDomain.order)
-
-        if self.is_premium():
-            return query.all()
+    def get_sl_domains(self, domain_name: Optional[str] = None) -> list["SLDomain"]:
```
nguyenkims, line 970: `list` should be `List` I think

app/models.py
acasajus: It has to be on top of SLDomain so the type resolver has seen it before it gets to SLDomain. Otherwise the static analizer gives an error.

app/models.py
acasajus: In python 3.10 the standard types are included like `list`, `tuple`... https://docs.python.org/3.10/library/typing.html

nguyenkims: ah ok, didn't know that!

PR #1680, opened 2023-04-05
Redesign new pricing page

D-Bao:
This new pricing page has been redesigned from scratch and also includes a FAQ section. The text of the list of features and FAQ questions/answers are taken from [https://simplelogin.io/pricing](https://simplelogin.io/pricing).
Feel free to edit the text.

Also for the formatting I ran the command `poetry run djlint --reformat templates` but it removed the indentation, I don't know if that could be an issue.

### DIFF
--- a/templates/dashboard/pricing.html
+++ b/templates/dashboard/pricing.html
@@ -4,211 +4,392 @@
 {% block title %}Pricing{% endblock %}
 {% block head %}
 
-  <script src="https://cdn.paddle.com/paddle/paddle.js"></script>
-  <script>
-    if (window.Paddle === undefined) {
-      console.log("cannot load Paddle from CDN");
-      document.write('<script src="/static/vendor/paddle.js"><\/script>')
-    }
+<script src="https://cdn.paddle.com/paddle/paddle.js"></script>
+<script>
+  if (window.Paddle === undefined) {
+    console.log("cannot load Paddle from CDN");
+    document.write('<script src="/static/vendor/paddle.js"><\/script>')
+  }
 </script>
 <style type="text/css">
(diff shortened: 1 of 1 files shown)


acasajus:
Hey, checking on the linting. If you replace in the header block

```
      document.write('<script src="/static/vendor/paddle.js"><\/script>')
```

for 

```  
document.write('<script src="/static/vendor/paddle.js"></script>'); 
```

linting with djlint will work again

acasajus:
Alternatively we can use
```js
    if (window.Paddle === undefined) {
      console.log("cannot load Paddle from CDN");
      var element = document.createElement('script');
      element.src = '/static/vendor/paddle.js';
      document.body.appendChild(element);
}
```

D-Bao:
Thanks for identifying the linting issue. I tried your alternative (with document.head instead of document.body because it said document.body is not defined) however it created an issue with the script at the bottom of the page:

```js
<script type="text/javascript">
  Paddle.Setup({vendor: {{ PADDLE_VENDOR_ID }}});
```
saying `Paddle is not defined`, even if the previous script was successfully added to the document head, probably because it was added too late in the scripts loading queue.

So instead I simply wrote this (may be a little ugly xD):
```js
document.write('<' + 'script src="/static/vendor/paddle.js"><\/script' + '>');
```
which fixed the djlint issue.

acasajus:
Can you try
```
document.prepend(element)
```
instead of `appendChild` to have the element as the first entry of the body?

D-Bao:
I just tried but it also shows the same error

acasajus:
LGTM Ugly but it works :)

acasajus (approved):
@nguyenkims Once you approve of the changes we can merge.

nguyenkims (approved):
LGTM!

PR #1685, opened 2023-04-07
Replace save/update buttons with an auto save feature

D-Bao:
The save/update buttons in the aliases page have been deleted and replaced by an auto save feature for:

- the alias description text area, automatically saved when the field loses focus
- the mailbox select option, automatically saved when checking an option
- the display name input, automatically saved when losing focus or pressing Enter

Just in case, I added the hint message "(automatically saved when you click outside the field)", displayed only when a field is in focus, but if you find the message unnecessary I can delete it.

I also improved the notification "Description|Mailbox|Display Name saved"  to "Description|Mailbox|Display Name saved **for {alias.email}**".

### DIFF
--- a/static/js/index.js
+++ b/static/js/index.js
@@ -155,10 +155,8 @@ $(".pin-alias").change(async function () {
   }
 });
 
-$(".save-note").on("click", async function () {
-  let oldValue;
-  let aliasId = $(this).data("alias");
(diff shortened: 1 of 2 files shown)


nguyenkims (approved):
LGTM! Adding "automatically saved when you click outside the field" is a good idea 👍

PR #1689, opened 2023-04-11
Alias domain as contact domain

acasajus:
Use the alias domain as the contact domain also.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1293,15 +1293,24 @@ class OauthToken(Base, ModelMixin):
         return self.expired < arrow.now()
 
 
-def generate_email(
+def available_sl_email(email: str, check_trash: bool = True) -> bool:
+    if Alias.get_by(email=email) or Contact.get_by(reply_email=email):
+        return False
+    if check_trash and DeletedAlias.get_by(email=email):
+        return False
+    return True
+
+
+def generate_random_alias_email(
     scheme: int = AliasGeneratorEnum.word.value,
     in_hex: bool = False,
-    alias_domain=config.FIRST_ALIAS_DOMAIN,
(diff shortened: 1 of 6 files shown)


app/models.py
```diff
@@ -1581,6 +1593,10 @@ def mailbox_email(self):
         else:
             return self.user.email
 
+    def get_domain(self) -> str:
+        splitPos = self.email.find("@")
```
nguyenkims, line 1597: nits: should we use underscore as variable name here?

nguyenkims: Maybe we can use this method `get_email_domain_part`

nguyenkims (commented):
I think we need to change the VERP to use the alias domain as well to make sure mail from and header from (DMARC) align.

app/models.py
acasajus: Replace usages with `get_email_domain_part`

acasajus:
Updated the VERP domain to align DMARC

app/models.py
```diff
@@ -1293,16 +1293,28 @@ def is_expired(self):
         return self.expired < arrow.now()
 
 
-def generate_email(
+def available_sl_email(email: str, check_trash: bool = True) -> bool:
```
nguyenkims, line 1296: It seems that no caller overrides the `check_trash` param, can we remove it?

nguyenkims (approved):
LGTM, I guess we need to run on it on staging to make sure there's no regression :)

app/models.py
acasajus: I've removed it. I used in some code that's no longer there.

acasajus:
It's already running in staging.

PR #1691, opened 2023-04-14
New alias detail page

D-Bao:
This new alias detail page shows the information for 1 alias, the URL of the page being `/dashboard/aliases/{alias_id}`. It replaces the usage of `dashboard/?highlight_alias_id={alias_id}` for some URL redirections like from the alias transfer page and the refused email page. We may also combine this new page with the alias log + contact manager pages in the future.

### DIFF
--- a/app/dashboard/views/alias_detail.py
+++ b/app/dashboard/views/alias_detail.py
@@ -1,0 +1,60 @@
+from flask import render_template, flash, redirect, url_for, request
+from flask_login import login_required, current_user
+
+from app import alias_utils
+from app.api.serializer import get_alias_info_v3
+from app.dashboard.base import dashboard_bp
+from app.db import Session
+from app.log import LOG
+from app.utils import CSRFValidationForm
+
+
+@dashboard_bp.route("/aliases/<int:alias_id>", methods=["GET", "POST"])
+@login_required
+def aliases(alias_id):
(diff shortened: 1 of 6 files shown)


app/dashboard/views/alias_detail.py
```diff
@@ -0,0 +1,60 @@
...
+    mailboxes = current_user.mailboxes()
+
+    csrf_form = CSRFValidationForm()
+
+    if request.method == "POST":
+        if not csrf_form.validate():
+            flash("Invalid request", "warning")
+            return redirect(request.url)
+        if request.form.get("form-name") in ("delete-alias", "disable-alias"):
+            if not alias or alias.user_id != current_user.id:
```
nguyenkims, line 37: this condition has been checked before so we can remove it here I think

nguyenkims:
Hey some feedbacks about the layout

CleanShot 2023-04-16 at 22 23 10@2x

I think when alias log and contact manager can be added to the right side panel, similarly to the domain detail page:
image

nguyenkims:
@D-Bao 
CleanShot 2023-04-22 at 21 17 43@2x

D-Bao:
Actually I planned to remove the text "0 forwarded... See all" once alias log is added to the right side panel.
But it's true for the moment it creates redundant info, do you want me to temporarily put it above "Created 3 months ago" and remove "No emails received..."?

PR #1697, opened 2023-04-17
Add option to show/hide stats in aliases page

D-Bao:
The stats for aliases (amount, replies/sent etc...) will be hidden by default. They can be shown/hidden by toggling a button and this option is persistent between page refreshes (similar to how the Filters button works).


nguyenkims:
@D-Bao the UI/UX looks good to me. However I think the stats should only be loaded when user open the stats so by default  stats shouldn't be loaded. This probably requires creating an endpoint to expose the stats info so the frontend can fetch and adding a loading indicator while waiting for the stats. We can do that in another PR to minimize the change.

PR #1702, opened 2023-04-20
fix the E501 check

nguyenkims:


email_handler.py
```diff
@@ -1020,12 +1020,13 @@ def handle_reply(envelope, msg: Message, rcpt_to: str) -> (bool, str):
...
-    if not reply_email.endswith(EMAIL_DOMAIN) or not SLDomain.get_by(
-        domain=reply_domain
-    ):
-        LOG.w(f"Reply email {reply_email} has wrong domain")
-        return False, status.E501
+
+    # reply_email must end with EMAIL_DOMAIN or a domain that can be used as reverse alias domain
+    if not reply_email.endswith(EMAIL_DOMAIN):
+        sl_domain: SLDomain = SLDomain.get_by(domain=reply_domain)
+        if sl_domain is None or not sl_domain.use_as_reverse_alias:
```
acasajus, line 1027: We should not check the `use_as_reverse_alias` if at some point we have a domain that was used to have reverse alias and we turn that off. All contacts already created will fail.

nguyenkims: I think once we set the flag on a domain, we shouldn't turn it off.

acasajus: Just in case. Better be safe than sorry. Once the contact is created we should always accept it. I can do a fix.

PR #1704, opened 2023-04-21
Fix: Allow contacts created with a domain to be delivered

acasajus:
even if the domain cannot be used any more for contact creation


nguyenkims (approved):
👌

PR #1708, opened 2023-04-26
feat(alias-used-on): Add websites where an alias is used ✨

mlec1:
This MR try to address the following this feature request https://github.com/simple-login/app/discussions/469

The JavaScript part may be a bit tricky, but I tried to adapt in order to keep the already existing package (here the multi-select one). I think the consistency might be better in terms of UI. 

If you find a cleaner way, feel free to drop a comment.

### DIFF
--- a/app/api/serializer.py
+++ b/app/api/serializer.py
@@ -15,6 +15,7 @@ from app.models import (
     AliasMailbox,
     CustomDomain,
     User,
+    AliasUsedOn
 )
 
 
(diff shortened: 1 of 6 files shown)


nguyenkims:
Thanks for making the PR.
It seems that all aliases have the same websites as you can see in the below screenshot:
CleanShot 2023-05-06 at 22 06 05@2x

And we have a display error when there's no websites used:
CleanShot 2023-05-06 at 22 06 49@2x

I've thought again about the feature request and am wondering maybe this isn't relevant any more because: 
a) now that in the browser extension we automatically add "Used on website.com", we can know easily where an alias is created
b) The AliasUsedOn is used to inform users that an alias has been created on a website before but I don't see the utility of this. 

I think what we can do instead is to 

- update the alias note to something like "Created on website1, website2, etc" based on AliasUsedOn data for old aliases that were created before "Used on website.com" was added
- remove the AliasUsedOn table

What do you think?

PR #1717, opened 2023-05-03
Documentation clarification

fm:
- Fix a markdown syntax error that prevented a section of the API docs from formatting correctly
- `spam` is an uncountable noun in english - `spams` should not be used, but rather `spam emails` as in 
_"I get many spam emails"_ instead of _"I get many spams"_ or _"They send many spam emails"_ instead of _"They send many spams"_


nguyenkims (approved):
Thanks for making this PR!

PR #1719, opened 2023-05-05
(WIP) new Aliases SPA using Vue.js

D-Bao:
This PR creates a new Aliases page which uses the latest [API endpoints](https://github.com/simple-login/app/blob/master/docs/api.md) to do operations on aliases such as updating, deleting, disabling, creating, searching etc... without having to refresh the page like the current Aliases page does.

### DIFF
--- a/static/js/dashboard-spa.js
+++ b/static/js/dashboard-spa.js
@@ -1,0 +1,416 @@
+new Vue({
+  el: '#dashboard-app',
+  delimiters: ["[[", "]]"], // necessary to avoid conflict with jinja
+  data: {
+    showFilter: false,
+    showStats: false,
+
+    mailboxes: [],
+
+    // variables for aliases list
+    isFetchingAlias: true,
+    aliasesArray: [], // array of existing alias
+    aliasesArrayOfNextPage: [], // to know there is a next page if not empty
+    page: 0,
+    isLoadingMoreAliases: false,
+    searchString: "",
+    filter: "", // TODO add more filters and also sorting when backend API is ready
+  },
+  computed: {
+    isLastPage: function () {
+      return this.aliasesArrayOfNextPage.length === 0;
+    },
+  },
+  async mounted() {
+
+    if (store.get("showFilter")) {
+      this.showFilter = true;
+    }
+
+    if (store.get("showStats")) {
+      this.showStats = true;
+    }
+
+    await this.loadInitialData();
+
+  },
+  methods: {
+    // initialize mailboxes and aliases
+    async loadInitialData() {
+      await this.loadMailboxes();
+      await this.loadAliases();
+    },
+
+    async loadMailboxes() {
+      try {
+        const res = await fetch("/api/mailboxes");
+        if (res.ok) {
+          const result = await res.json();
+          this.mailboxes = result.mailboxes;
+        } else {
+          throw new Error("Could not load mailboxes");
+        }
+      } catch (e) {
+        toastr.error("Sorry for the inconvenience! Could you try refreshing the page? ", "Could not load mailboxes");
+      }
+    },
+
+    async loadAliases() {
+      this.aliasesArray = [];
+      this.page = 0;
+
+      this.aliasesArray = await this.fetchAlias(this.page, this.searchString);
+      this.aliasesArrayOfNextPage = await this.fetchAlias(this.page + 1, this.searchString);
+
+      // use jquery multiple select plugin after Vue has rendered the aliases in the DOM
+      this.$nextTick(() => {
+        $('.mailbox-select').multipleSelect();
+        $('.mailbox-select').removeClass('mailbox-select');
+      })
+    },
+
+    async fetchAlias(page, query) {
+      this.isFetchingAlias = true;
+      try {
+        const res = await fetch(`/api/v2/aliases?page_id=${page}&${this.filter}`, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          ...(query && { body: JSON.stringify({ query }) }),
+        });
+        if (res.ok) {
+          const result = await res.json();
+          this.isFetchingAlias = false;
+          return result.aliases;
+        } else {
+          throw new Error("Aliases could not be loaded");
+        }
+      } catch (e) {
+        toastr.error("Sorry for the inconvenience! Could you try refreshing the page? ", "Aliases could not be loaded");
+        this.isFetchingAlias = false;
+        return [];
+      }
+    },
+
+    async toggleFilter() {
+      this.showFilter = !this.showFilter;
+      store.set('showFilter', this.showFilter);
+    },
+
+    async toggleStats() {
+      this.showStats = !this.showStats;
+      store.set('showStats', this.showStats);
+    },
+
+    async toggleAlias(alias) {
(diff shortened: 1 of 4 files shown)


static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+    aliasPrefixInput: "",
+    aliasPrefixError: "",
+    aliasSuffixes: [],
+    aliasSelectedSignedSuffix: "",
+    aliasNoteInput: "",
+    defaultMailboxId: "",
+
+    // variables for aliases list
+    isFetchingAlias: true,
+    aliasesArray: [], // array of existing alias
```
nguyenkims, line 26: can we rename this to just `aliases` for short?

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+    },
+
+    async loadAliasOptions() {
+      this.isLoading = true;
+      try {
+        const res = await fetch("/api/v5/alias/options");
+        if (res.ok) {
+          const aliasOptions = await res.json();
+          this.aliasSuffixes = aliasOptions.suffixes;
+          this.aliasSelectedSignedSuffix = this.aliasSuffixes[0].signed_suffix;
```
nguyenkims, line 83: as this is expired after a short time, I think it's better to only load alias options when user clicks on the "new custom alias" button.

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+    aliasNoteInput: "",
+    defaultMailboxId: "",
+
+    // variables for aliases list
+    isFetchingAlias: true,
+    aliasesArray: [], // array of existing alias
+    aliasesArrayOfNextPage: [], // to know there is a next page if not empty
+    page: 0,
+    isLoadingMoreAliases: false,
+    searchString: "",
```
nguyenkims, line 30: I think we can name this `query` :)

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+
+      // use jquery multiple select plugin after Vue has rendered the aliases in the DOM
+      this.$nextTick(() => {
+        $('.mailbox-select').multipleSelect();
+        $('.mailbox-select').removeClass('mailbox-select');
+      });
+    },
+
+    async fetchAlias(page, query) {
+      this.isFetchingAlias = true;
```
nguyenkims, line 109: should we show a loading indicator when isFetchingAlias = true?

templates/dashboard/dashboard_spa.html
```diff
@@ -0,0 +1,318 @@
...
+          <div v-else-if="alias.latest_activity" class="small-text mb-4">
+            [[ alias.latest_activity.contact.email ]]
+            <span v-if="alias.latest_activity.action === 'bounched'">
+              <i class="fa fa-exclamation-triangle mr-2"></i> Cannot be forwarded to your mailbox.
+            </span>
+            <span v-if="alias.latest_activity.action === 'block'">
+              <i class="fa fa-ban mr-2"></i> Blocked.
+            </span>
+            <!-- TODO improve date formatting -->
+            Created [[ alias.creation_date ]].
```
nguyenkims, line 171: I think we can use this code snippet https://stackoverflow.com/a/6109105 here to have a relative date time.

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+        if (res.ok) {
+          alias.pinned = !alias.pinned;
+          if (alias.pinned) {
+            // make alias appear at the top of the alias list
+            const index = this.aliasesArray.findIndex((a) => a.id === alias.id);
+            this.aliasesArray.splice(index, 1);
+            this.aliasesArray.unshift(alias);
+            toastr.success(`${alias.email} is pinned`);
+          } else {
+            // unpin: make alias appear at the bottom of the alias list
```
nguyenkims, line 254: Though moving a pinned alias to the top is a good idea, handling the reverse case when an alias is unpinned is much harder. I think let's just not handle this for now and leave the item where it is :)

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+    aliasPrefixError: "",
+    aliasSuffixes: [],
+    aliasSelectedSignedSuffix: "",
+    aliasNoteInput: "",
+    defaultMailboxId: "",
+
+    // variables for aliases list
+    isFetchingAlias: true,
+    aliasesArray: [], // array of existing alias
+    aliasesArrayOfNextPage: [], // to know there is a next page if not empty
```
nguyenkims, line 27: I think this can be renamed to nextPageAliases :)

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+      // use jquery multiple select plugin after Vue has rendered the aliases in the DOM
+      this.$nextTick(() => {
+        $('.mailbox-select').multipleSelect();
+        $('.mailbox-select').removeClass('mailbox-select');
+      });
+
+      this.isLoadingMoreAliases = false;
+    },
+
+    resetFilter() {
```
nguyenkims, line 414: We need to empty the aliases and load aliases again when resetting filter.

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+      try {
+        const res = await fetch("/api/v3/alias/custom/new", {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify({
+            alias_prefix: this.aliasPrefixInput,
+            signed_suffix: this.aliasSelectedSignedSuffix,
+            mailbox_ids: [this.defaultMailboxId],
```
nguyenkims, line 442: should we allow user to select multiple mailboxes in the create alias modal?

nguyenkims: we can also just add a TODO to not forget :)

static/js/dashboard-spa.js
```diff
@@ -0,0 +1,538 @@
...
+            note: this.aliasNoteInput,
+          }),
+        });
+
+        if (res.ok) {
+          const alias = await res.json();
+          this.aliasesArray.unshift(alias);
```
nguyenkims, line 449: we can also scroll to this new alias as user can be way below the viewport by using `scrollIntoView()`

nguyenkims (commented):
Hey the new page seems promising 👍. I just added some comments, some don't have to be done in this PR. And can we try to have the same layout as the current one? Apart from the filter that we can handle later, I notice some differences: 
-  there's no "more" button in alias card
- pin should be a toggle and not a button
- when an item is pinned, a pin icon is added next to its address and the pin toggle is on.
- the transfer button doesn't work: we can just send user to the transfer page as it's the case currently.

nguyenkims (commented):
Hey the new page seems promising 👍. I just added some comments, some don't have to be done in this PR. And can we try to have the same layout as the current one? Apart from the filter that we can handle later, I notice some differences: 
-  there's no "more" button in alias card
- pin should be a toggle and not a button
- when an item is pinned, a pin icon is added next to its address and the pin toggle is on.
- the transfer button doesn't work: we can just send user to the transfer page as it's the case currently.

PR #1721, opened 2023-05-06
fix(core): Open mailto: links in a new tab when the default email client is set to a web mail

mlec1:
- Open "mailto:" links in a new tab if using browser

Due to this commit: https://github.com/simple-login/app/commit/66c3a07c922bad0f11114e7047d6cb26adefaee6
the package-lock.json was not compatible with the docker image `npm:10`, I therefore updated the docker image as the `npm:10` is not supported for a long time now.

This MR is also opening the "mailto:" links into a new tab when the default client is set to Gmail, Proton, ... in the browser


templates/dashboard/alias_contact_manager.html
```diff
@@ -133,6 +133,7 @@ <h1 class="h3">
           <div>
             <span>
               <a href="{{ 'mailto:' + contact.website_send_to() }}"
+                 target="_blank"
```
nguyenkims, line 136: Hey I'm not sure what is the benefit of opening this link in a new tab?

mlec1: That's especially for this link, that I created this MR 😅.

When I open a new application, I like having it in a new tab, allowing me to switch from one to another without losing the change I did on one. It is also easier if I navigate a lot in my email client and want to go back to SL, I don't have to press the "previous" button X times.

But it is a personal preference

nguyenkims: I haven't tried on other OSes but in my case, the browser will ask me to open an email application without navigating to another page so I don't need to use the back button. I'm wondering if this only happens if the default email client is set to a web mail ?

mlec1: Yes, I was talking about having the default email client set to a web mail. I tried also with the default mail client set to Thunderbird and opening the "mailto" links into a new tab, and it has no effect when the mail client is a desktop app.

nguyenkims:
Hey I'm wondering if we can extract the Node update to another merge request?

mlec1:
> Hey I'm wondering if we can extract the Node update to another merge request?

I have reverted the Node Version update (I will see to update in a new merge request for this), and I have renamed this merge request.

PR #1756, opened 2023-05-29
Add timeout to any outbound connection

acasajus:


app/subscription_webhook.py
```diff
@@ -22,7 +22,7 @@ def execute_subscription_webhook(user: User):
         "active_subscription_end": sl_subscription_end,
     }
     try:
-        response = requests.post(webhook_url, json=payload)
+        response = requests.post(webhook_url, json=payload, timeot=2)
```
nguyenkims, line 25: there seems to be a typo in `timeot`

acasajus: Fixed it.

PR #1784, opened 2023-06-23
Preserve original from header in X-SimpleLogin-Original-From

acasajus:


xyxzxyz:
Thank you for this update! This is exactly what I needed along with preservation fix and they are now fixed today. Icons/favicons are showing up in Proton since this update.

image

acasajus:
Ah, that's fine then. We added integration in Proton so it's showing the original from natively. But the headers are indeed replaced.

acasajus:
If your sieve is using the from headers it should work... :thinking:

bartbutler:
We are going to replace the From header on Proton side with X-SimpleLogin-Original-From and the original From with the alias will reside in X-PM-Original-From (and usually Reply-To). Sorry for breaking filters but this is the cleanest way forward we think.

bartbutler:
My understanding is yes, though that will be still populated in the Reply-To header.

acasajus:
@notDavid that setting is still relevant for users that don't user Proton as their email provider :+1:

bartbutler:
Only if you keep the default and don't turn on hiding the original recipient. Can view this as more of a Proton-side integration than a SL feature, the SL headers and ARC signing allow it to be done safely.

PR #1829, opened 2023-08-01
Sanitize alias, contacts, mailboxes and users before creating them

acasajus:


app/models.py
```diff
@@ -1802,18 +1805,22 @@ def create(cls, **kw):
         commit = kw.pop("commit", False)
         flush = kw.pop("flush", False)
 
-        new_contact = cls(**kw)
-
-        website_email = kw["website_email"]
         # make sure email is lowercase and doesn't have any whitespace
-        website_email = sanitize_email(website_email)
+        website_email = sanitize_email(kw["website_email"]).replace("\u200f", "")
```
nguyenkims, line 1809: can we add `replace("\u200f", "")` to `sanitize_email()`?

app/models.py
```diff
@@ -2599,6 +2607,26 @@ def aliases(self) -> [Alias]:
 
         return ret
 
+    @classmethod
+    def create(cls, **kw):
+        # whether to call Session.commit
+        commit = kw.pop("commit", False)
+        flush = kw.pop("flush", False)
+
+        if "email" in kw:
```
nguyenkims, line 2616: can we call Base.create() after sanitizing the email that will take care of the commit and flush options?

acasajus: Done

crontab.yml
```diff
@@ -2,64 +2,66 @@ jobs:
   - name: SimpleLogin growth stats
     command: python /code/cron.py -j stats
     shell: /bin/bash
-    schedule: "0 0 * * *"
```
nguyenkims, line 5: For this one can we keep the same hour so we can compare it to past data?

acasajus: Sure. But 15m should not make a noticeable difference.

nguyenkims (approved):

PR #1835, opened 2023-08-04
chore: add upcloud monitoring

cquintana92:
This MR adds a job to `monitoring.py` in order to scrape metrics from Upcloud (specifically for the PostgreSQL databases) and report them to NewRelic.

The code for the monitoring is inside a new `monitor` director in order not to interfere with the already-existing `monitor` package inside `app`.


monitor/upcloud.py
```diff
@@ -0,0 +1,68 @@
...
+
+import base64
+import requests
+from typing import Any
+
+
+BASE_URL = "https://api.upcloud.com"
+
+
+def get_metric(json: Any, metric: str) -> UpcloudMetric:
```
nguyenkims, line 12: Nits: do we want to prefix these methods with _ to mark them as private?

monitor/newrelic.py
```diff
@@ -0,0 +1,24 @@
...
+class NewRelicClient:
+    def __init__(self, license_key: str):
+        self.__client = MetricClient(license_key)
+
+    def send(self, metrics: UpcloudMetrics):
+        batch = []
+
+        for metric in metrics.metrics:
+            for record in metric.records:
+                batch.append(
```
nguyenkims, line 15: Any reason that we don't use `newrelic.agent.record_custom_metric`?

acasajus: Custom metrics don't allow to add tags. We're using them to differentiate each db server.

acasajus: We're still trying this out though and see how it's shown in newrelic. it's weird that custom  metrics don't support tags though.

nguyenkims: by db server, do you mean staging and prod? Currently staging uses a completely different newrelic account.

acasajus: no, master and replicas

monitor/upcloud.py
cquintana92: Marked as public because:
- They have tests, so we can easily import them
- In case we wanted to reuse them for any other upcloud metric in another exporter

monitor/newrelic.py
cquintana92: `newrelic.agent.record_custom_metric` doesn't allow us to add metadata to the metric (such as the host).

In order to add that metadata, we are forced to do it as `Custom/name/label`, and it only supports one level of metadata.

The NewRelic Telemetry SDK allows us to define the variable type (Gauge in our case) and add any extra information to the metric we may need. For now it's only the server name, but could be expanded to more, should we need it.

acasajus: Ok. It's working and it allows to have metrics with dimensions per host. :tada:

monitor/upcloud.py
acasajus: It's used in the tests. It's fine. If we need to use them from other place they don't touch any private thing. They are just converters.

PR #1843, opened 2023-08-14
Update pricing page text

D-Bao:
This PR updates information from Proton Unlimited plan (price and amount of calendars) and removes the text "15 email addresses/aliases" as users found it confusing with the text "Unlimited aliases" on the left.


nguyenkims (approved):
LGTM!

PR #1863, opened 2023-09-04
add proton.ch to the is_proton check

nguyenkims:


xncelt:
Son,

Regarding your Proton 2023 article,  https://proton.me/blog/pass-roadmap-2023, you did not write anything about the hot topic of "passkeys".  Is support for passkeys in ProtonPass anywhere are the radar?

nguyenkims:
@xncelt it's indeed in our radar. We're closely monitoring it. Currently not a lot of websites support passkey yet so we don't prioritize its development for the moment.

xncelt:
Thx for your reply.

My crypto exchange, Gemini, just announced they are supporting passkeys.

I discovered that every website that lets you use your Google account to login by proxy supports passkeys when I am redirected to Google where I ogin with my Google passkey. Interesting, right?

Roger Newhall
***@***.***
215-510-2255

PR #1872, opened 2023-09-07
Schedule deletion of users

acasajus:
users to be scheduled to be deleted cannot:
 - receive or send emails
 - login
 
 For now the cron is disabled in case this gets deployed to prod before extensive tests


cron.py
```diff
@@ -1106,6 +1106,18 @@ def notify_hibp():
...
 
+def clear_users_scheduled_to_be_deleted():
+    users = User.filter(
+        and_(User.delete_on.isnot(None), User.delete_on < arrow.now())
+    ).all()
+    for user in users:
+        LOG.i(
+            f"Scheduled deletion of user {user} with scheduled delete on {user.delete_on}"
+        )
+        user.delete(user.id)
```
nguyenkims, line 1117: this should rather be `Session.delete(user)` right?

nguyenkims (commented):
LGTM, just 1 comment.

cron.py
acasajus: It's the same. I did it the same way as the job runner does https://github.com/simple-login/app/blob/master/job_runner.py#L222

nguyenkims: In this one it's `User.delete` (note that uppercase U)

acasajus: good one! Then how the ** do the test work :thinking: ?

acasajus: ah, bc it's a static method and the user object calls it regardless.. Changed it.

nguyenkims: Oh that's true.

PR #1901, opened 2023-09-29
Update dependencies

acasajus:


pyproject.toml
```diff
@@ -112,6 +111,7 @@ cryptography = "37.0.1"
 SQLAlchemy = "1.3.24"
 redis = "^4.5.3"
 newrelic-telemetry-sdk = "^0.5.0"
+aiospamc = "0.10"
```
nguyenkims, line 114: this change might require running SL on staging for a couple of days to make sure there's no regression.

acasajus: It's already in staging 👍

PR #1902, opened 2023-09-30
Delete old email_log entries in batches to avoid table lock

acasajus:


cron.py
```diff
@@ -85,23 +85,34 @@ def delete_logs():
...
+    LOG.d("Deleting EmailLog older than 2 weeks")
 
-    max_dt = arrow.now().shift(weeks=-2)
-    nb_deleted = EmailLog.filter(EmailLog.created_at < max_dt).delete()
-    Session.commit()
+    total_deleted = 0
+    batch_size = 500
+    Session.execute("set session statement_timeout=30000").rowcount
+    queries_done = 0
+    while True:
```
nguyenkims, line 104: can we replace the "while true" by a for loop by calculating the number of batches? Something like `select count(*) from email_log where created < now() - 15 days`?

acasajus: Won't clean as much because the cutoff time will be fixed in time. But the next run should pick it up. Soon we'll need to parallelize this to be able to keep up with the load. As it is now it takes hours to clean up.

PR #1909, opened 2023-10-07
Update procedure of selfhosting

hoyohayo:


README.md
```diff
@@ -240,7 +241,7 @@ docker run -d \
     -v $(pwd)/sl/db:/var/lib/postgresql/data \
     --restart always \
     --network="sl-network" \
-    postgres:12.1
+    postgres:13
```
nguyenkims, line 244: we haven't tested postgres:13 yet, as this is run by docker, maybe it's safer to use a tested version?

README.md
```diff
@@ -329,8 +330,8 @@ smtpd_recipient_restrictions =
    reject_unknown_recipient_domain,
    permit_mynetworks,
    reject_unauth_destination,
-   reject_rbl_client zen.spamhaus.org,
-   reject_rbl_client bl.spamcop.net,
+   reject_rbl_client zen.spamhaus.org=127.0.0.[2..11],
```
nguyenkims, line 333: do you have any doc on spamhaus pointing to this change?

nguyenkims (commented):
Can you please target this branch `new-self-host-version` instead? It contains the latest version of the doc and will be merged back to master soon.

README.md
hoyohayo: It probably is, I am running with version 13 since 6 months now. I can revert the change if needed

README.md
hoyohayo: I do not. Sorry, will revert.

README.md
```diff
@@ -74,9 +74,10 @@ Setting up DKIM is highly recommended to reduce the chance your emails ending up
 First you need to generate a private and public key for DKIM:
 
 ```bash
-openssl genrsa -out dkim.key 1024
+openssl genrsa -traditional -out dkim.key 1024
```
Spitfireap, line 77: @nguyenkims since dkimpy v1.1.0, [PKCS#8 is supported](https://git.launchpad.net/dkimpy/commit/?id=20d9ca1e1c2ce1fe4bb4b8b52c423de816aaebd7). What about you increase the minimum version instead?

nguyenkims: Do you know if it's compatible with the old key that people generated? I want to make sure that when people upgrade SL, they don't have any issue.

Spitfireap: It's a [try/except](https://git.launchpad.net/dkimpy/diff/dkim/crypto.py?id=20d9ca1e1c2ce1fe4bb4b8b52c423de816aaebd7) trying the old key standard first so still compatible :D.

README.md
Spitfireap: I agree with @hoyohayo : if using an open resolver almost every email would get blocked (https://www.spamhaus.org/news/article/788/spamhaus-dnsbl-return-codes-technical-update) and the standard return codes are [available here](https://www.spamhaus.org/faq/section/DNSBL%20Usage#200).

Metabaron1: I confirm en error when  using open resolvers, here are the error and spamhaus link:
554 5.7.1 Service unavailable; Client host [xxx] blocked using zen.spamhaus.org; Error: open resolver; https://www.spamhaus.org/returnc/pub/172.71.133.37;
As workaround I'm using google DNS for my dockers but I don't really like it

Spitfireap: @Metabaron1 I would recommend [something like this](https://www.linuxbabe.com/ubuntu/set-up-local-dns-resolver-ubuntu-20-04-bind9) alongside a PTR and you're good to go ;)

README.md
```diff
@@ -418,6 +421,12 @@ LOCAL_FILE_UPLOAD=1
 POSTFIX_SERVER=10.0.0.1
 ```
 
+Now it is time to build the latest docker. Replace the tag by latest version released (as of writing 4.35.2).
```
paulius-valiunas, line 424:
```suggestion
Now it is time to build the latest docker image. Replace the tag by latest version released (as of writing 4.35.2) or any other tag that you prefer.
```

README.md
```diff
@@ -430,11 +439,9 @@ docker run --rm \
     -v $(pwd)/dkim.pub.key:/dkim.pub.key \
     -v $(pwd)/simplelogin.env:/code/.env \
     --network="sl-network" \
-    simplelogin/app:3.4.0 flask db upgrade
+    simplelogin/app:4.35.2 alembic upgrade head
```
paulius-valiunas, line 442: Use the same tag as above (here and below). Users might forget to replace the version number.

README.md
paulius-valiunas: Also, is it safe to build directly from master? How stable is it?

paulius-valiunas (commented):
the docker commands are a bit unclear

Chinoman10:
As a self-hosting enthusiast, and while I'd typically prefer paying the $36/yr for 'managed SimpleLogin' (_too critical of infra to be worried of downtime_), I very much prefer having a simple/single **Docker Compose** file to copy-paste and use/customize, than multiple steps that require DB migrations and running specific commands inside the containers, etc... that allows us to keep the setup as portable as possible, and keeping the config's (and even the data, if we use named volumes) extremely easy to backup and service.

chrisblech:
> I very much prefer having a simple/single **Docker Compose** file

Check this repo: https://github.com/springcomp/self-hosted-simplelogin

@springcomp baked most steps inside a Compose file, and added the missing postfix container (instead of having postfix as host dependency). In the next few weeks (depending on my spare time 😅 ) I will contibute another PR there that will simplify setup by automate creating a DKIM key, set admin flag to the first user and use docker volumes instead of host mounts (goal: get this project up and running via Portainer, without a bash on the host)

Issue #1905, opened 2023-10-03
Some “bad” words


ghost:
Hello!

I think you should remove this words as they can trigger someone. It is not a best idea to pick the topic of them:

First:
https://github.com/simple-login/app/blob/master/local_data/words.txt#L92C9-L92C9

https://github.com/simple-login/app/blob/master/local_data/words.txt#L378 - an insult

https://github.com/simple-login/app/blob/master/local_data/words.txt#L379 - an insult

https://github.com/simple-login/app/blob/master/local_data/words.txt#L449 - an insult

https://github.com/simple-login/app/blob/master/local_data/words.txt#L3343 - insult

https://github.com/simple-login/app/blob/master/local_data/words.txt#L3360 - insult

@acasajus and @nguyenkims

nguyenkims:
Thanks, can you please make a PR to remove these words?

Tctxtvu:
@nguyenkims

nguyenkims:
Thanks, the PR has been merged.

PR #1921, opened 2023-10-15
Remove sensitive words

UserBob6:
Fix https://github.com/simple-login/app/issues/1905


nguyenkims (approved):
Thanks for removing these words!

PR #1935, opened 2023-11-06
remove sensitive words

Hulk667i:
@nguyenkims please take a look


acasajus:
LGTM. Thanks for looking into the wordlist.

Issue #527, opened 2021-07-10
Support for S/MIME


tatticoder:
Does Simple Login support digital signatures emails to verify you as the legitimate sender of the message. 
I was able to generate `Secure/Multipurpose Internet Mail Extensions` certificate for my private domain based email but could not attach it via Mozilla Thunderbird.
I checked documentation but could not find anything relevant.

nguyenkims:
Hi currently SimpleLogin doesn't support digital signature yet. Please feel free to create a feature request on https://github.com/simple-login/app/discussions along with additional info that might help the implementation :)

muibusan:
This is the link to the request: https://github.com/simple-login/app/discussions/535 :) Please vote

PR #1946, opened 2023-11-19
feat: support s/mime encryption

schwigri:
Hi, I use S/MIME to enable reading encrypted email using iOS and macOS Mail apps :) I've made this change and had it deployed on my own self-hosted instance, but thought I would share here. I've read through the contributing guide, but this is my first real PR so please let me know if there is anything I missed.

Also not sure if there should be other features considered (e.g. disable S/MIME for specific aliases like PGP, or some kind of fallback for S/MIME fails but fallback to PGP, etc.), so I don't really expect this will be merged immediately but thought it may be useful for others who may be interested in this feature. It could also be expanded to include signing, but for now it didn't seem necessary given the dynamic nature of the sending address.

Related to issue #527 and discussion #535!


email_handler.py
```diff
@@ -908,6 +925,26 @@ def forward_email_to_mailbox(
...
+            msg = prepare_smime_message(msg, mailbox.smime_public_key)
+        except Exception as exceptasdf:
+            LOG.w(
+                "Cannot S/MIME encrypt message %s -> %s. %s %s",
+                contact,
+                alias,
+                mailbox,
+                user,
+            )
+            LOG.w(exceptasdf)
```
schwigri, line 942: Oops, sorry; I can clean up this variable name and logging

nguyenkims:
Hi thanks a lot for making this PR and sorry for the late response! I'll ask our team to review the PR and come back to you at the beginning of 2024 (as most of the team is on holiday during Christmas :) )

Jekyle9444 (approved):

acasajus:
Hey, thanks for the contribution. As a general rule we'd like to have tests for new code that's merged into the repo to make sure that:

- the feature works as expected
- future changes don't break it.

Would you mind adding tests to it?

app/models.py
```diff
@@ -2566,6 +2566,12 @@ class Mailbox(Base, ModelMixin):
         sa.Boolean, default=False, nullable=False, server_default="0"
     )
 
+    # smime
+    smime_public_key = sa.Column(sa.Text, nullable=True)
```
acasajus, line 2570: Can you move this into a different model like `MailboxSMimeKey` ? We load mailboxes all the time from the db and the lighter they are, the better. We only need to load the key when we're about to use it.

app/dashboard/views/mailbox_detail.py
```diff
@@ -179,12 +179,53 @@ def mailbox_detail_route(mailbox_id):
...
+            )
+        elif request.form.get("form-name") == "smime":
+            if request.form.get("action") == "save":
+                if not current_user.is_premium():
+                    flash("Only premium plan can add S/MIME Key", "warning")
+                    return redirect(
+                        url_for("dashboard.mailbox_detail_route", mailbox_id=mailbox_id)
+                    )
+
+                mailbox.smime_public_key = request.form.get("smime")
```
acasajus, line 203: Can you:
 - Verify that this is a valid PEM file.
 - Add a check that the key is not bigger than 15KiB? A PEM chain with several certs shouldn't be bigger than that.

email_handler.py
acasajus: 👍

schwigri:
Thanks for reviewing and commenting. I’ll take some time to address them due to limited bandwidth. Appreciate it!

Teko012:
@schwigri Any updates on this? :)

szapp:
This is a highly desired feature especially for iOS users as discussed in #2174. Thanks @schwigri for tackling the implementation!

The author of this PR seems to not have the capacity at the moment to follow up the code review. @acasajus, is there anything, to be done to pick up this pull request again?

Would it be worth for me (new to the code base) to spend some of my free time on this, i.e. does this PR stand a chance to get merged in the near future?

schwigri:
@szapp (and everyone else here) — my apologies for opening this PR and not following up. I've had it running like this on a VPS and it has met all my needs, and due to my lack of familiarity with Docker and Python, I have not had enough time to dedicate to making this PR meeting high quality bar. I think it is a good foundation to build off of, but I can't commit to any updates soon. I don't want to take my current instance offline due to it handling all of my email, and I don't have the time to spin up a new VPS for testing, etc.

I see some useful comments, such like:

> * Verify that this is a valid PEM file
> * Add a check that the key is not bigger than 15KiB? A PEM chain with several certs shouldn't be bigger than that.

Without further research I'm not sure how to implement these simply, as well as add proper unit testing etc.

Really appreciate this product, the people that work on it, and glad to support via my Proton subscription separately :)

szapp:
@schwigri thanks for your response! I am glad you took the time to reply. It's good to hear that your implementation seems stable. If you don't mind, I can try and bring this PR to completion - although I cannot promise.

In order to continue this very PR, would you be willing to invite me as collaborator on your fork? Alternatively, I have no issue with making my own fork, but that would result in a new PR, leaving this one to die/possibly undermining your work.

Cheers!


Update: Because of limited time, I won't be able to contribute on this in the near future after all. Sorry about that!

schwigri:
I got a little bandwidth and recently spent some time to upgrade my self-hosted SL to latest (has been on 3.x since this PR was originally opened). I proposed a solution to the comment about validating the PEM file and its size, and left a clarification question in response to comment about the mailbox model.

In meantime I'll try out what I've proposed for the PEM validation and learn more about how the models work.

schwigri:
As brief follow up, I tested the PEM validation and file size validation and it works on my instance. Will take some time to investigate the models as requested.

PR #1976, opened 2023-12-20
Sanitize notes before exporting them

acasajus:


app/alias_utils.py
```diff
@@ -392,7 +392,10 @@ def alias_export_csv(user, csv_direct_export=False):
         )
 
         mailboxes = " ".join([mailbox.email for mailbox in alias_mailboxes])
-        data.append([alias.email, alias.note, alias.enabled, mailboxes])
+        alias_note = alias.note or ""
+        data.append(
```
nguyenkims, line 396: should we maybe quote the note to maintain the newline character? https://stackoverflow.com/questions/566052/can-you-encode-cr-lf-in-into-csv-files

acasajus: Hm.. then we'd better escape `"` also

acasajus: In fact the CSV exporter already does it. So we don't need to do anyting to escape `"`

PR #1981, opened 2023-12-29
Correct alias options version in docs

joedemcher:
I believe this may refer to the [published version](https://github.com/simple-login/app/blob/master/docs/api.md#get-apiv5aliasoptions) and not the former version.


nguyenkims (approved):
Thanks for spotting this!

PR #1989, opened 2024-01-01
Generate working DKIM keys by adding -traditional flag and update NGINX instructions to avoid breaking certbot

securityguy:
Add -traditional option to openssl genrsa to avoid Python DKIM library (dkimpy) error that prevents email from being sent:

dkim.asn1.ASN1FormatError: Unexpected tag (got 30, expecting 02)

Ref: https://bugs.launchpad.net/dkimpy/+bug/1708917

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,6 +515,8 @@ server {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


securityguy:
Update NGINX instructions to avoid certbot issues

nguyenkims (approved):
Thanks!

PR #1992, opened 2024-01-02
I don’t think we really need that

Kelp8:
Fix https://github.com/simple-login/app/issues/1995

This list is full of non words (like “a” letters) and other meaningless things. For now we have two lists main `words.txt` and `test_words.txt`

I think `words_alpha.txt` should be removed because it was depreciated by SL a long time ago (when SL users complained about bad words in aliases names). Also this list contains a lot of bad language and racism phrasing. I don’t think SL repo is good place for it.

If it will be needed sometime you will be always able to re-upload it

@nguyenkims, what will you say? :)


nguyenkims (approved):
Nice catch! We don't indeed use this file anywhere

PR #2011, opened 2024-01-13
Update README.md

edleeman17:
Updated README.md to prevent Nginx redirecting the browser to the local address of the machine

### DIFF
--- a/README.md
+++ b/README.md
@@ -510,7 +510,8 @@ server {
     server_name  app.mydomain.com;
 
     location / {
-        proxy_pass http://localhost:7777;
+        proxy_pass              http://localhost:7777;
+    	proxy_set_header        Host $host;
     }
 }
 ```


acasajus:
Thanks for your submission. Afaik that will pass the host header to the python code. It won't redirect the browser will it?

edleeman17:
Hey @acasajus, thanks for your comment! I had this issue when setting up the services. I was navigating to my public domain, but Nginx would redirect to the internal IP that the service was running on. I set `proxy_set_headers` to forward the Host. I found the following resolution:

```
The Host header is commonly used by web apps to determine how the client is getting to the web server. If it's not there - it typically defaults to root domain on the web server
```

Although, I'm concerned that this hasn't popped up before and whether it is only happening to me... Are you able to reproduce?

acasajus:
We use a slightly different config in prod. But I've reprod the issue with a dummy setup. Thanks for the fix 🚀

PR #2017, opened 2024-01-23
fix space issue on footer

nguyenkims:
Currently the logo touches the separator

image


raegedoc:
Hi!  There is still a problem in the "Settings" tab while self-hosted :(

image

nguyenkims:
I made a fix, should be available in the next version.

PR #2018, opened 2024-01-25
fix footer spacing again

nguyenkims:


nguyenkims:
The SL web app and landing page don't use the same frontend framework that explains the difference in terms of layout.

raegedoc:
Just updated to 4.42.2 and all is now perfect!  Thank you :)

PR #2045, opened 2024-03-02
fix typo

jal60:


templates/auth/fido.html
```diff
@@ -11,7 +11,7 @@
   <div class="card">
     <div class="card-body">
       <div class="mb-2">
-        Your account is protected with your security key (WebAuthn).
+        Your account is protected with your security key (WebAuth).
```
nguyenkims, line 14: WebAuthn is actually correct as it refers to https://webauthn.io/

PR #2046, opened 2024-03-03
feat: add generic OIDC connect

muhlba91:
This PR adds support for a generic OIDC authentication. Users can login and also signup using a custom OIDC server.

Administrators can deploy this by setting `OIDC_CLIENT_ID `, and the following environment variables are available to configure the OIDC flow:
- `CONNECT_WITH_OIDC_ICON`: the icon to be shown in the UI
- `OIDC_AUTHORIZATION_URL`: the OIDC auth endpoint
- `OIDC_USER_INFO_URL`: the OIDC user info endpoint to retrieve the user's email and name (if signing up)
- `OIDC_TOKEN_URL`: the OIDC token endpoint to exchange the code for a valid token
- `OIDC_SCOPES`: the OIDC scopes to use for performing the request
- `OIDC_NAME_FIELD`: the field containing the user's name (used for sign-up)
- `OIDC_CLIENT_ID`: the OIDC client id
- `OIDC_CLIENT_SECRET`: the OIDC client secret

This is based in the discussion in https://github.com/simple-login/app/pull/1425#issuecomment-1760301137.


app/auth/views/oidc.py
```diff
@@ -0,0 +1,131 @@
...
+    next_url = sanitize_next_url(request.args.get("next")) if request.args else None
+
+    return after_login(user, next_url)
+
+
+def create_user(email, oidc_user_data):
+    new_user = User.create(
+        email=email,
+        name=oidc_user_data.get(OIDC_NAME_FIELD),
+        password=random_string(20),
```
nguyenkims, line 123: I think you can use an empty password here.

app/config.py
```diff
@@ -244,6 +244,16 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
 FACEBOOK_CLIENT_ID = os.environ.get("FACEBOOK_CLIENT_ID")
 FACEBOOK_CLIENT_SECRET = os.environ.get("FACEBOOK_CLIENT_SECRET")
 
+CONNECT_WITH_OIDC = "CONNECT_WITH_OIDC" in os.environ
```
nguyenkims, line 247: can we add "ENABLE_" prefix to "CONNECT_WITH_OIDC" to make it more explicit that it's a boolean?

nguyenkims (commented):
I just left some comments.

acasajus:
Hey, thanks for the contribution. As a general rule we'd like to have tests for new code that's merged into the repo to make sure that:
 - the feature works as expected
  - future changes don't break it.
  Would you mind adding tests to it?

app/auth/views/oidc.py
```diff
@@ -0,0 +1,131 @@
...
+            "error",
+        )
+        return redirect(url_for("auth.login"))
+
+    email = sanitize_email(email)
+    user = User.get_by(email=email)
+
+    if not user and DISABLE_REGISTRATION:
+        flash(
+            "Sorry you cannot sign up via Github, please use email/password sign-up instead",
```
acasajus, line 102:
```suggestion
            "Sorry you cannot sign up via the OIDC provider. Please sign-up first with your email",
```

app/config.py
acasajus: Or even use directly if `OIRC_CLIENT_ID` is set to anything

muhlba91: i based it directly on `OIDC_CLIENT_ID`.

app/auth/views/oidc.py
muhlba91: thanks for noticing this one - fixed.

muhlba91:
@acasajus thank you for your review and comments! 😃
i have fixed your findings and added tests for the `oidc.py` methods.

app/auth/views/oidc.py
muhlba91: thanks, i used an empty string here now.

muhlba91:
@nguyenkims @acasajus bumping this one again - would you be able to review again and get it into main? 😃

Queuecumber:
So happy to see this merged! I'm going to close my PR for external auth since this should cover it

Would this be available in a docker container soon?

acasajus:
Should be already in the v4.41.2

Queuecumber:
Hey I just wanted to leave some feedback that over all this works well

To make it work with Google I did have to set the environment variable `OAUTHLIB_RELAX_TOKEN_SCOPE` otherwise I would get a warning because apparently google changes the order of the oauth scopes and that causes an exception to be thrown. I'm not entirely sure if there's a way to fix this in code or not.

Also a lot of the environment variables can/should probably be read from the .wellknown (https://accounts.google.com/.well-known/openid-configuration for example) rather than read manually

muhlba91:
@Queuecumber thank you for that feedback and info with the well-known endpoint!

do you happen to know if all OIDC provider must implement that well-known endpoint, meaning it's part of the specification?
the reason why i ask is that i would rather not like to rely reading from it if that means we potentially make it incompatible with some providers, whether self-hosted or not.
otherwise, i can see if the used oauth lib supports setting and reading from that endpoint.

Queuecumber:
Here's the spec for it: https://openid.net/specs/openid-connect-discovery-1_0.html

It's unclear if this is mandatory for the provider to implement, it is certainly optional for the RP (see Sec. 2). Google, for example, says applications should use the well-known endpoint. The advantage is that if they change one of the underlying URLs it will be reflected in the well-known endpoint without needing to reconfigure the application. The well-known endpoint shouldn't (ideally) change. It's also a lot less typing to set up.

The way I usually see this implemented is that the user can either fill out all the fields manually (i.e., what we have now) OR they can give just the well-known URL and have the application discover the endpoints.

muhlba91:
thank you! 👍🏻

i like your proposal of deciding it dynamically based on whether the well-known endpoint is given or not.

currently, i'm on travel but i'll see how i can integrate that! 😃

Queuecumber:
Awesome thanks, I did poke around and I don't think requests_oauthlib has support for it, but that may be intentional since it's just a simple API call and JSON parse to get the info 

I'm also noticing an error when the callback URL has a query string on it, like `https://<redacted>/auth/oidc/callback?next=%2Fdashboard%2F ` is a common one. It seems like you're not supposed to include those parameters in the query string of the callback, I can file a bug for this to keep things cleaner.

viq:
FWIW (which isn't necessarily much), from quickly checking the docs, following Open Source OIDC providers (that I could think of) **do** expose the `.well-known` endpoint:
- [Keycloak](https://www.keycloak.org/docs/latest/securing_apps/#endpoints)
- [Authentik](https://docs.goauthentik.io/docs/providers/oauth2/)
- [Dex](https://dexidp.io/docs/openid-connect/#discovery)
- [ORY Hydra](https://www.ory.sh/docs/hydra/self-hosted/production#exposing-administrative-and-public-api-endpoints)
- [OIDC connector for NextCloud](https://github.com/H2CK/oidc)
- [HashiCorp Vault](https://developer.hashicorp.com/vault/tutorials/auth-methods/oidc-identity-provider)
- [GitLab](https://docs.gitlab.com/ee/integration/openid_connect_provider.html)

viq:
> Awesome thanks, I did poke around and I don't think requests_oauthlib has support for it, but that may be intentional since it's just a simple API call and JSON parse to get the info

Looks like it may not, but from a quick search, [authlib](https://github.com/lepture/authlib) in their [docs for OAuth2](https://docs.authlib.org/en/latest/client/oauth2.html) say:
> This documentation covers the common design of a Python OAuth 2.0 client. Authlib provides three implementations of OAuth 2.0 client:
>   1.  [requests_client.OAuth2Session](https://docs.authlib.org/en/latest/client/api.html#authlib.integrations.requests_client.OAuth2Session) implementation of [OAuth for Requests](https://docs.authlib.org/en/latest/client/requests.html#requests-client), which is a replacement for requests-oauthlib.

and [this example](https://docs.authlib.org/en/latest/client/oauth2.html#automatically-refreshing-tokens) suggests they do support discovery via `.well-known`.

muhlba91:
#2077 should 1/ fix the redirect URL issue, and 2/ adds support for the OIDC well-known endpoint. as discussed, the endpoint takes precedence over all env variables but is an optional var.

Issue #2048, opened 2024-03-04
Add some useful info


ghost:
See https://github.com/simple-login/app/pull/2049

PR #2049, opened 2024-03-04
Add useful info

ghost:
Fix https://github.com/simple-login/app/issues/2048


nguyenkims (commented):
Hey thanks for the PR but I think most people don't know F-droid and for those who know, they can look for SL on F-droid. We also already show F-droid icon on our website 
image and footer 
image

which should be enough IMO.

nguyenkims:
We try to keep the emails as short & concise as possible and as the result, sometimes we need to remove some details that are only relevant for a subset of users.

PR #2050, opened 2024-03-04
Add more information about spoofer

TaDi6:
I think this changes will help potential victim to investigate situation

### DIFF
--- a/templates/emails/transactional/spoof-reply.html
+++ b/templates/emails/transactional/spoof-reply.html
@@ -10,6 +10,16 @@
 {% endcall %}
 
 {% call text() %}
+- From: <b>{{ alias }}</b>
+<br />
+- To: <b>{{ to_email }}</b>
+<br />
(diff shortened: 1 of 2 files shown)


TaDi6:
@nguyenkims

templates/emails/transactional/spoof-reply.html
```diff
@@ -9,6 +9,16 @@ <h1>
   </h1>
 {% endcall %}
 
+{% call text() %}
+- From: <b>{{ alias }}</b>
+<br />
+- To: <b>{{ to_email }}</b>
+<br />
+- Subject: <b>{{ subject }}</b>
```
nguyenkims, line 17: I think we don't have information like subject or to_email, you can see where this template is used on https://github.com/simple-login/app/blob/require-sudo-on-account-change/app/handler/dmarc.py#L173

PR #2056, opened 2024-03-11
ci: add arm docker images

martadams89:
Add arm64 images to ci pipeline

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,6 @@
 name: Test and lint
 
-on:
-  push:
-    paths:
-      - '.github/workflows/main.yml'
+on: [push, pull_request]
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


acasajus:
- can you update the action definition so it's also triggered when the workflow file changes so we can make sure this change doesn't break anything?

martadams89:
@acasajus thanks, I've updated to run only on

```
on:
  push:
    paths:
      - '.github/workflows/main.yml'
```

This means the workflow will only run when this file is updated - not sure if this is what you were asking for?

martadams89:
Anything you need me to do on the PR so that the action can be tested?

.github/workflows/main.yml
```diff
@@ -2,6 +2,8 @@ name: Test and lint
 
 on:
   push:
+    paths:
+      - '.github/workflows/main.yml'
```
acasajus, line 6:
```suggestion
on: [push, pull_request]
```

acasajus: Let's try this.

PR #2071, opened 2024-03-16
Add TOS notice on account activation

Geri680:
Fix https://github.com/simple-login/app/issues/2070

Something like Addy.io done:
IMG_2668


nguyenkims (commented):
Thanks for the idea. The UI is a bit not polished enough though, if you want we can ask our designers

image

PR #2073, opened 2024-03-16
Add TOS to start email to decrease accidental abuse

Geri680:
Fix https://github.com/simple-login/app/issues/2072

The same as https://github.com/simple-login/app/pull/2071 but for email


nguyenkims (commented):
Hi thanks for making the PR. We try to keep the emails as short as possible, and the info here actually doesn't concern most users. (most users aren't using SL for mass signups or any abusive activity)

PR #2077, opened 2024-03-24
feat: use oidc well-known url

muhlba91:
follow-up of #2046 to be able to use the well-known OIDC URL defined as `OIDC_WELL_KNOWN_URL` instead of specifying all OIDC related URLs.

if the `OIDC_WELL_KNOWN_URL` is specified all other OIDC related URL env variables are ignored and the configuration is read from the well-known endpoint.

additionally, the next URL is kept in the session (and cleared afterwards) to ensure the OIDC redirect URL is clean for the OIDC server to recognize correctly.


muhlba91:
@viq @Queuecumber this is the follow-up PR from #2046 

@nguyenkims @acasajus would you mind reviewing the PR? 😃

app/auth/views/oidc.py
```diff
@@ -60,19 +66,26 @@ def oidc_callback():
         flash("Please use another sign in method then", "warning")
         return redirect("/")
 
+    token_url = OIDC_TOKEN_URL
```
nguyenkims, line 69: As most IDP support the well-known URL, should we remove OIDC_TOKEN_URL, OIDC_USER_INFO_URL, etc and only keep OIDC_WELL_KNOWN_URL?

nguyenkims (commented):
The PR looks good to me, just left a question.

app/auth/views/oidc.py
viq: I think it may be useful to leave that capability as a fallback, since there probably is something out there that does not expose a proper well-known information.

muhlba91: in general, i agree with @viq that a fallback option could be useful for IDPs that don't expose it correctly - there are always some weird systems out there...
personally, i favor this one as it keeps flexibility for users. (and we'd also not break existing configurations.)

however, from the maintenance point of view it'd be a bit easier to maintain if we rely only on the well-known endpoint.

i believe this is a trade-off core maintainers need to decide.

nguyenkims: I think using well-known url is a great initiative as this avoids RP from having to hard code several values and even though if it isn't obligatory in the standard, it's supported by most IDP and we can decide to keep it as the only option supported in SL.

app/auth/views/oidc.py
```diff
@@ -16,14 +18,15 @@
 from app.email_utils import send_welcome_email
 from app.log import LOG
 from app.models import User, SocialAuth
-from app.utils import encode_url, sanitize_email, sanitize_next_url
+from app.utils import sanitize_email, sanitize_next_url
 
 
 # need to set explicitly redirect_uri instead of leaving the lib to pre-fill redirect_uri
 # when served behind nginx, the redirect_uri is localhost... and not the real url
 _redirect_uri = URL + "/auth/oidc/callback"
```
acasajus, line 26: Nitpicking. Can you rename this to `redirect_uri`? We don't need to keep the `_` prefix since it's already the url we're going to use.

muhlba91:
@acasajus @nguyenkims thank you for your feedback!
i have reduced to only supporting the well-known endpoint and renamed the `redirect_url` property as requested.
i'd appreciate if you could have another look. 😃

nguyenkims (approved):
LGTM, thanks!

PR #2084, opened 2024-04-09
Premium user can enable data breach monitoring

nguyenkims:


app/dashboard/views/setting.py
```diff
@@ -227,6 +227,21 @@ def setting():
             Session.commit()
             flash("Your preference has been updated", "success")
             return redirect(url_for("dashboard.setting"))
+        elif request.form.get("form-name") == "enable_data_breach_check":
+            if not current_user.is_premium():
+                flash(
+                    "Only premium plan can add enable data breach monitoring", "warning"
```
acasajus, line 233:
```suggestion
                    "Only premium plan can enable data breach monitoring", "warning"
```

app/dashboard/views/setting.py
```diff
@@ -227,6 +227,21 @@ def setting():
...
+                )
+                return redirect(url_for("dashboard.setting"))
+            choose = request.form.get("enable_data_breach_check")
+            if choose == "on":
+                current_user.enable_data_breach_check = True
```
acasajus, line 238:
```suggestion
                LOG.i("User {current_user} has enabled data breach monitoring")
                current_user.enable_data_breach_check = True
```

app/dashboard/views/setting.py
```diff
@@ -227,6 +227,21 @@ def setting():
...
+                flash("Data breach monitoring is enabled", "success")
+            else:
+                current_user.enable_data_breach_check = False
```
acasajus, line 241:
```suggestion
                LOG.i("User {current_user} has disabled data breach monitoring")
                current_user.enable_data_breach_check = False
```

app/dashboard/views/setting.py
acasajus: So when users enable or disable it's logged and we can answer CS queries

app/dashboard/views/setting.py
acasajus: Same here

PR #2086, opened 2024-04-11
Prevent proton mailboxes from enabling pgp encryption

acasajus:


app/dashboard/views/mailbox_detail.py
```diff
@@ -179,8 +179,15 @@ def mailbox_detail_route(mailbox_id):
...
-                flash(f"PGP is enabled on {mailbox.email}", "success")
+                if mailbox.is_proton():
+                    mailbox.disable_pgp = True
+                    flash(
+                        "Enabling PGP for a Proton Mail mailbox is redundant and does not add any security benefit",
+                        "info",
+                    )
+                else:
+                    mailbox.disable_pgp = False
+                    flash(f"PGP is enabled on {mailbox.email}", "info")
```
nguyenkims, line 190: "enabled" should be "disabled" here I think

acasajus: this request is to enable php. If we want to enable pgp then `disabled` has to be set to `true`. We only force it to `False` if it's a proton mailbox.

acasajus: The disable_x logic is quite confusing. I event had to edit the previous comment twice to get it right.

nguyenkims: ah that's true, I think the reason it was called disable_pgp as I want the default value for a new column is false

PR #2113, opened 2024-05-21
Add sync events

cquintana92:

1. Adds a new `proto/events.proto` file that defines events that a party may want to consume.
2. Adds a new `app/events/event_listener.py` that exposes an `EventListener` for dispatching events from within the app.
3. Adds a new `event_listener.py` entrypoint that can act as an event consumer or as a dead-letter consumer for the pending events.
4. Adds the event emit calls within the code.


app/api/views/alias.py
```diff
@@ -184,7 +184,7 @@ def toggle_alias(alias_id):
     if not alias or alias.user_id != user.id:
         return jsonify(error="Forbidden"), 403
 
-    alias.enabled = not alias.enabled
+    alias_utils.change_alias_status(alias, enabled=not alias.enabled)
     Session.commit()
```
acasajus, line 188: Here you have a double commit. The method is doing a commit and there's a commit afterwards here.

app/alias_utils.py
```diff
@@ -459,3 +465,14 @@ def transfer_alias(alias, new_user, new_mailboxes: [Mailbox]):
...
+
+def change_alias_status(alias: Alias, enabled: bool, commit: bool = True):
+    alias.enabled = enabled
+    if commit:
+        Session.commit()
+
+    event = AliasStatusChange(
+        alias_id=alias.id, alias_email=alias.email, enabled=enabled
+    )
+    EventDispatcher.send_event(alias.user, EventContent(alias_status_change=event))
```
acasajus, line 478: I would put the commit after the send_event. This way the modification and the event happen in a single Tx

app/dashboard/views/index.py
```diff
@@ -146,7 +146,7 @@ def index():
                 alias_utils.delete_alias(alias, current_user)
                 flash(f"Alias {email} has been deleted", "success")
             elif request.form.get("form-name") == "disable-alias":
-                alias.enabled = False
+                alias_utils.change_alias_status(alias, enabled=False)
                 Session.commit()
```
acasajus, line 150: Same here. it's a double commit. The function does commit and then commit outside.

app/dashboard/views/unsubscribe.py
```diff
@@ -31,7 +32,7 @@ def unsubscribe(alias_id):
 
     # automatic unsubscribe, according to https://tools.ietf.org/html/rfc8058
     if request.method == "POST":
-        alias.enabled = False
+        alias_utils.change_alias_status(alias, False)
         flash(f"Alias {alias.email} has been blocked", "success")
         Session.commit()
```
acasajus, line 37: double commit also here

app/models.py
```diff
@@ -1625,6 +1625,18 @@ def create(cls, **kw):
...
+        from app.events.event_dispatcher import EventDispatcher
+        from app.events.generated.event_pb2 import AliasCreated, EventContent
+
+        event = AliasCreated(
+            alias_id=new_alias.id,
+            alias_email=new_alias.email,
+            alias_note=new_alias.note,
+            enabled=True,
+        )
+        EventDispatcher.send_event(user, EventContent(alias_created=event))
```
acasajus, line 1638: I'd move this before the flush/commit to ensure it's in the same tx

app/models.py
```diff
@@ -3648,3 +3660,42 @@ def create(cls, **kwargs):
...
 
         return super().create(code=code, **kwargs)
+
+
+class SyncEvent(Base, ModelMixin):
+    """This model holds the events that need to be sent to the webhook"""
+
+    __tablename__ = "sync_event"
+    content = sa.Column(sa.LargeBinary, unique=False, nullable=False)
+    taken_time = sa.Column(ArrowType, default=None, nullable=True, server_default=None)
```
acasajus, line 3670: We'll need an index by taken time and create time so that the deadletter uses it

cron.py
```diff
@@ -1221,6 +1223,7 @@ def clear_users_scheduled_to_be_deleted(dry_run=False):
         )
         if dry_run:
             continue
+        EventDispatcher.send_event(user, EventContent(user_deleted=UserDeleted()))
         User.delete(user.id)
```
acasajus, line 1227: Why not put it inside the `User.delete` method?

app/subscription_webhook.py
```diff
@@ -31,3 +33,6 @@ def execute_subscription_webhook(user: User):
             )
     except RequestException as e:
         LOG.error(f"Subscription request exception: {e}")
+
+    event = UserPlanChange(plan_end_time=sl_subscription_end)
```
acasajus, line 37: There are many routes that create subscriptions. ManualSub, AppleSub, Coinbase... I would put it in the create of each model so to avoid losing events. Except in PartnerSubscription

app/dashboard/views/unsubscribe.py
cquintana92: Changed it so it doesn't commit by default

app/models.py
cquintana92: Changed it

app/subscription_webhook.py
cquintana92: This is called for every subscription, so it will be invoked regardless of where it comes from

app/models.py
cquintana92: Added indexes

cron.py
cquintana92: Because it didn't exist beforehand, but I've created it now and moved the logic there

app/subscription_webhook.py
cquintana92: This method is used regardless of which subscription is used, so it will be called for all of them :+1:

acasajus (approved):

PR #2116, opened 2024-05-23
feat: implement HTTP event sink

cquintana92:
This PR implements the HTTP event sink, for sending the sync events using an HTTP transport.
It also adds a new config that allows to disable the SSL verification for that request (only if defined).

### DIFF
--- a/events/event_sink.py
+++ b/events/event_sink.py
@@ -1,4 +1,7 @@
+import requests
+
 from abc import ABC, abstractmethod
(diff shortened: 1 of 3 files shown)


events/event_sink.py
```diff
@@ -11,7 +14,23 @@ def process(self, event: SyncEvent):
 
 class HttpEventSink(EventSink):
     def process(self, event: SyncEvent):
-        pass
+        if not EVENT_WEBHOOK:
+            LOG.warning("No event webhook configured")
```
acasajus, line 18:
```suggestion
            LOG.warning("Skipping sending event because there is no webhook configured")
```

PR #2135, opened 2024-06-27
fix: convert POSTFIX_TIMEOUT to int

danfate:
When POSTFIX_TIMEOUT reading from env, it's default type is str, which will case an error
This PR converts it to int.


nguyenkims (approved):
LGTM!

PR #2136, opened 2024-06-28
Use partner emails when the user has used alias from a partner

acasajus:


app/email_utils.py
```diff
@@ -74,11 +75,16 @@ def render(template_name, **kwargs) -> str:
 
     template = env.get_template(template_name)
 
+    use_partner_template = False
```
nguyenkims, line 78: Usually the user is provided in kwargs so I think we can get the user by something like 
`user = kwargs.get("user") or current_user`

PR #2144, opened 2024-07-05
chore: QOL improvements on alias delete due to cascade FKs

cquintana92:

1. Adds a new "reason" column to DeletedAlias and DomainDeletedAlias.
2. In all the places that aliases are deleted due to a FK delete cascade, delete the aliases manually and add the reason.


app/models.py
```diff
@@ -2448,6 +2469,13 @@ def delete(cls, obj_id):
         if obj.is_sl_subdomain:
             DeletedSubdomain.create(domain=obj.domain)
 
+        from app import alias_utils
+
+        for alias in Alias.filter_by(custom_domain_id=obj_id):
```
acasajus, line 2474: This should also be done only if the CustomDomain is an SL domain. If the alias belongs to a user domain is should be able to be re-created.

PR #2176, opened 2024-08-01
remove “shady” words

ccb78:
@nguyenkims please take a look


acasajus:
Hey, thanks It looks great!

PR #2183, opened 2024-08-08
Admin improvements

acasajus:
Limit the alias and mailbox search to 10 entries to avoid timing out


cquintana92 (approved):

app/admin_model.py
```diff
@@ -753,12 +753,37 @@ def index(self):
...
+            )
+
+        def user_mailboxes_count(user_id: int) -> int:
+            return Mailbox.filter_by(user_id=user_id).order_by(Mailbox.id.asc()).count()
+
+        def user_aliases(user_id: int) -> list[Alias]:
+            return (
+                Alias.filter_by(user_id=user_id)
+                .order_by(Alias.id.asc())
+                .limit(10)
```
nguyenkims, line 771: this means we only return 10 aliases right?

acasajus: Yep, if there are more than 10 in the template we explicitly mention we only show the first 10.

acasajus: This is to solve some timeouts when checking account with loads of aliases.

PR #2184, opened 2024-08-09
chore: add config for enabling sync for specific users

cquintana92:


app/config.py
```diff
@@ -588,3 +588,17 @@ def getRateLimitFromConfig(
...
+
+
+def read_webhook_enabled_user_ids() -> Optional[List[int]]:
+    user_ids = os.environ.get("EVENT_WEBHOOK_ENABLED_USER_IDS", None)
+    if user_ids is None:
+        return None
+
+    ids = []
+    for id in user_ids.split(","):
+        ids.append(int(id.strip()))
```
acasajus, line 600:
```suggestion
        try:
           ids.append(int(id.strip()))
        catch ValueError:
           pass
```

acasajus (approved):

PR #2199, opened 2024-08-28
(feat)Add draft /.well-known/webfinger support (RFC 7033)

zipy124:
Implement /.well-known/webfinger ([RFC 7033](https://www.rfc-editor.org/rfc/rfc7033#section-3.1).) response as requested in #1927 .

This enables using simplelogin as a provider for tailscale for example.

n.b: this serves as a mere proof of concept and requires tests before merging.

### DIFF
--- a/server.py
+++ b/server.py
@@ -335,6 +335,20 @@ def setup_openid_metadata(app):
         res = {"keys": [get_jwk_key()]}
         return jsonify(res)
 
+def discover_openid_webfinger(app):
+    @app.route("/.well-known/webfinger")
+    @cross_origin()
+    def openid_webfinger():
+        res = {
+            "subject": request.args.get(resource),
+            "links": [{
+                    "rel": "http://openid.net/specs/connect/1.0/issuer",
+                    "href": URL
+            },],
+        }
+
+        return jsonify(res)
+
 
 def get_current_user():
     try:


nguyenkims (commented):
Thanks for adding the PR, is it possible to add a simple test for this new endpoint?

jan-leila:
what would be needed as a part of those tests? Would it just be making sure that arbitrary subjects get parsed correctly?

nguyenkims (approved):

nguyenkims:
@jan-leila sorry for the delay, I just merge it, it should be available in the next version

Billy5804:
@nguyenkims Before this gets added to production, do we know what will happen with domain ownership on tailscale?
When I signed up with an email using my custom domain using webfinger it made me the admin of the entire domain, so that any other users who signed up with my domain would be added to my tailnet.
https://tailscale.com/kb/1259/domain-ownership

jan-leila:
given that we might want to put in some filtering on this endpoint before it gets merged to not resolve for any simple-login owned domains, otherwise we might open up users who think it would be funny to join the aleeas.com tailnet to network vulnerablies due to some other untrusted user who got there first being the admin of that tailnet

jan-leila:
I think we would want something like this:
```py
def discover_openid_webfinger(app):
    @app.route("/.well-known/webfinger")
    @app.errorhandler(403)
    @cross_origin()
    def openid_webfinger():
        subject = request.args.get(resource)
        domain = subject[rfind("@"):]
        
        # get real value for this I don't have a full list of domains managed by simple login 
        simple_login_domains = ["aleeas.com"]
        
        if domain in simple_login_domains:
          return jsonify(error="Forbidden"), 403
          
        res = {
            "subject": request.args.get(resource),
            "links": [{
                    "rel": "http://openid.net/specs/connect/1.0/issuer",
                    "href": URL
            },],
        }
        
        return jsonify(res)
```

I'm not familiar with the project as a whole though and don't know where I would source the complete value for `simple_login_domains` (or even if there is a single source of truth for that)

And I think given the security implications of this we definitely should write some tests. I have never worked with flask before but if someone could point me to some example tests I would be more then willing to give writing them a stab.

Billy5804:
SimpleLogin isn't responsible for handling webfinger for users custom domains, so if you were to exclude all of SimpleLogins owned domains what would be the purpose of the webfinger route?

If you are after using a custom domain you can check this comment I've added to the discussion: https://github.com/simple-login/app/discussions/1927#discussioncomment-12318678.

wurongxin52-cloud:
twitter

wurongxin52-cloud:
twitter

PR #2215, opened 2024-09-13
feat: extract custom domain utils to a service

cquintana92:

1. Offer a new function that loads a dictionary from an env var
2. Allow to define partner domains (not in use yet)
3. Extracting logic from the controller into a service so it can be reused
4. Move the DMARC record to `app.constants`

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -35,6 +35,33 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
     return literal_eval(value)
 
 
+def get_env_dict(env_var: str) -> dict[str, str]:
+    """
+    Get an env variable and convert it into a python dictionary with keys and values as strings.
+    Args:
+        env_var (str): env var, example: SL_DB
+
+    Syntax is: key1=value1;key2=value2
+    Components separated by ;
+    key and value separated by =
+    """
+    value = os.getenv(env_var)
+    if not value:
+        return {}
+
+    components = value.split(";")
+    result = {}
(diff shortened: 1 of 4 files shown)


app/config.py
```diff
@@ -35,6 +35,34 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
...
+
+    components = value.split(";")
+    result = {}
+    for component in components:
+        if component == "":
+            continue
+        parts = component.split("=")
+        if len(parts) != 2:
+            raise Exception(f"Invalid config for env var {env_var}")
+        result[parts[0]] = parts[1]
```
acasajus, line 61:
```suggestion
        result[parts[0].strip()] = parts[1].strip()
```

app/config.py
```diff
@@ -35,6 +35,34 @@ def sl_getenv(env_var: str, default_factory: Callable = None):
     return literal_eval(value)
 
 
+def get_env_dict(env_var: str, default_factory: Callable = None) -> dict[str, str]:
```
acasajus, line 38: Worth leaving default as empty dict directly?

cquintana92: Or even better, not allowing to pass it as it will always be a dict

acasajus: 👍

app/config.py
cquintana92: Changed

acasajus (approved):

PR #2221, opened 2024-09-16
chore: refactor create custom domain

cquintana92:

1. Fix the `scripts/new-migration.sh` script to use poetry commands.
2. Add a migration to add `custom_domain.partner_id column`.
3. Refactor CustomDomain creation to a service so it can be reused.
4. Allow to specify partner_id when creating a custom_domain.


app/custom_domain_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    if new_domain.startswith("http://"):
+        new_domain = new_domain[len("http://") :]
+
+    if new_domain.startswith("https://"):
+        new_domain = new_domain[len("https://") :]
+
+    return new_domain
+
+
+def can_domain_be_used(user: User, domain: str) -> Optional[str]:
```
acasajus, line 51: I would rename this method to `check_if_domain_can_be_used` . I would expect a method that's called `can_....` to return `True` if it suceeds

app/custom_domain_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    if not is_valid_domain(domain):
+        return "This is not a valid domain"
+    elif SLDomain.get_by(domain=domain):
+        return "A custom domain cannot be a built-in domain."
+    elif CustomDomain.get_by(domain=domain):
+        return f"{domain} already used"
+    elif get_email_domain_part(user.email) == domain:
+        return "You cannot add a domain that you are currently using for your personal email. Please change your personal email to your real email"
+    elif Mailbox.filter(
+        Mailbox.verified.is_(True), Mailbox.email.endswith(f"@{domain}")
```
acasajus, line 61: Note: We may need to check if postgres allows to have two indexes over a field to have a `gin` index on the mailbox email so that we can do efficient `LIKE` queries.

app/custom_domain_utils.py
cquintana92: Refactored to return a result type with the cause

acasajus (approved):

PR #2224, opened 2024-09-17
chore: refactor dns to improve testability

cquintana92:

1. Encapsulate all the DNS logic into classes that implement base class methods.
2. Remove global functions in favour of passed-on implementations (IoC).
3. Add tests to `app/custom_domain_validation.py` before extending the code.

### DIFF
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -39,15 +39,31 @@ class CustomDomainValidation:
         Check if dkim records are properly set for this custom domain.
         Returns empty list if all records are ok. Other-wise return the records that aren't properly configured
         """
+        correct_records = {}
         invalid_records = {}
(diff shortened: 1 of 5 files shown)


app/custom_domain_validation.py
```diff
@@ -39,9 +40,9 @@ def validate_dkim_records(self, custom_domain: CustomDomain) -> dict[str, str]:
         Returns empty list if all records are ok. Other-wise return the records that aren't properly configured
         """
         invalid_records = {}
-        for prefix, expected_record in self.get_dkim_records():
+        for prefix, expected_record in self.get_dkim_records().items():
```
acasajus, line 43: Did it fail before?

cquintana92: No, but it was returning a list of tuples and using it as a dict, so now it's changed to a dict

PR #2225, opened 2024-09-18
feat: allow to define partner records

cquintana92:
This PR adds the ability to define partner records and prefixes for custom domain validation.
It perorms the following changes:

1. Extract the `get_ownership_dns_txt_value` function from the model into a function of the `CustomDomainValidation`.
2. Remove the direct call from the template to the aforementioned function in order to pass the value from the controller.
3. Add the new configs that define the domain and the verification prefix for each partner.
4. Use these configs in the CustomDomainValidation, with fallbacks for the existing behaviour.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -638,19 +638,22 @@ EVENT_WEBHOOK_ENABLED_USER_IDS: Optional[List[int]] = read_webhook_enabled_user_
 EVENT_LISTENER_DB_URI = os.environ.get("EVENT_LISTENER_DB_URI", DB_URI)
 
 
-def read_partner_domains() -> dict[int, str]:
-    partner_domains_dict = get_env_dict("PARTNER_DOMAINS")
-    if len(partner_domains_dict) == 0:
+def read_partner_dict(var: str) -> dict[int, str]:
+    partner_value = get_env_dict(var)
+    if len(partner_value) == 0:
         return {}
 
     res: dict[int, str] = {}
-    for partner_id in partner_domains_dict.keys():
+    for partner_id in partner_value.keys():
         try:
             partner_id_int = int(partner_id.strip())
-            res[partner_id_int] = partner_domains_dict[partner_id]
+            res[partner_id_int] = partner_value[partner_id]
         except ValueError:
             pass
     return res
 
 
-PARTNER_DOMAINS: dict[int, str] = read_partner_domains()
+PARTNER_DOMAINS: dict[int, str] = read_partner_dict("PARTNER_DOMAINS")
+PARTNER_DOMAIN_VALIDATION_PREFIXES: dict[int, str] = read_partner_dict(
+    "PARTNER_DOMAIN_VALIDATION_PREFIXES"
+)
--- a/app/custom_domain_validation.py
+++ b/app/custom_domain_validation.py
@@ -1,4 +1,12 @@
-from app.config import EMAIL_SERVERS_WITH_PRIORITY, EMAIL_DOMAIN
+from dataclasses import dataclass
+from typing import Optional
+
+from app.config import (
+    EMAIL_SERVERS_WITH_PRIORITY,
+    EMAIL_DOMAIN,
+    PARTNER_DOMAINS,
+    PARTNER_DOMAIN_VALIDATION_PREFIXES,
+)
 from app.constants import DMARC_RECORD
 from app.db import Session
 from app.dns_utils import (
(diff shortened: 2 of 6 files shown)


app/config.py
```diff
@@ -638,19 +638,22 @@ def read_webhook_enabled_user_ids() -> Optional[List[int]]:
 EVENT_LISTENER_DB_URI = os.environ.get("EVENT_LISTENER_DB_URI", DB_URI)
 
 
-def read_partner_domains() -> dict[int, str]:
-    partner_domains_dict = get_env_dict("PARTNER_DOMAINS")
-    if len(partner_domains_dict) == 0:
+def read_partner_dict(var: str) -> dict[int, str]:
```
acasajus, line 641: I would rename this method to `read_config_dict` so that it can be used in other places that have a similar need.

app/custom_domain_validation.py
```diff
@@ -1,4 +1,12 @@
-from app.config import EMAIL_SERVERS_WITH_PRIORITY, EMAIL_DOMAIN
+from dataclasses import dataclass
+from typing import Optional
+
+from app.config import (
```
acasajus, line 4: If we import `from app import config` and then use it as `config.VAR_NAME` (for example `config.EMAIL_DOMAIN`) it will allow to change the config from tests if needed. Otherwise the value is stuck in when the file is initially loaded.

app/custom_domain_validation.py
```diff
@@ -18,21 +25,46 @@ class DomainValidationResult:
...
+        self._partner_domains = partner_domains or PARTNER_DOMAINS
+        self._partner_domain_validation_prefixes = (
+            partner_domains_validation_prefixes or PARTNER_DOMAIN_VALIDATION_PREFIXES
+        )
 
-    def get_dkim_records(self) -> {str: str}:
-        """
-        Get a list of dkim records to set up. It will be
+    def get_ownership_verification_record(self, domain: CustomDomain) -> str:
+        prefix = "sl-verification"
```
acasajus, line 42: Maybe having the prefix as `sl` and then forcing the record to have `{prefix}-verification` so that we cannot have a prefix without the `verification` keyword? wdyt?

app/custom_domain_validation.py
```diff
@@ -18,21 +25,46 @@ class DomainValidationResult:
...
         """
-        return self._dkim_records
+
+        # By default use the default domain
+        dkim_domain = self.dkim_domain
+        if domain.partner_id is not None:
+            # Domain is from a partner. Retrieve the partner config and use that domain if exists
+            partner_domain = self._partner_domains.get(domain.partner_id)
+            if partner_domain is not None:
+                dkim_domain = partner_domain
```
acasajus, line 62:
```suggestion
            partner_domain = self._partner_domains.get(domain.partner_id, dkim_domain)
```
`get` has a default param to streamline this case

app/config.py
cquintana92: This method uses the `get_env_dict`, which is generic, and adapts it to reading partner config, which will always have an integer as the key, and the value as string.

templates/dashboard/domain_detail/dns.html
```diff
@@ -38,7 +38,7 @@ <h1 class="h2">{{ custom_domain.domain }}</h1>
             Value: <em data-toggle="tooltip"
      title="Click to copy"
      class="clipboard"
-     data-clipboard-text="{{ custom_domain.get_ownership_dns_txt_value() }}">{{ custom_domain.get_ownership_dns_txt_value() }}</em>
+     data-clipboard-text="{{ ownership_record }}">{{ ownership_record }}</em>
```
acasajus, line 41: 👍

app/custom_domain_validation.py
cquintana92: The config variables are only used as default if no other variables are passed, so in the tests we can send the full configuration and we don't need to use the config as a side-effect, we specify which values are used

app/custom_domain_validation.py
cquintana92: I have no strong feeling, I'll change it 👍

app/config.py
acasajus: ah true 👍

app/custom_domain_validation.py
acasajus: fair enough

cquintana92: Changed it

app/custom_domain_validation.py
cquintana92: Changed it

acasajus (approved):

PR #2262, opened 2024-10-14
chore: unify change alias mailboxes logic

cquintana92:


app/api/views/custom_domain.py
```diff
@@ -100,23 +102,14 @@ def update_custom_domain(custom_domain_id):
...
-
-            # first remove all existing domain-mailboxes links
-            DomainMailbox.filter_by(domain_id=custom_domain.id).delete()
-            Session.flush()
-
-            for mailbox in mailboxes:
-                DomainMailbox.create(domain_id=custom_domain.id, mailbox_id=mailbox.id)
-
+        result = set_custom_domain_mailboxes(user.id, custom_domain, mailbox_ids)
+        if result.success:
```
acasajus, line 106: To align with the rest of the utils, we could just check if the error is there since it's optional so we don't have to have an extra boolean.

cquintana92: Changed it and adapted the calling code

acasajus (approved):

PR #2276, opened 2024-10-20
Quarantine moved to the menu

br4n-d0n:
- Moved Quarantine for better accessibility because it was hidden under Settings
- Fixed some icons and rearranged to accommodate for Quarantine


nguyenkims:
When a bounce happens, SL sends a notification (notice the red dot on the top right) as email notification isn't reliable anymore.

image

PR #2331, opened 2024-11-26
Add lifetime in partner subscriptions

acasajus:


templates/dashboard/setting.html
```diff
@@ -26,7 +26,7 @@
     <div class="card">
       <div class="card-body">
         <div class="card-title mb-3">Current Plan</div>
-        {% if current_user.lifetime %}
+        {% if current_user.is_lifetime() %}
```
cquintana92, line 29: Where is this function implemented? I cannot find it

acasajus: good one. fixed.

cquintana92 (approved):

PR #2351, opened 2024-12-26
Add sudo to Docker commands in README

wikmor:
Ensure the commands work for all users, including those who haven’t configured Docker for non-sudo access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ If you already have a Postgres database in use, you can skip this section and ju
 Run a Postgres Docker container as your Postgres database server. Make sure to replace `myuser` and `mypassword` with something more secret.
 
 ```bash
-docker run -d \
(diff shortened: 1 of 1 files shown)


cquintana92:
Hi! Thanks for your suggestion.

However, I'm not too keen on adding `sudo` to all docker commands, as It's up to the user to set up proper docker usage/access permissions on their hosts.

If you have your used added to the docker group (as it's recommended, only to grant execution permissions on the `docker` CLI, and not on the whole system), then you don't need sudo to run docker. Root should only be needed to add yourself to the docker group.
In environments where the user executing these commands are not the system owner / root user, they wouldn't be able to run the commands with `sudo`, but the root owner / sysadmin can add the user to the `docker` group and the commands should work as they were written before.

PR #2359, opened 2025-01-16
feat: use rye as project manager

cquintana92:
Move from poetry to rye


acasajus (approved):

.github/workflows/main.yml
```diff
@@ -1,6 +1,12 @@
-name: Test and lint
```
nguyenkims, line 1: I think the previous name is more explicit.

.github/workflows/main.yml
```diff
@@ -1,6 +1,12 @@
...
+name: SimpleLogin actions
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
nguyenkims, line 9: does this mean only PR that target master will be run?

pyproject.toml
```diff
@@ -1,20 +1,100 @@
...
+keywords = ["email", "alias", "privacy", "oauth2", "openid"]
+packages = [
+    { include = "app/" },
+    { include = "migrations/" },
+]
+include = ["templates/*", "templates/**/*", "local_data/*.txt"]
+
+requires-python = "~=3.10"
+
+dependencies = [
```
nguyenkims, line 17: is there a way to move this one at the same location as `[tool.poetry.dependencies]` and `[tool.poetry]` to faciliate git diff comparison?

requirements.lock
```diff
@@ -0,0 +1,389 @@
+# generated by rye
```
nguyenkims, line 1: is there a way to check if all the pinned version here match the poetry one? To minimize issues coming from dependencies.

nguyenkims (commented):
Hey it'd be very useful if we can make sure the pinned dependencies have the same version as the poetry ones to reduce regression risk due to dependencies.

.github/workflows/main.yml
acasajus: This name is the prefix for the jobs. Before it was shown as:
`Test and lint / lint` and it was confusing to see the same job twice. You'll still see lint and test in the jobs.

requirements.lock
acasajus: We're checking the locked version deps from poetry to rye to make sure they are the same for our deps and the transitive deps also. We'll run this in staging for a while  and if looks ok for a couple days we'll run in a canary in prod also before rolling it out.

pyproject.toml
acasajus: The issue is that poetry uses their own non-standard way of defining dependencies. Rye follows the standard convention for python dependency management with the standard format.

acasajus: We basically copied the poetry ones and replaced the format to the new one.

PR #2367, opened 2025-01-28
feat: add request_id in log

cquintana92:
Log messages now include a request_id parameter across all the log messages of the same request in order to improve log searching

See the request_id after the PID
```
2025-01-28 14:53:09,827 - SL - DEBUG - 35908 - TaH0/qx5 - "/Users/carlos/code/simple-login/app/server.py:286" - after_request() -  - 127.0.0.1 GET /auth/login ImmutableMultiDict([]) 200, takes 0.07936620712280273
```

### DIFF
--- a/app/request_utils.py
+++ b/app/request_utils.py
@@ -1,5 +1,6 @@
-def generate_request_id() -> str:
-    from random import randbytes
-    from base64 import b64encode
+from random import randbytes
+from base64 import b64encode
 
+
+def generate_request_id() -> str:
     return b64encode(randbytes(6)).decode()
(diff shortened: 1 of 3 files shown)


app/request_utils.py
```diff
@@ -0,0 +1,5 @@
+def generate_request_id() -> str:
+    from random import randbytes
+    from base64 import b64encode
```
acasajus, line 3: Why not top level import?

acasajus (approved):

app/request_utils.py
cquintana92: Total overlook, probably mistake rearranging lines. Fixed it!

PR #2380, opened 2025-02-05
chore: offer version in newrelic events

cquintana92:

- A new `VERSION` variable in the `app/build_info.py` file that will contain the tag name/branch name when generating a docker image on the CI
- A new `/version` endpoint that will return the version
- A new newrelic event that is sent on every request, containing the service and the version, so we can better monitor which version is present on each node


acasajus:
We should add it to the email_handler also

acasajus (approved):

PR #2384, opened 2025-02-06
Ability to block domains when receiving emails #2

tozo:
Sorry, I had to recreate this PR.

This PR adds the functionality to block domains so that the user won't receive emails no matter which email was used from that domain.
It's mainly to prevent spam emails coming from the same domain but from different email addresses.

Since this feature is not tied to a specific alias/custom domain I added it to the setting page.
Let me know if there is a better place for this feature or if anything else needs to be changed.

Note: This feature was discussed here: [#1344](https://github.com/simple-login/app/discussions/1344)

Couple of screenshots:

Default view:
blocked_domains_default_view

Page once domains are blocked:
blocked_domains


app/models.py
```diff
@@ -1213,6 +1215,15 @@ def has_used_alias_from_partner(self) -> bool:
             > 0
         )
 
+    def is_domain_blocked(self, domain: str) -> bool:
+        """checks whether the provided domain is blocked for a given user"""
+        domain_names = []
+
+        for blocked_domain in BlockedDomain.filter_by(user_id=self.id):
+            domain_names.append(blocked_domain.domain)
```
acasajus, line 1223: Can you add the domain to the filter directly? This you skip a loop and the filtering happens DB side directly.

app/models.py
```diff
@@ -3106,6 +3117,20 @@ class DomainMailbox(Base, ModelMixin):
...
+
+    __tablename__ = "blocked_domain"
+
+    __table_args__ = (
+        sa.UniqueConstraint("user_id", "domain", name="uq_blocked_domain"),
+    )
+
+    domain = sa.Column(sa.String(128), nullable=False)
+
+    user_id = sa.Column(sa.ForeignKey(User.id, ondelete="cascade"), nullable=False)
```
acasajus, line 3131: It's missing a composite index with user_id and domain to avoid full table scans.

app/custom_domain_utils.py
```diff
@@ -89,12 +97,14 @@ def sanitize_domain(domain: str) -> str:
     return new_domain
 
 
-def can_domain_be_used(user: User, domain: str) -> Optional[CannotUseDomainReason]:
+def can_domain_be_used(
```
acasajus, line 100: Although code reuse is good I'd split this into three functions:
 - one that has all the logic without taking into account custom domain or blocked domain
 - one that calls the previous and checks also for duplicate custom domains 
 - one that calls the first and checks that this user doesn't have any duplicate blocked domain

app/models.py
tozo: In the previous line I have this constraint:
```python
    __table_args__ = (
        sa.UniqueConstraint("user_id", "domain", name="uq_blocked_domain"),
    )
```

This should create that composite index. I checked in my environment and I see that index created automatically.

But let me know if I misunderstand something.

tozo:
@aksolotl222 If I understand you correctly this PR does what you describe.

You can block complete domains so no matter what email addresses they use on that domain they all will be blocked.

tozo:
Hi @acasajus 

Could you let me know if there is any other changes that need to be done?
I think this would be a great feature for many people.

alik-090:
Guys, can you also add ability to
1. block sender (ex example@example.com) from sending emails to ANY alias (non only domain)
2. Separate in settings behavior. So I I disable alias - user won't see that email rejected. And when I BLOCK sender he should get error that email doesn't exist. Now there is such feature but it is combined with block and disable alias. This is bad because some services, when see "email does not exist" will not try to send email even after I turn on alias. But when I BLOCK sender, I think sender should see "email does not exist". This will prevent most soammers to use different adress/domain because they will think there is no such address.

What do you think?

tozo:
@alik-090 

For your first point, this PR somewhat covers that, since this blocks the whole sender domain (not just a specific email). But I can see that in some situations it might be better to block only some sender emails.

Either way I would implement those features in a separate PR, so that we keep this as simple as possible

erik65679:
@acasajus @cquintana92 @nguyenkims 

Any progress on this? It is really neat feature. I am waiting for it.

app/dashboard/views/setting.py
```diff
@@ -284,6 +292,33 @@ def setting():
...
+            Session.commit()
+            flash(f"Added blocked domain [{domain}]", "success")
+            return redirect(url_for("dashboard.setting"))
+        elif request.form.get("form-name") == "blocked-domains-remove":
+            domain_id = request.form.get("domain_id")
+            domain_name = request.form.get("domain_name")
+
+            domain = BlockedDomain.get(domain_id)
+
+            BlockedDomain.delete(domain.id)
```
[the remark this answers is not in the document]
tozo: I now validate if the current user is the owner of the domain before deleting it, let me know if that looks ok
Other option is to have a call like this:

```python
deleted = (
    BlockedDomain.filter_by(
        id=domain_id_int,
        user_id=current_user.id,
    ).delete()
)
```

tozo:
all the notes should be fixed now, let me know if anything else needs to be fixed/changed

PR #2392, opened 2025-02-20
Send plan change events on subscription creation

acasajus:


acasajus:
It's done in the subscription webhook

PR #2468, opened 2025-05-14
[WIP][DONT MERGE] feat: added abuser utils for main abuser (un)archiving flows.

virtyaluk:


app/models.py
```diff
@@ -357,6 +357,31 @@ class Fido(Base, ModelMixin):
...
+    encrypted_bundle = sa.Column(sa.LargeBinary(), nullable=False)
+
+    __table_args__ = (sa.Index("ix_abuser_data_id", "id"),)
+
+
+class AbuserLookup(Base, ModelMixin):
+    __tablename__ = "abuser_lookup"
+
+    hashed_address = sa.Column(sa.String(), nullable=False)
+    blob_id = sa.Column(
```
acasajus, line 373:
```suggestion
    abuser_data_id = sa.Column(
```
We always use `table_name_id` as foreign keys to make it clear the table they reference.

app/models.py
```diff
@@ -357,6 +357,31 @@ class Fido(Base, ModelMixin):
...
+        sa.ForeignKey("abuser_data.id", ondelete="cascade"), nullable=False
+    )
+    encrypted_k_bundle = sa.Column(sa.LargeBinary(), nullable=False)
```
acasajus, line 376:
```suggestion
    abuser_data_key = sa.Column(sa.LargeBinary(), nullable=False)
```
Naming is hard.. I'd suggest this name to make it clear that this is the key to open the `abuser_data`

app/abuser_utils.py
```diff
@@ -0,0 +1,182 @@
...
+from hashlib import sha256
+from typing import List, Dict
+
+from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+
+from app.db import Session
+from app.models import User, Alias, Mailbox, AbuserData, AbuserLookup
+
+
+def check_if_abuser_email(new_address: str) -> bool:
```
acasajus, line 12:
```suggestion
def check_if_abuser_email(new_address: str) -> AbuserLookup|None:
```

app/models.py
acasajus: Also we need to support the same address having more than one abuse report since an email can exist as a mailbox for several users.

app/abuser_utils.py
```diff
@@ -0,0 +1,197 @@
...
+def archive_abusive_user(user: User) -> None:
+    """
+    Archive the given abusive user's data and update blocklist/lookup tables.
+    """
+    if not user.email:
+        raise ValueError(f"User ID {user.id} must have a primary email to be archived.")
+
+    try:
+        primary_email: str = user.email
+        aliases: List[Alias] = Alias.filter_by(user_id=user.id).all()
```
acasajus, line 58: User may have tons of alias. I'd recommend doing 
```suggestion
        aliases: List[Alias] = Alias.filter_by(user_id=user.id).enable_eagerloads(False).yield_per(500).all()
```
This way it will page the queries to the db behind the scenes

app/abuser_utils.py
```diff
@@ -0,0 +1,197 @@
...
+                    "address": mailbox.email,
+                    "created_at": mailbox.created_at.isoformat()
+                    if mailbox.created_at
+                    else None,
+                }
+                for mailbox in mailboxes
+            ],
+        }
+        bundle_json_bytes = json.dumps(bundle, sort_keys=True).encode("utf-8")
+        derived_enc_key = _derive_encryption_key(config.MASTER_ENC_KEY, user.id)
```
acasajus, line 84: We need to:
 - Generate random encryption key for AES_GCM
 - Encrypt bundle with this key
 - Encrypt the key for each email using the HKDF and store it in abuserLookup

app/abuser_utils.py
```diff
@@ -0,0 +1,197 @@
...
+    check_hmac = hmac.new(
+        mac_key_bytes, new_address.encode("utf-8"), sha256
+    ).hexdigest()
+
+    return (
+        AbuserLookup.filter(AbuserLookup.hashed_address == check_hmac).limit(1).first()
+    )
+
+
+def archive_abusive_user(user: User) -> None:
```
acasajus, line 49:
```suggestion
def mark_as_abusive_user(user: User) -> None:
```

app/abuser_utils.py
```diff
@@ -0,0 +1,197 @@
...
+
+            seen_identifiers.add(identifier_address)
+
+            identifier_hmac = hmac.new(
+                mac_key_bytes, identifier_address.lower().encode("utf-8"), sha256
+            ).hexdigest()
+            abuser_lookup_entry = AbuserLookup(
+                hashed_address=identifier_hmac, abuser_data_id=blob_id
+            )
+
```
acasajus, line 117: We're missing a `user.disabled=True` to disable the user properly

app/abuser_utils.py
```diff
@@ -0,0 +1,197 @@
...
+            Session.add(abuser_lookup_entry)
+
+        Session.commit()
+    except Exception:
+        Session.rollback()
+        raise
+
+
+def unarchive_abusive_user(user_id: int) -> None:
```
acasajus, line 126:
```suggestion
def unmark_as_abusive_user(user_id: int) -> None:
```

app/abuser_utils.py
```diff
@@ -0,0 +1,197 @@
...
+    Fully remove abuser archive and lookup data for a given user_id.
+    This reverses the effects of archive_abusive_user().
+    """
+    abuser_data_entry = AbuserData.filter_by(user_id=user_id).first()
+
+    if not abuser_data_entry:
+        LOG.e("No abuser data found for user %s", user_id)
+        return
+
+    Session.delete(abuser_data_entry)
```
acasajus, line 137: We should also remove the `AbuserLookup` entries that match this abuser data

virtyaluk: Handled by cascade removal. I have dedicated test case just for that `test_abuser_data_deletion_cascades_to_lookup`.

app/abuser_utils.py
virtyaluk: Shouldn't the disabling/enabling switch happen in the upstream flow that actually responsible for marking the user as an abuser? To me, the abuser_utils looks more like a utility side flow to just store the encrypted bundle.

app/abuser_utils.py
acasajus: 👍

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        raise ValueError(
+            "Identifier address must be a non-empty string for key derivation."
+        )
+
+    normalized_identifier = identifier_address.lower()
+    hkdf_info = (constants.HKDF_INFO_TEMPLATE % normalized_identifier).encode("utf-8")
+    hkdf = HKDF(
+        algorithm=crypto_hashes.SHA256(),
+        length=32,
+        salt=None,
```
acasajus, line 31: We should use 32 bytes of random data. We can add that to the config.

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+    check_hmac = hmac.new(
+        mac_key_bytes, normalized_address.encode("utf-8"), sha256
+    ).hexdigest()
+
+    return (
+        AbuserLookup.filter(AbuserLookup.hashed_address == check_hmac).limit(1).first()
+    )
+
+
+def mark_as_abusive_user(user: User, actor: User) -> None:
```
acasajus, line 54:
```suggestion
def mark_user_as_abuser(abuse_user: User: note: str, admin: User|None) -> None:
    user.disabled =True
    _emit_abuser_audit_log(abuse_user, action: AbuserLogAction.MarkAbuser, message = note, admin=admin) 
    _store_abuse_data_(abuse_user)

def _store_abuse_data(user: User, actor: User) -> None:
```

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+
+            k_identifier_derived = _derive_key_for_identifier(
+                master_key_bytes, normalized_identifier
+            )
+            aesgcm_key_enc = AESGCM(k_identifier_derived)
+            nonce_key_encryption = secrets.token_bytes(12)
+            encrypted_k_bundle_for_this_identifier = (
+                nonce_key_encryption
+                + aesgcm_key_enc.encrypt(
+                    nonce_key_encryption, k_bundle_random, aad_bytes
```
acasajus, line 142:
```suggestion
                    nonce_key_encryption, k_bundle_random, f"{aad_bytes}.bundle"
```
So the tag is different for the bundle as from the keys

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+                if mailbox.email
+            ],
+        }
+        bundle_json_bytes = json.dumps(bundle, sort_keys=True).encode("utf-8")
+        k_bundle_random = AESGCM.generate_key(bit_length=256)
+        aesgcm_bundle_enc = AESGCM(k_bundle_random)
+        nonce_bundle = secrets.token_bytes(12)
+        aad_bytes = constants.AEAD_AAD_DATA.encode("utf-8")
+        encrypted_bundle_data = nonce_bundle + aesgcm_bundle_enc.encrypt(
+            nonce_bundle, bundle_json_bytes, aad_bytes
```
acasajus, line 101:
```suggestion
            nonce_bundle, bundle_json_bytes, f"{aad_bytes}.bundle"
```

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        LOG.e("No abuser data found for user %s", user_id)
+        return
+
+    Session.delete(abuser_data_entry)
+    Session.commit()
+    emit_abuser_audit_log(
+        user=actor,
+        action=AbuserAuditLogAction.UnmarkAbuser,
+        message=f"An user {user_id} was unmarked as abuser.",
+    )
```
acasajus, line 182:
```suggestion
def unmark_as_abusive_user(user_id: int, actor: User) -> None:
    """
    Fully remove abuser archive and lookup data for a given user_id.
    This reverses the effects of archive_abusive_user().
    """
    abuser_data_entry = AbuserData.filter_by(user_id=user_id).first()

    if not abuser_data_entry:
        LOG.e("No abuser data found for user %s", user_id)
        return

    Session.delete(abuser_data_entry)
    emit_abuser_audit_log(
        user=user_id,
        admin = actor
        action=AbuserAuditLogAction.UnmarkAbuser,
        message=f"An user {user_id} was unmarked as abuser.",
    )
    Session.commit()
```
We want to store in the audit log the user that was removed as abuser and the admin that did it

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+            )
+            continue
+
+        encrypted_main_bundle_data = abuser_data_record.encrypted_bundle
+
+        try:
+            nonce_k_decryption = encrypted_k_bundle_from_entry[:12]
+            ciphertext_k_decryption = encrypted_k_bundle_from_entry[12:]
+            plaintext_k_bundle = aesgcm_key_dec.decrypt(
+                nonce_k_decryption, ciphertext_k_decryption, aad_bytes
```
acasajus, line 240:
```suggestion
                nonce_k_decryption, ciphertext_k_decryption, f"{constant.AEAD_AAD_DATA}.key".encode("utf-8"))
```

app/abuser_utils.py
```diff
@@ -0,0 +1,272 @@
...
+            )
+            aesgcm_bundle_dec = AESGCM(plaintext_k_bundle)
+            nonce_main_bundle = encrypted_main_bundle_data[:12]
+            ciphertext_main_bundle = encrypted_main_bundle_data[12:]
+            decrypted_bundle_json_bytes = aesgcm_bundle_dec.decrypt(
+                nonce_main_bundle, ciphertext_main_bundle, aad_bytes
```
acasajus, line 246:
```suggestion
                nonce_main_bundle, ciphertext_main_bundle, f"{constant.AEAD_AAD_DATA}.bundle".encode("utf-8"))
```

app/api/views/auth.py
```diff
@@ -113,6 +115,15 @@ def auth_register():
...
 
+    if check_if_abuser_email(email):
+        # TODO: do I need to call RegisterEvent in here?
+        emit_abuser_audit_log(
+            None,
+            AbuserAuditLogAction.CreateUser,
+            "An attempt to register a previously banned email",
+        )
+        return jsonify(error=f"cannot use {email} as it was previously banned"), 400
+
```
acasajus, line 126:
```suggestion
    if check_if_abuser_email(email):
        LOG.warn("User with email {email} that was marked as abuser tried to register again")
        return jsonify(error=f"cannot use {email} as an email to register"), 400
```
The less info we give abusers the better :)

app/models.py
```diff
@@ -357,6 +357,49 @@ class Fido(Base, ModelMixin):
...
+class AbuserAuditLog(Base, ModelMixin):
+    __tablename__ = "abuser_audit_log"
+
+    user_id = sa.Column(sa.Integer, nullable=False)
+    action = sa.Column(sa.String(255), nullable=False)
+    message = sa.Column(sa.Text, default=None, nullable=True)
+
+    __table_args__ = (
+        sa.Index("ix_abuser_audit_log_user_id", "user_id"),
+        sa.Index("ix_abuser_audit_log_created_at", "created_at"),
```
acasajus, line 399: Why do we need an index by `created_at` ?

app/models.py
```diff
@@ -357,6 +357,49 @@ class Fido(Base, ModelMixin):
...
+        index=True,
+    )
+    bundle_k = sa.Column(sa.LargeBinary(), nullable=False)
+
+    __table_args__ = (
+        sa.Index(
+            "ix_abuser_lookup_hashed_address_abuser_data_id",
+            "hashed_address",
+            "abuser_data_id",
+        ),
```
acasajus, line 386: Shouldn't they be two independent indexes? Either we query by hashed address or by abuser_data_id, but afaics we won't query by both of them at the same time.

app/api/views/auth.py
acasajus: We have to also do it in `app/auth/views/register.py` line 74

app/abuser_utils.py
```diff
@@ -0,0 +1,290 @@
...
+        Session.delete(abuser_data_entry)
+
+    user = User.get(user_id)
+    user.disabled = False
+
+    emit_abuser_audit_log(
+        user_id=user.id,
+        admin_id=admin_id,
+        action=AbuserAuditLogAction.UnmarkAbuser,
+        message=f"An user {user_id} was unmarked as abuser.",
```
acasajus, line 194: Same here, message should be something that the admin can enter to explain why.

app/config.py
```diff
@@ -686,3 +687,7 @@ def read_partner_dict(var: str) -> dict[int, str]:
 ALIAS_TRASH_DAYS = int(os.environ.get("ALIAS_TRASH_DAYS", 30))
 ALLOWED_OAUTH_SCHEMES = get_env_csv("ALLOWED_OAUTH_SCHEMES", "auth.simplelogin,https")
 MAX_EMAIL_FORWARD_RECIPIENTS = int(os.environ.get("MAX_EMAIL_FORWARD_RECIPIENTS", 30))
+
+MASTER_ENC_KEY = bytes.fromhex(os.environ.get("MASTER_ENC_KEY_HEX"))
+MAC_KEY = bytes.fromhex(os.environ.get("MAC_KEY_HEX"))
+ABUSER_HKDF_SALT = base64.b64decode(os.environ.get("ABUSER_HKDF_SALT"))
```
acasajus, line 693: These three can be set to nothing for people that self host. If they aren't set to anything we should a default value derived from the flask secret for instance. This way people self hosting won't have their installations failing.

We could do something like 
```python
def decode_hex_data(key: string, default: bytes) -> bytes:
  data = os.environ.get(key)
  if not data:  
   return default
  else;
    return bytes.fromhex(data)

MASTER_ENC_KEY = decode_hex_data("MASTER_ENC_KEY_HEX", (FLASK_SECRET + 'enckey').encode("UTF-8+")
MAC_KEY = decode_hex_data("MAC_KEY", (FLASK_SECRET + 'mackey').encode("UTF-8+")
ABUSER_HKDF_SALT = decode_hex_data("ABUSER_HKDF_SALT", (FLASK_SECRET + 'absalt').encode("UTF-8+")
```

acasajus (approved):

PR #2472, opened 2025-05-19
feat: adds abuser lookup admin view.

virtyaluk:
image


acasajus:
Nice start! I would:
 - format a bit the data so it's easier to read for a human. 
 - Look also for the user object and if it exists show it with a link to the email_search page with the user email in the query string so admins can see the user data directly.
 The rest looks 👍

oneshot/backfill_abuser_data.py
```diff
@@ -0,0 +1,41 @@
...
+
+parser = argparse.ArgumentParser(
+    prog="Backfill abuser data",
+    description="Iterate over disabled users and generate abuser bundle",
+)
+args = parser.parse_args()
+
+print("I'm going to generate abuser bundle for disabled users.")
+
+disabled_users: List[User] = Session.query(User).filter(User.disabled == True).all()  # noqa: E712
```
acasajus, line 17: We need to do this by batches with userId > n and userId < m because we don't have an index on disabled and this will trigger a full table scan

oneshot/backfill_abuser_data.py
```diff
@@ -0,0 +1,41 @@
...
+archived_users: int = 0
+
+for disabled_user in disabled_users:
+    if not disabled_user.email:
+        print(f"Disabled user {disabled_user.id} has no email address. Skipping...")
+        continue
+
+    if check_if_abuser_email(disabled_user.email):
```
acasajus, line 25: thinking about it.. this may match the mailbox from another user. We can check the user_id in the AbuserData model to make sure.

Issue #916, opened 2022-04-21
 Docker Image Outdated?


viveksjain:
Thanks for this great project. I guess this is pretty much same request as #364, it looks like docker image is over a year old. Is it possible to publish a new release? Docker-compose would be useful as well for quickly getting started but I guess that is tracked in #355.

nguyenkims:
@viveksjain we'll publish the new image in the next coming months. Our calendar is unfortunately packed at the moment but rest assured that releasing new image is in our todo list :).

henokv:
@nguyenkims does the site simple checkout a specific commit and runs on this version instead of docker images?
Would it be possible to document somewhere which commit the site is currently working on?
I assume that this version would be stable enough for most use-cases while still giving the new functionality.

nguyenkims:
@henokv We deploy new versions quite often, you can see the current version running at https://app.simplelogin.io/git

And yes you can run the latest version on your server. This version isn't tested in a self hosted setup though so run it at your own risk :).

Heziode:
The latest edit in changelog goes back to a year (3.4.0).
You should consider shorter iteration cycles, especially when publishing patches (if necessary).

You could setup a CI/CD, to release at least a `simplelogin/app-dev` on docker hub, for those who are adventurous 😉

pocki:
Any news about a new up-to-date docker image? 
Or should now every selfhoster build it's own docker image?

alpha-tango-kilo:
There's `simplelogin/app-ci` on Docker Hub which appears to mirror GitHub tags, but I doubt it's 'officially' supported. I haven't tried it myself yet - no migration instructions etc.

pocki:
Thanks @alpha-tango-kilo 

I used the upgrade infos from here [4.6.2-beta/docs/upgrade.md](https://github.com/simple-login/app/blob/4.6.2-beta/docs/upgrade.md)  but used the `simplelogin/app-ci:v4.7.3` images instead

At the upgrade and also at starting the containers I am getting some warnings, but it seems everything is working.
```
Paddle param not set
Upload files to local dir
Symbol 'mailbox_or_url_list' is unreachable
Symbol 'delim' is unreachable
Symbol 'mailbox_or_url' is unreachable
Symbol 'mailbox' is unreachable
Symbol 'url' is unreachable
Symbol 'angle_addr' is unreachable
Symbol 'name_addr' is unreachable
Symbol 'phrase' is unreachable
[...]

PR #2491, opened 2025-06-07
Change timeout of app's worker

ZnqbuZ:
To solve https://github.com/simple-login/app/issues/916 and https://github.com/simple-login/app/discussions/2107, thanks to [@loeff](https://github.com/loeffelpan)


acasajus:
Hey, this means that you have something in your setup that's not right. 15 seconds should be more than enough to boot the app. Maybe your postgres has some issues to. Id check into that.

PR #2568, opened 2025-10-01
Handle None case for default_factory in config:---->The error is caused because your sl_getenv function is being called without a proper default_factory.

Shubhi-Sri:
Return None if default_factory is None.


acasajus:
It's only used in the initial config. No need to fix it in the function. But thanks for your contribution.

PR #2583, opened 2025-10-24
user can set a display name for an autocreate rule

nguyenkims:
For https://github.com/simple-login/app/discussions/821

CleanShot 2025-10-24 at 16 41 55@2x

image

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2710,7 +2710,7 @@ class AutoCreateRule(Base, ModelMixin):
     display_name = sa.Column(
         sa.String(128),
         nullable=True,
-        server_default="",
+        server_default=None,
     )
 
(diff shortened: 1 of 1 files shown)


migrations/versions/2025_111315_f3d65fe0b5b4_add_auto_create_rule_display_name.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'f3d65fe0b5b4'
+down_revision = '3ffdea52697d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('auto_create_rule', sa.Column('display_name', sa.String(length=128), nullable=True))
```
cquintana92: Same here

PR #2600, opened 2025-11-09
Webauthn fixes

a2kolbasov:


nguyenkims:
Thanks! Do you have an example of a situation where the bug happens and this MR solves it?

a2kolbasov:
For example, this is evident in the KeePassXC-browser password manager. It has a setting that allows you not to return the request to the browser if the keys are not found or the request is canceled by the user. This is necessary if you do not use hardware or smartphone keys and do not want to see a browser suggestion to insert a key.

KeePassXC-Browser passkeys fallback settings

---

Login attempt:

[keepass-browser_example.webm](https://github.com/user-attachments/assets/97271a7e-f053-4186-b894-61f46f588c24)

1-st error in console:
```console
Error when trying to get credential: DOMException: CredentialContainer request is not allowed.
    verifyKey https://app.simplelogin.io/auth/fido:158
    <anonymous> https://app.simplelogin.io/auth/fido:174
fido:165:28
Init sentry fido:186:13
init Analytics bundle.min.js:2:12178
[Debug keepassxc-browser.js:230] KeePassXC-Browser - Input fields found: global.js:173:13
Array []
```

The code is executed before the extension can intercept the request. The code already has `$('document').ready`. But the function is not passed as a callback, it is called immediately.

2-nd error:

```console
[Debug passkeys.js:136] KeePassXC-Browser - Passkey request global.js:173:13
⁞
Uncaught (in promise) TypeError: can't access property "response", newAssertion is null
    transformAssertionForServer https://app.simplelogin.io/static/assets/js/vendors/webauthn.js:123
    verifyKey https://app.simplelogin.io/auth/fido:168
    onclick https://app.simplelogin.io/auth/fido:1
webauthn.js:123:35
```

After that, the user does not have the opportunity to make a repeat request (without refreshing the page) because the button remains inactive due to an uncaught error.

a2kolbasov:
And here is an attempt to log in to github:

Sign in to GitHub

```console
20:14:53.278 Uncaught (in promise) TypeError: can't access property "toJSON", t is null
    E webauthn-json.browser-ponyfill.js:211
    prompt webauthn-get-element.ts:196
    handleWebauthnSubtle webauthn-get-element.ts:79
    prompt webauthn-subtle-element.ts:12
    l bind.js:73
```

Github's code:
```js
// src/webauthn-json/browser-ponyfill.ts
async function create(options) {
  const response = await navigator.credentials.create(
    options
  );
  response.toJSON = () => createResponseToJSON(response);
  return response;
}
async function get(options) {
  const response = await navigator.credentials.get(
    options
  );
  response.toJSON = () => getResponseToJSON(response);
  return response;
}
```
The conversion of the response takes place before the 'catch` block.

---
---

So, alternatively, it is possible just to omit the catch block below `$('#formRegisterKey').submit();` without checking for `null`.

https://github.com/simple-login/app/blob/300aeca0af28de0fc92ae1bf804e4227b2fe24bd/templates/auth/fido.html#L46-L69

a2kolbasov:
Sorry, I had to change the first commit due to an error.

If I pass an asynchronous function to `$.ready`, it may not start (https://stackoverflow.com/questions/69984178/async-await-with-jquery-document-ready). Now I'm passing this as a synchronous function.

In addition, `$('document').ready` call is marked as deprecated (https://jquery.com/upgrade-guide/3.0/#deprecated-document-ready-handlers-other-than-jquery-function)

acasajus:
Thanks for the fix 👍

PR #2641, opened 2026-01-13
feat: use Rust PGP

cquintana92:
Use a new compiled dependency in order to perform all the PGP operations in memory, replacing the invocations to the `pgp` command line tool.

### DIFF
--- a/app/pgp_utils.py
+++ b/app/pgp_utils.py
@@ -54,7 +54,6 @@ def _record_pgp_metric(
         {
             "implementation": implementation,
             "success": "success" if success else "fail",
-            "duration_ms": elapsed * 1000,
             "is_fallback": "fallback" if is_fallback else "direct",
         },
     )
@@ -253,6 +252,10 @@ def sign_data(
             # Ensure the signing key is loaded
             if PGP_SENDER_PRIVATE_KEY:
                 ctx.set_signing_key(PGP_SENDER_PRIVATE_KEY)
+            else:
+                raise Exception(
+                    "Tried to sign data but PGP_SENDER_PRIVATE_KEY is not set"
+                )
             if isinstance(data, str):
                 data = data.encode("utf-8")
             result = ctx.sign_detached(data)


app/pgp_utils.py
```diff
@@ -19,38 +22,100 @@ class PGPException(Exception):
...
+    elapsed: float,
+    is_fallback: bool = False,
+):
+    """Record PGP operation metrics to NewRelic."""
+    # Record timing metric
+    metric_suffix = f"_{implementation}"
+    if is_fallback:
+        metric_suffix = f"_fallback{metric_suffix}"
+    newrelic.agent.record_custom_metric(
+        f"Custom/pgp_{operation}_time{metric_suffix}", elapsed
```
acasajus:
```suggestion
        f"Custom/pgp_{operation}{metric_suffix}_time", elapsed
```

app/pgp_utils.py
```diff
@@ -19,38 +22,100 @@ class PGPException(Exception):
...
+    event_name = f"Pgp{operation.title()}"
+    if is_fallback:
+        event_name += "Fallback"
+    newrelic.agent.record_custom_event(
+        event_name,
+        {
+            "implementation": implementation,
+            "success": success,
+            "duration_ms": elapsed * 1000,
+            "is_fallback": is_fallback,
```
acasajus:
```suggestion
            "is_fallback": "fallback" if is_fallback else "direct",
```
Could not come up with a good name for not fallback :P

app/pgp_utils.py
```diff
@@ -59,60 +124,172 @@ def hard_exit():
...
+    start_time = time.time()
+
+    try:
+        if force_use_rust or USE_RUST_PGP:
+            # Ensure the signing key is loaded
+            if PGP_SENDER_PRIVATE_KEY:
+                ctx.set_signing_key(PGP_SENDER_PRIVATE_KEY)
+            if isinstance(data, str):
+                data = data.encode("utf-8")
+            result = ctx.sign_detached(data)
```
acasajus: Always uses the `set_signing_key` for signature? if it's not set what happens?

app/pgp_utils.py
```diff
@@ -19,38 +22,100 @@ class PGPException(Exception):
...
+        metric_suffix = f"_fallback{metric_suffix}"
+    newrelic.agent.record_custom_metric(
+        f"Custom/pgp_{operation}_time{metric_suffix}", elapsed
+    )
+
+    # Record count metrics for easier aggregation
+    status = "success" if success else "failure"
+    newrelic.agent.record_custom_metric(
+        f"Custom/pgp_{operation}_count_{implementation}_{status}", 1
+    )
```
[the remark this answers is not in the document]
cquintana92: Sure, removed the metric

app/pgp_utils.py
```diff
@@ -59,60 +124,172 @@ def hard_exit():
...
+                if mailbox:
+                    LOG.d("(re-)load public key for %s", mailbox)
+                    ctx.load_public_key(mailbox.pgp_public_key)
+
+                contact = Contact.get_by(pgp_finger_print=fingerprint)
+                if contact:
+                    LOG.d("(re-)load public key for %s", contact)
+                    ctx.load_public_key(contact.pgp_public_key)
+
+            result = ctx.encrypt(data_bytes, fingerprint)
```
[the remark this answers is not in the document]
cquintana92: In the block above, it's checking whether the PgpContext has the key for the fingerprint, and if it doesn't, it gets loaded into the context.

That being said, the first thing the library does when calling the `encrypt` method is retrieving the public key by its fingerprint, and if it doesn't have it loaded, it returns an error.

PR #2677, opened 2026-02-11
Fix: cbor2 build failure on Python 3.12 (Clean/Local build)

lobovkin:
**The Problem**
Clean installations and local builds fail with ModuleNotFoundError: No module named 'pkg_resources' when installing cbor2==5.2.0.

**Technical Root Cause**
- Build from Source: Legacy cbor2 versions require setuptools (which provides pkg_resources) to compile from source (sdist).
- Python 3.12+: Isolated build environments in Python 3.12 no longer include setuptools by default.
- Why it worked before: The original CI environment relied on cached binaries (wheels). Compilation was never actually triggered. On any fresh environment (local or new CI) where the cache is empty, the build process starts from scratch and crashes.

**The Fix**
Updated uv.lock to bump cbor2 to v5.6.5+. Modern versions provide pre-compiled Python 3.12 wheels. This allows uv to install the package directly without local compilation, fixing both CI and local development setups.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -258,9 +258,19 @@ wheels = [
 
 [[package]]
 name = "cbor2"
-version = "5.2.0"
+version = "5.8.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


nguyenkims:
thanks, I created this MR to fix this cbor2 issue and also pyre2
https://github.com/simple-login/app/pull/2678
I think we can close this one.

PR #2689, opened 2026-03-15
Add global sender blacklist (regex) with admin UI and SMTP blocking

chrisblech:
Adds a global sender blacklist to SimpleLogin: Admin can block sender addresses (or whole domains / TLDs) by regex pattern.

- New DB table `global_sender_blacklist` (pattern regex + enabled + comment)
- Admin UI (Flask-Admin) to manage patterns
- Inbound SMTP handling: blocks senders matching the blacklist, using the same behavior as a blocked Contact (2xx vs 5xx depending on user.block_behaviour)
- Uses re2-backed regex search to avoid catastrophic backtracking

Example patterns:
- `@spamdomain\.com$`
- `^no-?reply@.*`

Migration and test are included.


chrisblech:
Hmm - while testing this for two days, I realized the feature can be improved. Will open another PR when ready.

PR #2720, opened 2026-05-05
Disable custom domains if MX records are disabled

acasajus:
Currently we just notify owners of domains that the MX records are not properly set. Now they will be disabled and aliases won't be forwarded if the domain is not properly configured.


cquintana92 (approved):

Slevin90:
How does this work on a self-hosted server?
Could id couse any issues? when reverifying DNS settings?

PR #2740, opened 2026-06-04
feat: maintenance mode for web

cquintana92:
This PR implements a "Maintenance Mode" for the webapp, where all requests (except static and healthcheck) return 503 and show a maintenance page. It does it by extracting logic from `server.py` and moving it into `simplelogin_app.py` alongside a `maintenance_app.py`, making `server.py` determine whether it's in maintenance mode based on the new `config.py` directive. If in maintenance mode, the web app doesn't even try to connect to the database, so it always works regardless of the connection status.

Maintenance mode for `email_handler.py` is not implemented in this PR as it requires some bigger changes.


maintenance_app.py
```diff
@@ -0,0 +1,45 @@
...
+    app = Flask(__name__)
+    app.wsgi_app = ProxyFix(app.wsgi_app, x_for=1, x_host=1)
+
+    @app.route("/health", methods=["GET"])
+    def healthcheck():
+        return "success", 200
+
+    @app.before_request
+    def maintenance():
+        if (
```
oliver-gordon, line 19: nit: you could convert this to an `if not` and return the jsonify error below and then everything would 503 except these no?

oliver-gordon (approved):

simplelogin_app.py
```diff
@@ -0,0 +1,513 @@
...
+            },
+            "ignore": ["^/static/.*", "/git", "/exception", "/health"],
+        }
+        flask_profiler.init_app(app)
+
+    # enable CORS on /api endpoints
+    CORS(app, resources={r"/api/*": {"origins": "*"}})
+
+    @app.before_request
+    def check_maintenance_mode():
```
acasajus: If the maintenance mode creates a different app why is this check also here? Or is it a left over copypasta?

cquintana92: Nice catch, it's a leftover. Removed 👍

acasajus (approved):

PR #2773, opened 2026-07-27
feat: invalidate email change token on password change

cquintana92:


app/auth/views/reset_password.py
```diff
@@ -71,6 +71,8 @@ def reset_password():
 
         # remove all reset password codes
         ResetPasswordCode.filter_by(user_id=user.id).delete()
+        # invalidate pending email change tokens
```
acasajus, line 74: Nit: Since both behaviours are the same we could move it to a common function. Feel free to ignore for now.

acasajus (approved):

PR #2776, opened 2026-07-30
allow "-" and "_" as directory alias separators

ukrcat227:
This is critically important, because more and more services reject existing "/" or "+" separators as invalid while "-" or "_" never rejected


ualinuxoid:
@acasajus or @nguyenkims can you please take a look?

PR #2785, opened 2026-08-04
Clarified CONTRIBUTING instructions

springcomp:
This pull request clarifies the `CONTRIBUTING.md` documentation for local development and testing.


CONTRIBUTING.md
```diff
@@ -62,26 +62,50 @@ To install it in your development environment.
...
 
-- replacing the `DB_URI` in `test.env` file by `DB_URI=postgresql://localhost:5432/test`
+- replacing the `DB_URI` in `tests/test.env` file by `DB_URI=postgresql://localhost:5432/test`
 
-- then run all tests with `uv run pytest -c pytest.ci.ini`
+- then running all tests with:
+
+```bash
+export PGPASSWORD=mypassword
+uv run pytest -c pytest.ci.ini
```
acasajus, line 103: passwords are already set up for a pre-set in the text.env file with also a pre-set test user. Let's keep it that way so we don't have to finaggle with env vars to run the tests.

springcomp: On a fresh Ubuntu install with docker installed I reviewed those instructions.

> passwords are already set up for a pre-set in the `test.env` file with also a pre-set test user.
> Let's keep it that way so we don't have to finaggle with env vars to run the tests.

While the contributing instructions work when using the `sl-test-db` docker-based PostgreSQL test DB,
they **do not work** when running a **local** PostgreSQL DB instead.

The reason for this is that a local DB is using the current logged in unix user as the database user.

The instructions for running tests against a local DB are conflicting.
One is instructed to change `DB_URI` to `DB_URI=postgresql://localhost:5432/test`.

This can only work in either of two ways:
- Either the `PGPASSWORD` environment variable exists with the db password for the current unix user.
- Or one has to modify `/etc/postgresql/12/main/pg_hba.conf` to enable passwordless connections.

PR #2802, opened 2026-09-02
Dmarc dkim verification

oliver-gordon:
Introduce validation checks for dmarc and dkim, use the same issue counter as mx record checks to trigger notifications. This allows us to avoid new columns and a migration. Two new email templates for user notifications

### DIFF
--- a/tasks/check_custom_domains.py
+++ b/tasks/check_custom_domains.py
@@ -118,8 +118,10 @@ def check_single_custom_domain(custom_domain: CustomDomain):
         return
     user = custom_domain.user
     # snapshot before validate_dkim_records()/validate_dmarc_records() below can commit
-    # and bump updated_at, which would otherwise throw off the once-a-day throttle
-    last_updated_at = custom_domain.updated_at
+    # and bump these, which would otherwise throw off the once-a-day throttles
+    mx_last_updated_at = custom_domain.updated_at
+    dkim_last_updated_at = custom_domain.dkim_nb_failed_checks_updated_at
+    dmarc_last_updated_at = custom_domain.dmarc_nb_failed_checks_updated_at
 
     mx_domains = get_mx_domains(custom_domain.domain)
     validator = CustomDomainValidation(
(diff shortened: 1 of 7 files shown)


tasks/check_custom_domains.py
```diff
@@ -54,60 +84,91 @@ def check_all_custom_domains():
...
         partner_domains=config.PARTNER_DNS_CUSTOM_DOMAINS,
         partner_domains_validation_prefixes=config.PARTNER_CUSTOM_DOMAIN_VALIDATION_PREFIXES,
     )
     expected_custom_domains = validator.get_expected_mx_records(custom_domain)
-    if not is_mx_equivalent(mx_domains, expected_custom_domains):
-        user = custom_domain.user
+    mx_ok = is_mx_equivalent(mx_domains, expected_custom_domains)
+    if custom_domain.is_sl_subdomain:
+        dkim_ok = True
+        dmarc_ok = True
```
acasajus: id move this to the top and if it's `is_sl_subdomain` return like the `custom_domain.user.disabled`. There are some domains with whacky states and I'd avoid messing with those for now.

tasks/check_custom_domains.py
```diff
@@ -54,60 +84,123 @@ def check_all_custom_domains():
...
-    if not is_mx_equivalent(mx_domains, expected_custom_domains):
-        user = custom_domain.user
+    mx_ok = is_mx_equivalent(mx_domains, expected_custom_domains)
+
+    dkim_errors = validator.validate_dkim_records(custom_domain)
+    dkim_ok = len(dkim_errors) == 0
+    dmarc_ok = validator.validate_dmarc_records(custom_domain).success
+
+    domain_dns_url = f"{config.URL}/dashboard/domains/{custom_domain.id}/dns"
+    provider = "Proton" if user.has_used_alias_from_partner() else "SimpleLogin"
```
cquintana92: Probably we should check whether the domain has created via partner, as a user can have domains added via Pass or via SL

cquintana92 (approved):

